PR #31, opened 2020-07-21
feat: Magically move your site to FC from the dashboard

gavindsouza:
Screenshot 2020-08-13 at 1 59 44 PM

Depends on https://github.com/frappe/agent/pull/12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,25 @@
+os: linux
+dist: focal
+language: python
+python: 3.7
+
+git:
+  depth: 1
+
+cache:
+  pip: true
+
+install:
+  - pip install $TRAVIS_BUILD_DIR
+
+jobs:
+  include:
+  - name: Vagrant Hello World
+    before_script: python3 $TRAVIS_BUILD_DIR/backbone/setup.py
+    script:
+      - cd $TRAVIS_BUILD_DIR/backbone
+      - sudo -E su $USER -c 'vagrant up --provider=libvirt'
+      - sudo -E su $USER -c 'vagrant ssh -c "echo Hello, World"'
+
+  - name: Backbone Tests
+    script: backbone tests
--- a/backbone/README.md
+++ b/backbone/README.md
@@ -1,0 +1,7 @@
+# Backbone
+
+> Note: Not to be confused with the scrapped project **Frappe Backbone**
+
+## Installation
+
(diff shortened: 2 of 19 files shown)


press/utils.py
```diff
@@ -106,3 +106,56 @@ def get_minified_script():
 	migration_script = "../apps/press/press/scripts/migrate.py"
 	script_contents = open(migration_script).read()
 	return minify(script_contents)
+
+
+def verify_frappe_site(site):
```
adityahase: Verify with credentials.

press/utils.py
```diff
@@ -106,3 +106,56 @@ def get_minified_script():
...
+	api = api_key and api_secret
+	passwd = usr and pwd
+
+	def url(path):
+		host = site.split(":")[0]
+		file = path.lstrip(f"./{host}/private/")
+		url = f"{schema}://{site}/{file}?sid={sid}"
+		return url
+
+	if api:
```
adityahase: Get rid of this. Setting up a token, just for this is too much work anyway.

press/utils.py
```diff
@@ -106,3 +106,56 @@ def get_minified_script():
...
+		data = res.json()
+		if data.get("message") == "pong":
+			status = True
+	except Exception:
+		pass
+
+	return {"schema": schema, "status": status}
+
+
+def get_frappe_backups(site, auth):
```
adityahase: Can we move this to client-side? So local sites would also work?

dashboard/src/views/NewSite.vue
```diff
@@ -150,6 +150,120 @@
...
+										class="w-5 h-5 p-1 mr-2 text-green-500 bg-green-100 rounded-full"
+										v-if="verifiedFrappeSite === true"
+									/>
+									<FeatherIcon
+										name="x"
+										class="w-5 h-5 p-1 mr-2 text-red-500 bg-red-100 rounded-full"
+										v-if="verifiedFrappeSite === false"
+									/>
+								</p>
+								<Button class="mt-6 w-30" type="primary" @click="verifySite">
```
adityahase: Too many visible primary buttons at once.

press/utils.py
gavindsouza: we're passing the file URL directly to agent, local sites wouldn't work anyway

PR #46, opened 2020-09-01
feat: Marketplace

netchampfaris:
image

## Development:

1. Run `yarn build-marketplace-css` in `apps/press`
1. Visit `/marketplace`

image

image

image


press/public/marketplace/js/call.js
```diff
@@ -0,0 +1,72 @@
+export default async function call(method, args) {
```
rmehta, line 1: yet another call implementation 😅

PR #64, opened 2020-11-02
fix: Dashboard updates

gavindsouza:
* Added 'Site Logs' tab in Pending state
* Added 'Site Config', 'Database' tabs in Broken state
* Added 'Backups' tab in Suspended state
* Added 'Backups' tab in Inactive state
* Disable taking backups via Dashboard when site not in Active state -- don't show button
* Don't allow updating Site config unless in Active and Broken state -- show ErrorMessage saying not allowed

### DIFF
--- a/dashboard/src/views/SiteBackups.vue
+++ b/dashboard/src/views/SiteBackups.vue
@@ -86,11 +86,11 @@
 				<div class="px-6 mt-2 text-base text-gray-600" v-else>
 					No backups found
 				</div>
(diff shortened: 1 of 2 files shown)


dashboard/src/views/SiteBackups.vue
```diff
@@ -90,7 +90,7 @@
 					<Button
```
netchampfaris, line 90: Hide the button if site is not active

PR #76, opened 2020-12-09
fix(SiteDomain): Remove host for default domain only if redirect exists

balamurali27:


press/press/doctype/site_domain/site_domain.py
```diff
@@ -85,7 +88,11 @@ def retry(self):
 
 	def on_trash(self):
 		self.disavow_agent_jobs()
-		self.create_remove_host_agent_request()
+		if self.default:
+			if self.redirect_to_primary:
+				self.create_remove_host_agent_request()
+		else:
+			self.create_remove_host_agent_request()
```
gavindsouza:
```suggestion
		if not self.default:
			self.create_remove_host_agent_request()
		elif self.redirect_to_primary:
			self.create_remove_host_agent_request()
```

Lesser indents?

balamurali27: Damn. Niice

PR #172, opened 2021-08-09
feat: Marketplace App Add flow

NagariaHussain:
Flow to add an app to marketplace either from existing owned apps or directly from GitHub. 

Also, includes overall polishes to marketplace UI. 

🚀


dashboard/src/views/NewApp.vue
```diff
@@ -60,13 +62,25 @@
 					>
 						Validate App
 					</Button>
+					<select
+						class="my-2 form-select block mr-2"
+						v-if="validatedApp && !benchName"
+						v-model="selectedVersion"
+					>
+						<option
+							v-for="v in ['Version 13', 'Version 12', 'Nightly']"
```
netchampfaris: I hope this is going to be replaced by a server fetched list

dashboard/src/views/NewApp.vue
```diff
@@ -92,7 +107,8 @@ export default {
 		return {
 			selectedRepo: null,
 			selectedInstallation: null,
-			selectedBranch: null
+			selectedBranch: null,
+			selectedVersion: 'Version 13'
```
netchampfaris: No hardcoding please

dashboard/src/views/NewApp.vue
```diff
@@ -128,19 +144,26 @@ export default {
 		},
 		addApp() {
 			return {
-				method: 'press.api.app.new',
+				method: this.benchName
+					? 'press.api.app.new'
+					: 'press.api.developer.new_app',
```
netchampfaris: This is not the right way to reuse this component. It should be separated into a component that only gives us a github app. Let me know if you want to discuss this.

press/api/developer.py
```diff
@@ -16,6 +18,14 @@
 )
 
 
+@frappe.whitelist()
+def become_developer():
+	current_team = get_current_team()
+	team = frappe.get_doc("Team", current_team)
```
netchampfaris:
```suggestion
	team = get_current_team(get_doc=True)
```

PR #173, opened 2021-08-11
feat: Login with frappe.io

balamurali27:


press/api/account.py
```diff
@@ -415,3 +417,19 @@ def redirect_to(location):
 		301,
 		{"Location": location, "Cache-Control": "no-store, no-cache, must-revalidate"},
 	)
+
+
+def get_frappe_io_auth_url() -> Union[str, None]:
+	"""Get auth url for oauth login with frappe.io."""
+
+	provider = frappe.get_last_doc(
```
netchampfaris: This will break if the doc is not found? This feature implicitly depends on Social Login Key doc to be created manually.

balamurali27: Yeah, Travis actually caught this. I'll fix

press/api/account.py
```diff
@@ -415,3 +419,22 @@ def redirect_to(location):
...
+		)
+	except DoesNotExistError:
+		return
+
+	if (
+		get_oauth_keys(provider.name)
+		and provider.get_password("client_secret")
+		and provider.client_id
+		and provider.base_url
+	):
```
gavindsouza, line 439:
```suggestion
	if (
		provider.base_url
		and provider.client_id
		and get_oauth_keys(provider.name)
		and provider.get_password("client_secret")
	):
```

minor perf

balamurali27: ah yes. works. Had copied this from login.py in frappe though :see_no_evil:

PR #221, opened 2021-09-28
feat: Option to disable site auto-update

NagariaHussain:
- Introduces `Skip Auto Updates` field (check) in site doctype
- Skip auto updates for the sites with the above field checked
- Update `app_email` to use `frappe.io` domain

### DIFF
--- a/press/press/doctype/site_update/site_update.py
+++ b/press/press/doctype/site_update/site_update.py
@@ -160,7 +160,7 @@ def sites_with_available_update():
 			"status": ("in", ("Active", "Inactive", "Suspended")),
 			"bench": ("in", benches),
 		},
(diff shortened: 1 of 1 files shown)


press/press/doctype/site_update/site_update.py
```diff
@@ -200,6 +201,10 @@ def schedule_updates():
 			frappe.db.rollback()
 
 
+def should_not_skip_auto_updates(site):
+	return not site.skip_auto_updates
```
netchampfaris, line 205: site is not a doc here, the `skip_auto_updates` field needs to fetched in the `get_all` method

NagariaHussain: My bad, fixed it.

PR #223, opened 2021-09-28
feat: login as admin reason popup in dashboard

kamaljohnson:
image
image

### DIFF
--- a/dashboard/src/views/Site.vue
+++ b/dashboard/src/views/Site.vue
@@ -151,7 +151,8 @@ export default {
 			}
 			this.$confirm({
 				title: 'Login as Administrator',
-				message: 'State the reason to login to as administrator.',
+				message:
+					'Please enter reason for this login. Site owner will be notified of this action.',
 				actionLabel: 'Login',
 				textBox: true,
 				action: (closeDialog, textBoxInput) => {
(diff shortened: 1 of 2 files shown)


dashboard/src/components/ConfirmDialogs.vue
```diff
@@ -15,6 +15,13 @@
 				v-if="dialog.resource"
 				:error="dialog.resource.error"
 			/>
+			<Input
```
NagariaHussain, line 18: I think it would better to add a slot here instead of directly putting the Input component and pass the Input from the parent component and manage the contents of the textbox from there. Although, I am not 100% this approach will work, but we can give it a try.

kamaljohnson: @NagariaHussain  the CustomDialog component is present in the app.vue file and it is called globally with a single instance. 
do you know how to pass vue component to the $confirm() as prop? or will this even work?

balamurali27: I shall merge this PR now. @kamaljohnson can refactor it later on.

kamaljohnson: ok

PR #230, opened 2021-09-30
fix: Remove app from bench action

NagariaHussain:
- [x] Throw if the given `app` to remove is installed on a non-archived site
- [x] Fix remove app from release group app
- [x] Fetch App Updates from bench dashboard
- [x] Deploy the updated bench (in the update dialog itself `Will be uninstalled`)
- [x] Refactor Update Card
- [x] Show errors in the confirm dialog

Fetch latest update option in dropdown:

Screenshot 2021-10-02 at 3 21 44 PM

Refactored App Card: 

updated_card

### DIFF
--- a/press/press/doctype/blocked_domain/blocked_domain.py
+++ b/press/press/doctype/blocked_domain/blocked_domain.py
@@ -4,5 +4,6 @@
 # import frappe
 from frappe.model.document import Document
 
+
 class BlockedDomain(Document):
 	pass
--- a/press/press/doctype/blocked_domain/test_blocked_domain.py
+++ b/press/press/doctype/blocked_domain/test_blocked_domain.py
@@ -4,5 +4,6 @@
 # import frappe
 import unittest
 
+
 class TestBlockedDomain(unittest.TestCase):
 	pass


press/api/bench.py
```diff
@@ -194,6 +195,14 @@ def installable_apps(name):
 	return [app for app in apps if app["name"] not in installed_apps]
 
 
+@frappe.whitelist()
+@protected("Release Group")
+def fetch_latest_app_update(name, app):
+	rg: ReleaseGroup = frappe.get_doc("Release Group", name)
+	app_source = rg.get_app_source(app)
+	app_source.create_release()
```
gavindsouza, line 203: Shouldn't the API return something since it's named `fetch_*`?

press/api/bench.py
```diff
@@ -204,11 +213,30 @@ def add_app(name, source, app):
...
-			break
+	release_group: ReleaseGroup = frappe.get_doc("Release Group", name)
+
+	# Sites on this release group
+	sites = frappe.get_all(
+		"Site", filters={"group": name, "status": ("!=", "Archived")}, pluck="name"
+	)
+
+	site_apps = frappe.get_all(
+		"Site App", filters={"parent": ("in", sites), "app": app}, fields=["parent"]
```
gavindsouza, line 224: could've used `pluck="parent"` instead to avoid the get below

PR #235, opened 2021-10-04
feat(UserSSHCertificate): Log employee logins to FC servers

balamurali27:
This allows System Users to ssh into FC servers as `frappe` user after creating a User SSH Certificate doc. Copied the implementation from https://github.com/frappe/central

### DIFF
--- a/press/press/doctype/user_ssh_certificate/user_ssh_certificate.py
+++ b/press/press/doctype/user_ssh_certificate/user_ssh_certificate.py
@@ -35,6 +35,7 @@ class UserSSHCertificate(Document):
 				"user": self.user,
 				"valid_until": [">", frappe.utils.now()],
 				"access_server": self.access_server,
+				"all_servers": self.all_servers,
 				"docstatus": 1,
 			},
 		):
@@ -61,7 +62,10 @@ class UserSSHCertificate(Document):
 			public_key.write(self.ssh_public_key)
 			public_key.flush()
 
-		principal = self.access_server
+		if self.all_servers:
+			principal = "all-servers"
+		else:
+			principal = self.access_server
 
 		# try generating a certificate for the /tmp key.
 		try:
--- a/press/press/doctype/user_ssh_certificate/user_ssh_certificate.json
+++ b/press/press/doctype/user_ssh_certificate/user_ssh_certificate.json
@@ -5,6 +5,7 @@
  "editable_grid": 1,
  "engine": "InnoDB",
  "field_order": [
+  "all_servers",
   "server_type",
   "access_server",
   "column_break_1",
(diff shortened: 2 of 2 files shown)


press/press/doctype/team/team.json
```diff
@@ -28,7 +28,10 @@
   "address_html",
   "custom_apps_section",
   "github_access_token",
-  "feature_flags_section"
+  "feature_flags_section",
+  "ssh_details_section",
```
netchampfaris: ssh details should be per user and not per team?

balamurali27: Yeah, we did it like this just for a POC. Maybe we could add it to team member? Adding custom field to User doc wasn't allowed. This feature will only be used by our devs anyway. So it's probably fine?

netchampfaris: The SSH feature will also need per user track of SSH keys. Maybe create a doctype for keeping records of SSH keys?

balamurali27: Oh yeah, sounds simpler yeah. Will do

press/press/doctype/user_ssh_certificate/user_ssh_certificate.py
```diff
@@ -0,0 +1,119 @@
...
+	def before_save(self):
+		# decode the ssh key and generate a fingerprint
+		ssh_key = self.ssh_public_key.strip().split()[1]
+		ssh_key_b64 = base64.b64decode(ssh_key)
+		sha256_sum = hashlib.sha256()
+		sha256_sum.update(ssh_key_b64)
+		self.ssh_fingerprint = safe_decode(base64.b64encode(sha256_sum.digest()))
+
+	def before_submit(self):
+		# extract key_type (eg: rsa, ecdsa, ed25519.) for naming convention
```
gavindsouza, line 53: Seems like a lot of different things? Could be moved into individual methods like `self.flush_public_key()`, `self.generate_something_certificate()`, etc.

press/press/doctype/user_ssh_certificate/user_ssh_certificate.py
```diff
@@ -0,0 +1,119 @@
...
+def get_ssh_command(docname):
+	ssh_port = 22
+	ssh_user = "frappe"
+	certificate = frappe.get_doc("User SSH Certificate", docname)
+	server = certificate.access_server
+
+	if certificate.ssh_command:
+		return certificate.ssh_command
+
+	return f"ssh {server} -p {ssh_port} -l {ssh_user}"
```
gavindsouza, line 115: docname is not a very nice arg name 😢 . Also, might've gotten away without using the get_doc, although it has it's benefits.

press/press/doctype/user_ssh_certificate/user_ssh_certificate.py
balamurali27: I copied most of this from central. So, it seems wasteful to refactor when it works :sweat_smile:

press/press/doctype/user_ssh_certificate/user_ssh_certificate.py
balamurali27: https://github.com/frappe/press/pull/295 fixed here

PR #237, opened 2021-10-07
fix: add notify email as optional field for site and team.

kamaljohnson:
- site activities (login as admin) will be send to the email added in notify email block in team
 - this field can also be configured within each sites
 - by default the team notify email is set in site notify email

### DIFF
--- a/press/press/doctype/team/team.json
+++ b/press/press/doctype/team/team.json
@@ -171,6 +171,8 @@
   },
   {
    "depends_on": "eval:doc.send_notifications;",
+   "fetch_from": "user.email",
+   "fetch_if_empty": 1,
    "fieldname": "notify_email",
    "fieldtype": "Data",
    "label": "Notify Email",
@@ -233,7 +235,7 @@
    "link_fieldname": "team"
   }
  ],
- "modified": "2021-10-08 09:49:10.725936",
+ "modified": "2021-10-08 10:25:56.640088",
  "modified_by": "Administrator",
  "module": "Press",
  "name": "Team",


press/press/doctype/site/site.py
```diff
@@ -104,6 +105,9 @@ def validate_site_config(self):
 		# if not self.is_new() and self.has_value_changed("config"):
 		# 	Agent(self.server).update_site_config(self)
 
+	def validate_notify_email(self):
+		frappe.utils.validate_email_address(self.notify_email, True)
```
balamurali27:
```suggestion
	def validate_notify_email(self):
		if self.notify_email:
			frappe.utils.validate_email_address(self.notify_email, True)
```

press/press/doctype/site_activity/site_activity.py
```diff
@@ -11,17 +11,18 @@
...
-				args={"site": self.site, "user": self.owner, "reason": self.reason},
-				reference_doctype=self.doctype,
-				reference_name=self.name,
-				now=True,
-			)
+			recipient = frappe.db.get_value("Site", self.site, "notify_email")
+			if recipient:
+					frappe.sendmail(
+						recipients=[recipient],
+						# TODO: update recipient for email to use custom recipient configured in team <29-09-21, Balamurali M> #
```
balamurali27:
```suggestion
						recipients=[recipient],
```

press/press/doctype/team/team.py
```diff
@@ -28,6 +28,12 @@ def validate(self):
 		self.set_team_currency()
 		self.set_default_user()
 		self.set_billing_name()
+		self.validate_notify_email()
+
+	def after_insert(self):
+		if self.send_notifications:
+			self.notify_email = self.name
+		self.save()
```
gavindsouza, line 36: Seems hacky but lol

press/patches/v0_0_1/add_team_name_as_default_notify_email.py
```diff
@@ -0,0 +1,14 @@
+import frappe
+
+frappe.reload_doctype("Team")
+frappe.reload_doctype("Site")
+
+def execute():
+    """Sets the value of notify email as team name"""
+    teams = frappe.get_all("Team", pluck="name")
+    for team in teams:
+        frappe.db.set_value("Team", team, "notify_email", team)
```
gavindsouza, line 10: This could've been one query, and run only for active teams

----

After https://github.com/frappe/frappe/pull/14060, you could do something like this:

```python
from frappe.query_builder import Column
frappe.db.set_value("Team", {"status": "Active"}, "notify_email", Column("name"))
```

PR #298, opened 2021-12-06
feat: country dropdown -> text field

kamaljohnson:


press/api/account.py
```diff
@@ -69,6 +69,18 @@ def setup_account(
...
+		if not is_invitation and country:
+			cl = country_list()
+			flag = False
+			for c in cl:
+				if country.casefold() == c.name.casefold():
+					country = c.name
+					flag = True
+					break
+			if not flag:
+				frappe.throw("Country filed should be a valid country name")
```
NagariaHussain: This is not a clean way to do it. Consider using find:

```python
from frappe.core.utils import find

find(["India", "Australia"], lambda x: x.lower()  == "India".lower()) # returns 'India'
find(["India", "Australia"], lambda x: x.lower()  == "Indiana".lower()) # returns `None`, which can be used to test
````

File: press/api/central.py [hunk 2]
```diff
@@ -21,6 +21,19 @@ def account_request(
...
+		flag = False
+		for c in cl:
+			if country.casefold() == c.name.casefold():
+				country = c.name
+				flag = True
+				break
+		if not flag:
+			frappe.throw("Country filed should be a valid country name")
+	else:
+		frappe.throw("Country field should not be empty")
```
NagariaHussain: Similar to above. Also, it is better to use `pluck='name'`:

```python
cl = frappe.db.get_all("Country", pluck="name")
```

press/www/signup.html
```diff
@@ -153,7 +149,7 @@ <h2 class="font-size-base">Verification Email Sent!</h2>
 		},
 		country: (value) => {
 			if (!value) {
-				return "Please select country";
+				return "Please enter country";
```
NagariaHussain:
```suggestion
				return "Please enter your country name";
```

?

PR #301, opened 2021-12-06
feat: add checkbox for tc and pp agreement in signups

kamaljohnson:
## Signups from ERPNext
telegram-cloud-photo-size-5-6158955542115102380-y

## General signups
telegram-cloud-photo-size-5-6158955542115102381-y

### DIFF
--- a/press/api/account.py
+++ b/press/api/account.py
@@ -77,9 +77,9 @@ def setup_account(
 			country = find(all_countries, lambda x: x.lower() == country.lower())
 			if not country:
 				frappe.throw("Country filed should be a valid country name")
-    
-    if not accepted_user_terms:
-      frappe.throw("You need to accept our Terms of Service & Privary Policy")
+
+	if not accepted_user_terms:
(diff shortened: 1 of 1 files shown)


dashboard/src/views/SetupAccount.vue
```diff
@@ -100,7 +111,8 @@ export default {
 			userExists: null,
 			invitationToTeam: null,
 			isInvitation: null,
-			country: null
+			country: null,
+			userAcceptTerms: false
```
NagariaHussain: The variable name could be:

```suggestion
			termsAccepted: false
```

press/api/account.py
```diff
@@ -52,6 +52,7 @@ def setup_account(
 	is_invitation=False,
 	country=None,
 	user_exists=False,
+	user_accept_terms=False
```
NagariaHussain: Again,  boolean variables should be named like `accepted_user_terms`.

PR #323, opened 2021-12-22
feat: Improvements to team section

rutwikhdev:
- [x] ability to remove team members (!only owners)
- [x] refactor all member actions in single "Manage" dialog
- [x] allow members to leave teams

### DIFF
--- a/press/api/account.py
+++ b/press/api/account.py
@@ -393,7 +393,7 @@ def leave_team(team):
 	cur_team = frappe.session.user
 
 	if team_to_leave.user == cur_team:
-		frappe.throw("Cannot leave this team as your the owner.")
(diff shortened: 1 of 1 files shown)


press/api/account.py
```diff
@@ -381,6 +387,17 @@ def switch_team(team):
 		}
 
 
+@frappe.whitelist()
+def leave_team(team):
+	team_to_leave = frappe.get_doc("Team", team)
+	cur_team = frappe.session.user
+
+	if team_to_leave.user == cur_team:
+		frappe.throw("Cannot leave this team as your the owner.")
```
NagariaHussain:
```suggestion
		frappe.throw("Cannot leave this team as you are the owner.")
```

PR #327, opened 2022-01-04
fix: Temporary fix for failing patch

rutwikhdev:


press/patches/v0_0_1/create_child_table_records_in_teams.py
```diff
@@ -1,14 +1,18 @@
...
-		team.save()
+		try:
+			team = frappe.get_doc("Team", name)
+			team.append("communication_emails", {"type": "invoices", "value": team.name})
+			team.append(
+				"communication_emails", {"type": "marketplace_notifications", "value": team.name}
+			)
+			team.save()
+		except Exception as e:
+			log_error(title="Weird Country Name", data=e)
```
ankush, line 18: catching exceptions and ignoring can quite likely result in weird partial commits. 

e.g. if there is some post-save validation and that validation fails.. the document will still get saved in the database. 

This doctype does seem to have such validations! 

https://github.com/frappe/press/blob/03810ddbb4fbf573f4b34df9295e6afb647cd1f8/press/press/doctype/team/team.py#L213 

Consider rolling back or using savepoints.

PR #337, opened 2022-01-19
feat: use marketplace api to validate email app subscriptions

rutwikhdev:
* changed update site config logic because update_config(whitelisted) function throws Permission Errors.
* validating email plans from marketplace developer api

### DIFF
--- a/press/api/email.py
+++ b/press/api/email.py
@@ -10,6 +10,10 @@ from datetime import datetime
 
 import frappe
 from press.utils import log_error
+from press.api.developer.marketplace import (
+	get_subscription_info,
+	get_subscription_status,
+)
 
 
 @frappe.whitelist(allow_guest=True)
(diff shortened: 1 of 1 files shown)


press/api/email.py
```diff
@@ -67,19 +67,35 @@ def get_analytics(**data):
...
 	"""
-	# ToDo: verify this key and validate plan from marketplace
-	if secret_key == "fcmailfree100":
-		count = frappe.db.count("Mail Log", filters={"site": site, "status": "delivered"})
-		print(count)
-		if count < 100:
-			return True
-	elif secret_key == "fcmailfrappeteam$1152":
+	# TODO: remote this wildcard key after marketplace api and docs is up
+	if secret_key == "fcmailfrappeteam$1152":
```
ankush, line 76: bruh

ankush: This can still be moved in env variables / secret store or something I guess? why put it in code where it can potentially live forever.

NagariaHussain: This can be removed, right @rutwikhdev since marketplace apis are up?

PS: It was removed from master, ok you meant version control. Yeah, it doesn't matter in this particular case since it was some random string selected by Rutwik for beta testing.

But for secrets it is better to use password fields, since that's what we do for any secret in press...

rutwikhdev: > This can be removed, right @rutwikhdev since marketplace apis are up?

Yep removed this was for testing before paid apps was up.

PR #341, opened 2022-01-26
feat: Available on plans table for Marketplace App Plan

NagariaHussain:
We can now define what versions of frappe (fetched from the selected release group) does the plan support. Based on this the plans are filtered. 

Use case: Wiki Plans only for Nightly version

Screenshot 2022-01-27 at 2 55 03 AM

### DIFF
--- a/dashboard/src/components/ChangeAppPlanSelector.vue
+++ b/dashboard/src/components/ChangeAppPlanSelector.vue
@@ -55,11 +55,6 @@ export default {
 	},
 	resources: {
 		getAppPlans() {
-			console.log({
-				app: this.app.name,
-				release_group: this.group,
-				frappe_version: this.frappeVersion
-			});
 			return {
 				method: 'press.api.marketplace.get_app_plans',
 				params: {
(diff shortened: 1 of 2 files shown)


press/api/marketplace.py
```diff
@@ -419,13 +419,17 @@ def get_marketplace_subscriptions_for_site(site: str):
...
+
+	if (not frappe_version) and release_group:
+		frappe_version = frappe.db.get_value("Release Group", release_group, "version")
+
+	return marketplace_app.get_plans(frappe_version)
 
 
 @frappe.whitelist()
-def get_apps_with_plans(apps):
+def get_apps_with_plans(apps, release_group: str = None):
```
ankush, line 432: If default is `None`. it should be `Optional[str]` 😬

NagariaHussain: Yeah, that's more correct. But we sparingly use types for anything, I use them only when the parameter name can confuse wether to pass a doc or doc name, it's cleaner than doing app_name instead of app... 

I don't know, maybe we should use types everywhere or nowhere 😅

ankush: nah using them just as docs is also better than not having them 😄 

I just pointed out in case you didn't know. `Optional` lets your editor/checker know that return value can be potentially None and hence check it before operating on it.

PR #352, opened 2022-02-22
feat: Discounts on invoices

NagariaHussain:
- [x] A Child table for discounts in Invoice
- [x] Flat On Total Discount (Both discount & Percentage)
- [x] A Child table for discount in Team
- [x] Add Tests

The discounts from team will be carried over to new invoices created.

## No Discount

Screenshot 2022-02-23 at 12 08 49 PM
Screenshot 2022-02-23 at 12 08 02 PM 1

## Discount Applied

Screenshot 2022-02-23 at 12 09 40 PM
Screenshot 2022-02-23 at 12 09 53 PM

### DIFF
--- a/press/press/doctype/invoice/test_invoice.py
+++ b/press/press/doctype/invoice/test_invoice.py
@@ -303,5 +303,131 @@ class TestInvoice(unittest.TestCase):
 		# balance should 900 after buying prepaid credits
 		self.assertEqual(team.get_balance(), 900)
 
+	def test_single_x_percent_flat_on_total(self):
+		team = frappe.get_doc(
+			doctype="Team", name="testuser@example.com", country="India", enabled=1
+		).insert()
+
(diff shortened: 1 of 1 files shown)


press/press/doctype/invoice/invoice.json
```diff
@@ -347,6 +351,30 @@
...
+   "fieldtype": "Currency",
+   "label": "Total Before Discount",
+   "options": "currency"
+  },
+  {
+   "fieldname": "total_discount_amount",
+   "fieldtype": "Currency",
+   "label": "Total Discount Amount",
+   "read_only": 1
   }
```
ruchamahabal, line 378: Although this is not user-facing, can we only show these fields if the discount is applied?

NagariaHussain: Agreed 👍🏻

PR #355, opened 2022-02-28
feat: Track email service setup

rutwikhdev:
* track the setup from a doctype Mail Setup
* if already setup then ignore
* setup configs in single agent job instead of looping

### DIFF
--- a/press/api/email.py
+++ b/press/api/email.py
@@ -13,6 +13,7 @@ from press.utils import log_error
 from press.api.developer.marketplace import (
 	get_subscription_info,
 )
+from press.api.site import site_config, update_config
 
 
 class PlanExpiredError(Exception):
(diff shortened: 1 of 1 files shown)


press/api/email.py
```diff
@@ -28,17 +29,33 @@ def setup(site):
...
-	site = frappe.get_doc("Site", site)
+	doc_exists = frappe.db.exists({"doctype": "Mail Setup", "site": site})
 
-	config = [
-		{"mail_login": "example@email.com"},
-		{"mail_password": "eDwuygx2j"},
-		{"mail_server": "smtp.gmail.com"},
-		{"mail_port": 587},
+	if doc_exists:
+		doc = frappe.get_doc("Mail Setup", doc_exists[0][0])
```
ruchamahabal, line 35: If you are using frappe `develop` then there was a breaking change in what `frappe.db.exists` returns. 

**Ref PR**: https://github.com/frappe/frappe/pull/16200
**Ref breaking change**: https://github.com/frappe/erpnext/pull/30342#issuecomment-1074101391
**Ref new usage**: https://github.com/frappe/frappe/wiki/Migrating-to-Version-14#return-value-of-frappedbexists

You should rewrite this as:

```python
doc = frappe.get_doc("Mail Setup", doc_exists)
```

You will have to replace all instances in press that are dependent on this behaviour 🙈
Since it returns a str and not a list anymore.

rutwikhdev: > If you are using frappe `develop` then there was a breaking change in what `frappe.db.exists` returns.
> 
> **Ref PR**: [frappe/frappe#16200](https://github.com/frappe/frappe/pull/16200) **Ref breaking change**: [frappe/erpnext#30342 (comment)](https://github.com/frappe/erpnext/pull/30342#issuecomment-1074101391) **Ref new usage**: https://github.com/frappe/frappe/wiki/Migrating-to-Version-14#return-value-of-frappedbexists
> 
> You will have to replace all instances in press that are dependent on this behaviour 🙈 Since it returns a str and not a list anymore.

Got it. We are using `develop` but it's hasn't been updated in a while, guess that's why this is still working

PR #402, opened 2022-06-29
show only paid sites in daily backup audit

shadrak98:
Also, check_bench_fields was broken for some reason. Fixed it.
#### Traceback
```python3
Traceback (most recent call last):
  File "apps/frappe/frappe/core/doctype/scheduled_job_type/scheduled_job_type.py", line 84, in execute
    frappe.get_attr(self.method)()
  File "apps/press/press/press/audit.py", line 176, in check_bench_fields
    BenchFieldCheck()
  File "apps/press/press/press/audit.py", line 56, in __init__
    log[bench_name].update({"Sites on press only": sites_on_press_only})
KeyError: 'bench-0001-003543-f3-bahrain'```

### DIFF
--- a/press/press/audit.py
+++ b/press/press/audit.py
@@ -99,7 +99,7 @@ class BackupRecordCheck(Audit):
 	def __init__(self):
 		log = {self.list_key: []}
 		interval_hrs_ago = datetime.now() - timedelta(hours=self.interval)
-		trial_plans = tuple(frappe.get_all("Plan", dict(disable_backups=1), pluck="name"))
+		trial_plans = tuple(frappe.get_all("Plan", dict(is_trial_plan=1), pluck="name"))
 		cond_filters = " AND site.plan NOT IN {trial_plans}" if trial_plans else ''
 		tuples = frappe.db.sql(
 			f"""
--- a/press/press/doctype/plan/plan.json
+++ b/press/press/doctype/plan/plan.json
@@ -20,8 +20,8 @@
   "max_database_usage",
   "max_storage_usage",
   "column_break_13",
+  "is_trial_plan",
   "offsite_backups",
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: press/press/doctype/plan/plan.json [hunk 1]
```diff
@@ -130,11 +131,17 @@
    "fieldname": "database_access",
    "fieldtype": "Check",
    "label": "Database Access"
+  },
+  {
+   "default": "0",
+   "fieldname": "disable_backups",
+   "fieldtype": "Check",
+   "label": "Disable Backups"
```
NagariaHussain: Negative names for checkbox (booleans) are most of the times confusing and lead to problems later while using them IMO. "Backups Enabled" or "Enable Backups" would be a better name. 

Compare: 


`if backups_enabled: ... else: ...`

Vs 

`if not disable_backups: ... else: ...`

press/press/audit.py
```diff
@@ -98,6 +99,8 @@ class BackupRecordCheck(Audit):
 	def __init__(self):
 		log = {self.list_key: []}
 		interval_hrs_ago = datetime.now() - timedelta(hours=self.interval)
+		trial_plans = tuple(frappe.get_all("Plan", dict(disable_backups=1), pluck="name"))
```
NagariaHussain:
```suggestion
		trial_plans = frappe.get_all("Plan", dict(is_trial_plan=True), pluck="name")
```

Why not this? Then we can use the "is_trial_plan" for more things in the future.

shadrak98: Yeah, it's better to use `is_trial_plan` rather than focusing on backups. Also, `disable_backups` would be misleading if it never actually disable the backup.
Thanks!

PR #428, opened 2022-07-29
chore: check type when validating keys

sagarvora:


NagariaHussain (approved):

press/api/developer/saas.py
```diff
@@ -1,9 +1,7 @@
 import json
 import frappe
 
-
-class InvalidSecretKeyError(Exception):
-	http_status_code = 401
+from press.api.developer.marketplace import raise_invalid_key_error
```
shadrak98: isn't this should be `from press.api.developer` instead of `from press.api.developer.marketplace`? 🤔

shadrak98 (approved):
LGTM

PR #434, opened 2022-08-06
fix(marketplace): documentation link label

nextchamp-saqib:
Fix:

CleanShot 2022-08-06 at 15 50 55@2x


NagariaHussain:
🙈 It has been like this forever... 😂

Thanks!

PR #436, opened 2022-08-16
feat: click on compare app update arrow to see diff

ankush:
Github has special URL for showing diff between two commits. While upgrading you
can now click on the tiny arrow to see the diff.

Screenshot 2022-08-16 at 5 36 42 PM

Example URL: https://github.com/frappe/frappe/compare/cadbf91c5874af6b95165d299039cac26eb6cb10..3ec4a618e3a9c56b56c26f11cd76e157dddd3179 

CAUTION / note to reviewer: 
- I do not have local setup, I've edited this totally #YOLO. So please don't trust my vuejs skillz 
- Some styles _might_ be broken since this moves arrow one level down in element tree)

Other changes:
- Added [.editorconfig](https://editorconfig.org/) so contributors don't need to modify their settings when they work on this project. Inline with other frappe projects. Make sure this is okay for you, I only added basic things based on black and prettier settings.

### DIFF
--- a/dashboard/src/components/AppUpdateCard.vue
+++ b/dashboard/src/components/AppUpdateCard.vue
@@ -37,7 +37,7 @@
 			</a>
 			<a
 				v-if="deployFrom(app)"
-				class="block cursor-pointer"
+				class="flex cursor-pointer flex-col justify-center"
 				:href="`${app.repository_url}/compare/${app.current_hash}..${app.next_hash}`"
 				target="_blank"
 			>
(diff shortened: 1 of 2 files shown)


dashboard/src/components/AppUpdateCard.vue
```diff
@@ -35,7 +35,14 @@
 					{{ deployFrom(app) }}
 				</Badge>
 			</a>
-			<FeatherIcon v-if="deployFrom(app)" name="arrow-right" class="w-4" />
+			<a
+				v-if="deployFrom(app)"
+				class="block cursor-pointer"
```
NagariaHussain:
```suggestion
				class="flex cursor-pointer flex-col justify-center"
```

NagariaHussain (approved):

PR #487, opened 2022-09-13
fix: disable all subs for inactive sites

shadrak98:
For some weird reasons, on-site update subscriptions were not disabled and there was no error log for the same.
Due to this lots of subscriptions remain enabled, even for suspended and archived sites, which in turn causes timeout while creating `usage_record` in loop here:

https://github.com/frappe/press/blob/master/press/press/doctype/subscription/subscription.py#L146-L153

### DIFF
--- a/press/patches/v0_0_4/disable_subscriptions_for_inactive_sites.py
+++ b/press/patches/v0_0_4/disable_subscriptions_for_inactive_sites.py
@@ -1,0 +1,24 @@
+# Copyright (c) 2022, Frappe Technologies Pvt. Ltd. and Contributors
+# For license information, please see license.txt
+
+import frappe
+
+
+def execute():
+	subscription = frappe.qb.DocType("Subscription")
+	site = frappe.qb.DocType("Site")
+
+	inactive_sites = (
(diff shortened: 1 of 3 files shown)


NagariaHussain (approved):

press/patches/v0_0_4/disable_subscriptions_for_inactive_sites.py
```diff
@@ -0,0 +1,24 @@
...
+		.select(subscription.name)
+		.where(site.status.isin(["Archived", "Broken", "Suspended"]))
+	).run(pluck=True)
+
+	(
+		frappe.qb.update(subscription)
+		.set(subscription.enabled, 0)
+		.where(subscription.enabled == 1)
+		.where(subscription.name.isin(inactive_sites))
+	).run()
```
NagariaHussain: Let's not use QB for write/update. We can do this instead:

```
frappe.db.set_value("Subscription", {"enabled": 1, "name": ("in", inactive_sites)}, 0)
```

NagariaHussain (changes requested):

PR #560, opened 2022-11-04
fix: validate version upgrades before insert

shadrak98:
fixes: https://github.com/frappe/press/issues/559


rmehta:
is this still active?

PR #577, opened 2022-11-13
feat: add docker compose

nk521:
Adding docker compose file, init script and env file

To build, just `docker compose up [-d]`.

### DIFF
--- a/docker/.env_example
+++ b/docker/.env_example
@@ -1,0 +1,14 @@
+# MariaDB password
+MYSQL_ROOT_PASSWORD="..."
+
+# Bench admin password
+BENCH_ADMIN_PASSWORD="..."
+
+# Press
+PRESSREPO="https://github.com/frappe/press"
+PRESSBRANCH=""
+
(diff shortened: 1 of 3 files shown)


nk521:
I haven't added the support to run bootstrap.py in this PR.

You may find that here (https://github.com/nk521/press/tree/dockerize/docker).

nk521:
proxy/db/app servers config won't work directly as the lack of correct ssh keys.

I'm thinking of wrapping this docker compose with script which copies the correct ssh keys from host to a folder, which I can mount as a volume instead of mounting the .ssh directly.

nk521:
a small update -

Due to docker builds being local, making a container for press doesn't make sense. I am trying to make it work by using the unix socket in press container (which is the same unix socket as host). This in theory can work but is really hacky.

I still have to look for some other solutions. Nestybox has done some great work that solves this problem (kinda).

nk521:
Closing for now.

PR #646, opened 2022-12-15
fix: reduce concurrent downloads on app servers

nk521:
fixes https://github.com/frappe/press/issues/645


adityahase (commented):
Don't duplicate template files. You can use template from other roles with ../../docker etc.

You'll find an example in reconfigure_prometheus role

Issue #550, opened 2022-10-26
Tabular report from prometheus for load overview on all servers


balamurali27:
To have a birds eye view of server metrics like CPU load, memory, I/O etc in a single report

PR #648, opened 2022-12-16
feat: prometheus report for server stats

shadrak98:
* New Report for having overview of server stats. Currently, it will only display cpu, memory, disk usage in percent.
* Added  a new method in `/api/server.py` to calculate total available resources.


shadrak98:
Fixes https://github.com/frappe/press/issues/550

Issue #675, opened 2023-01-11
Verification of backup files before restoring them


nk521:
We should be checking if the backup that is about be restored is of the same FF version or not.

https://github.dev/frappe/press/blob/64e540c843dd587a1755d870c8b7190ee02530ac/press/agent.py#L104-L130

phot0n:
this is already done on framework level?: https://github.com/frappe/frappe/blob/53f7bca5f3ad8cfffb74563a01f9d42264224482/frappe/commands/site.py#L339C5-L339C17

PR #676, opened 2023-01-11
feat: verify backup file before restore

nk521:
fixes https://github.com/frappe/press/issues/675

### DIFF
--- a/press/agent.py
+++ b/press/agent.py
@@ -10,6 +10,8 @@ from typing import List
 
 import frappe
 import requests
+import gzip
+import re
 from frappe.utils.password import get_decrypted_password
 from press.utils import log_error, sanitize_config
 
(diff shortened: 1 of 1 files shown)


NagariaHussain (commented):
We use frappe.throw(..) everywhere instead of raise exception. IMO better to use it; since it gets better handled in desk and dashboard.

press/agent.py
```diff
@@ -101,6 +103,21 @@ def reinstall_site(self, site):
...
+		frappe_version = re.search("Frappe (\d{1,2})\.\d{1,2}\.\d{1,2}", gzip.decompress(mem_database_file.content).decode("utf-8"))
+
+		if not frappe_version:
+			raise Exception("The backup file is corrupted!")
+
+		source = frappe.get_doc("Bench", site.bench).apps[0].source
+		target_frappe_version = frappe.get_doc("App Source", source).versions[0].version
+
+		if frappe_version != target_frappe_version:
+			raise Exception("You're trying to restore site in the wrong version!")
```
ankush, line 119: - Only reverse restores are not possible, forward restores should be fine (as they're followed with migrate. 
- App sources store `Version 14`, `Version 13` etc not exact version string which is present in backup headers.
- This should be warning an not completely blocking ?

press/agent.py
```diff
@@ -101,6 +103,21 @@ def reinstall_site(self, site):
 			site=site.name,
 		)
 
+	def check_database_file(self, site):
+		remote_database_file = site.remote_database_file
+		mem_database_file = requests.get(url = frappe.get_doc("Remote File", site.remote_database_file).download_link)
```
ankush, line 108: umm, this could be in GBs :sweat_smile: 

Maybe validate this on destination server before restore?

ankush (changes requested):

press/agent.py
```diff
@@ -101,6 +103,21 @@ def reinstall_site(self, site):
...
+		# get frappe version from the first line
+		frappe_version = re.search("Frappe (\d{1,2})\.\d{1,2}\.\d{1,2}", gzip.decompress(mem_database_file.content).decode("utf-8"))
```
shadrak98, line 110: decompressing the backup file just to check version doesn't seem to optimal.
as @ankush pointed out it could be in GBs.

Maybe add a cleanup function?

nk521:
I forgot to mark it as a WIP/draft PR. I'm aware of all the issues you guys mentioned :)

press/agent.py
nk521: Yep. Will have to rewrite in agent. Should've marked this as  a draft PR :P

PR #698, opened 2023-01-31
feat: Manage SAAS subscriptions directly from site

rutwikhdev:
- [x] Load floating banner through a cdn (add to site config https://github.com/frappe/frappe/pull/18841)
- [x] Load susbcriptions and plans through an iframe
- [x] Update address if not added 
- [x] Handle payments

saas

### DIFF
--- a/press/api/developer/marketplace.py
+++ b/press/api/developer/marketplace.py
@@ -171,7 +171,7 @@ def update_billing_info(secret_key: str, data: Dict) -> str:
 
 
 @frappe.whitelist(allow_guest=True)
-def make_payment(secret_key: str, data: Dict) -> str:
+def saas_payment(secret_key: str, data: Dict) -> str:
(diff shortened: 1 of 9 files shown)


press/api/developer/marketplace.py
```diff
@@ -86,6 +88,53 @@ def get_login_url(self):
...
+			self.app_subscription_doc.app,
+			self.app_subscription_doc.site,
+			data["new_plan"]["name"],
+			data["total"],
+			data["total"],
+			data["billing"],
+			False,
+		)
+
+	def login_as_team(self):
```
rutwikhdev, line 135: @adityahase @ankush Do you think there's a better way to do this? Or this looks good enough?

ref: update_billing_info and prepaid_saas_payment does a lot of restricted stuff which need user to be logged in

PR #753, opened 2023-03-23
feat: Bring your Own Device

athul:
This PR aims to bring in BYOD for FC, ie Allows press to use a single server as App and Database Server.

> **Warning**
> This is Yet to be completed. This PR is draft

## What's New?

- A New doctype called `Self Hosted Server` which acts as the source of truth for everything related to Self Hosted Server
- A few new ansible playbooks to fetch data from an existing bench present in the server and create new press sites from the existing bench
- Creates Release Groups from the Existing Bench and deploys sites to Production

### Screenshots

Screenshot 2023-03-23 at 9 13 55 AM
Screenshot 2023-03-23 at 9 14 37 AM

## Tasks Left
- [x] Create a new Server on adding a new doc to the Self Hosted Servers
- [x] Clean up unwanted edits to main doctypes
- [ ] Test it a few times, check for edge cases
- [x] Figure out how to use restore public and private files


athul:
Feels like this is done.

PR #754, opened 2023-03-23
fix: show timestamp and value on site uptime graph

ankush:


dashboard/src/views/site/SiteAnalyticsUptime.vue
```diff
@@ -4,7 +4,7 @@
 			<div class="flex h-8 justify-between">
 				<div
 					v-for="d in data"
-					:key="d.timestamp"
+					:key="d.date"
```
ankush, line 7: I think the key was wrong

reference: https://github.com/frappe/press/blob/541189f0d679490da79f01cf84f7fb04feb5973b/press/api/analytics.py#L83-L94

NagariaHussain (approved):

ankush:
tested on staging w/ @shadrak98

PR #780, opened 2023-04-20
fix(ux): allow app renames

ssiyad:
- Add a DocType (App Rename Map) to keep a list of apps to be renamed
- Pass scripts from these docs to agent

!not tested

### DIFF
--- a/press/press/doctype/site_update/site_update.py
+++ b/press/press/doctype/site_update/site_update.py
@@ -364,11 +364,11 @@ def is_site_in_deploy_hours(site):
 
 
 def update_site_record_for_site_update(
-	site_name: str, site_update: frappe._dict, rollback=False
+	site_name: str, site_update: frappe._dict, recover=False
 ):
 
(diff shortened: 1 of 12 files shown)


press/press/doctype/site_update/site_update.py
```diff
@@ -99,12 +99,32 @@ def validate_apps(self):
 		site_apps = [app.app for app in frappe.get_doc("Site", self.site).apps]
 		bench_apps = [app.app for app in frappe.get_doc("Bench", self.destination_bench).apps]
 
+		fields = ["old_name", "new_name"]
+		filters = { "old_name": ["in", site_apps] }
+		app_renames = frappe.get_list("App Rename Map", fields=fields, filters=filters)
+
+		for i in app_renames:
+			if i.old_name in site_apps:
+				site_apps.remove(i.old_name)
```
balamurali27: I think there is some frappe.utils.find we can use here.

`i` feels weird tbh :sweat_smile:

Issue #779, opened 2023-04-17
Remove "unsaved" status in Server docs


balamurali27:
server docs are "dirty" since the addition of the ssh fields added recently.

athul:
Possibly will be closed with #781

PR #781, opened 2023-04-20
chore: Fix Server Doctypes and Include Proxy Server to work with Single Server

athul:
- Normalized Proxy Server doctype with common doctypes from Server and DB Server. 
- Updated Fixtures to read DB Keys as internal
- Added some small playbooks for self hosted servers

1. Added the missing fields in Proxy Server which were present in Server and MariaDB Server Doctypes
2. Rewrote the flow for Self Hosted Server Setup

```mermaid
graph LR;
      New_Self_Hosted_Server -->Proxy_Server;
      Proxy_Server-->MariaDB_Server-->App_Server;
```
If the Proxy and MariaDB fields are set, then only will the App Server will be created. This is due to App Server `on_insert` requires the Proxy and DB to write to DNS config.

3. Add 2 new playbooks for Self Hosted Proxy and DB Setup
4. Updated Fixtures to reflect the DB related keys as internal in Site's Configuration
5. Use SSH User and Port to run Ansible playbooks on Server related doctypes
6. Added a few fields to Self Hosted Server Doctype(Cluster, Agent Password etc...) Cluster will be selected on Proxy Cluster.

Screenshot 2023-04-24 at 8 10 06 PM

Screenshot 2023-04-24 at 8 13 15 PM

### "One Server to rule them All"

Screenshot 2023-04-24 at 8 14 33 PM


athul:
Merging. Closes #779

PR #782, opened 2023-04-21
feat: create common invoice for child and parent team

shadrak98:
## Changes:

* Now a team can be considered as a child team to avoid creating two separate invoices
* Usage Records for the child team are made under the parent team account
* Common Invoice is created for both the parent and child teams.
* Create/Invite a child team from the dashboard
* Switch to child team
* Transfer a site to another team from dashboard
* Hide Billing section from child team 
* Parent team can see all the sites and benches from its child teams.

### Child Team 
Screenshot 2023-04-21 at 1 58 06 PM

### Site under child team account
Screenshot 2023-04-21 at 2 00 16 PM

### Site under parent team account
Screenshot 2023-04-21 at 2 00 44 PM

### Common Invoice under parent team account for both team
Screenshot 2023-04-21 at 2 01 23 PM

### Can now see child teams in "Teams you are part of section "
Screenshot 2023-05-02 at 3 23 21 PM

### Create a new child team by sending an invite or by adding email of the existing one.
Screenshot 2023-05-02 at 3 21 06 PM

### Transfer a site to another team
Screenshot 2023-05-02 at 5 53 00 PM

### DIFF
--- a/dashboard/src/views/settings/AccountTeam.vue
+++ b/dashboard/src/views/settings/AccountTeam.vue
@@ -4,6 +4,14 @@
 			title="Team"
 			subtitle="Teams you are part of and the current active team"
 		>
+			<template #actions>
+				<Button
+					v-if="showManageTeamButton"
+					@click="showManageTeamDialog = true"
+				>
+					Manage
+				</Button>
+			</template>
 			<ListItem v-for="team in teams" :title="team" :key="team">
 				<template #actions>
 					<div v-if="$account.team.name === team">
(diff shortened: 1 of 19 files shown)


dashboard/src/views/site/Site.vue
```diff
@@ -118,6 +118,32 @@
 				>
 			</template>
 		</Dialog>
+
+		<Dialog
+			:options="{ title: 'Transfer Site to Child Team' }"
```
rutwikhdev: 'Transfer Site to Team'

Rename this since they can transfer the site to any team?

dashboard/src/views/site/Site.vue
```diff
@@ -118,6 +118,32 @@
...
+			v-model="showTransferSiteDialog"
+		>
+			<template #body-content>
+				<Input
+					label="Enter name of the child team"
```
rutwikhdev: 'Enter email of the team'

Since we use email for team names

press/press/doctype/usage_record/usage_record.py
```diff
@@ -29,6 +29,9 @@ def on_cancel(self):
 	def update_usage_in_invoice(self):
 		invoice_type = "Summary" if self.prepaid else "Subscription"
 		team = frappe.get_doc("Team", self.team)
```
rutwikhdev: Here you can do something like,

team = frappe.get_doc("Team", self.parent_team or self.team)

instead of if loop below

press/press/doctype/team/team.py
```diff
@@ -124,19 +124,24 @@ def create_new(
...
+			team.append("communication_emails", {"type": "invoices", "value": user.name})
+			team.append(
+				"communication_emails", {"type": "marketplace_notifications", "value": user.name}
+			)
+		else:
+			team.parent_team = account_request.invited_by
 		team.save(ignore_permissions=True)
 
-		team.create_stripe_customer()
+		if not account_request.invited_by_parent_team:
```
rutwikhdev: Invited teams should also be allowed to create their own sites. This isn't required.

dashboard/src/views/settings/AccountTeam.vue
```diff
@@ -18,20 +26,87 @@
...
+								:loading="$resources.removeMember.loading"
+							>
+								Remove
+							</Button>
+						</template>
+					</ListItem>
+					<div v-if="showManageTeamForm">
+						<h5 class="mt-5 text-sm font-semibold"> Create child team</h5>
+						<Input
+							label="Enter the team Id of child team."
```
rutwikhdev: 'Enter the email address to create a new child team for shared access'

Let's be a little more explicit here

rutwikhdev:
I think parent team will also want access to resources shared with child team. Everything else looks fine.

press/press/doctype/team/team.py
shadrak98: @rutwikhdev They can create their own site but since the billing is handled via parent team thought this step isn't required. Thoughts?

rutwikhdev: It's just a stripe object for unique customer. In case the child team wants to branch out from parent in future, you shouldn't need to create this stuff manually again.

PR #783, opened 2023-04-21
Validation to not skip Major Upgrades Sequentially

athul:
Major Upgrades when skipped can make the whole DB corrupt. Example v12 -> v14 directly shouldn't be done.

- feat: Implement skip Major Upgrades Validation
- fix: Update migrate script to make use of new API

Screenshot 2023-04-22 at 12 20 44 AM

Screenshot 2023-04-22 at 12 21 18 AM

### DIFF
--- a/press/press/doctype/version_upgrade/version_upgrade.py
+++ b/press/press/doctype/version_upgrade/version_upgrade.py
@@ -11,9 +11,9 @@ class VersionUpgrade(Document):
 	doctype = "Version Upgrade"
 
 	def validate(self):
+		self.validate_versions()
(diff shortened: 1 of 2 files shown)


press/press/doctype/version_upgrade/version_upgrade.py
```diff
@@ -13,6 +13,7 @@ class VersionUpgrade(Document):
 	def validate(self):
 		self.validate_same_server()
 		self.validate_apps()
+		self.validate_versions()
```
shadrak98: Maybe add this validation before `validate_same_server`

PR #785, opened 2023-04-25
fix: redirect url

nextchamp-saqib:


press/api/marketplace.py
```diff
@@ -1060,9 +1060,7 @@ def login_via_token(token, team, site):
 		doc.save(ignore_permissions=True)
 		frappe.local.login_manager.login_as(team)
 		frappe.local.response["type"] = "redirect"
-		frappe.local.response[
-			"location"
-		] = f"/dashboard/saas/remote/success?team={team}&site={site}"
+		frappe.local.response["location"] = f"/dashboard/sites/{site}/apps"
 	else:
 		frappe.local.response["type"] = "redirect"
 		frappe.local.response["location"] = "/dashboard/saas/remote/failure"
```
ankush: This one probably doesn't work too @rutwikhdev

PR #808, opened 2023-05-17
feat: Login and signup using google(oauth) for frappe cloud and saas signups

rutwikhdev:
Pros:
- No verification steps(lesser drop-offs 👀)
- Wicked fast

Flows:

1. Login
login

2. Signup
signup


phot0n:
can FW's googleoauth suport be used for this?

https://github.com/frappe/frappe/blob/d0a24e66c5c567f25278dfc5db747b794f8123ee/frappe/integrations/google_oauth.py#L34

we can modify the `_SCOPE` and `redirect_uri` to be passed by the defining function.

PR #809, opened 2023-05-18
Self Service for BYOD

athul:
A UI for Setting up Servers from the Dashboard.

## TODO
- [x] Basic Server Setups
- [x] Update APIs to support Self Hosted Servers
- [x] Trigger Ansible Plays from UI
- [x] Self Serviceable Setups

# Implemented
- New Flow
- New APIs

## Refer Video
https://github.com/frappe/press/assets/40897573/71407c2d-1d65-4d6b-9cc1-eb36932ada21

Post setup, it will be taken to the Servers page and the Ansible plays will be executed which could be seen on that page(usually the setup plays takes around 10-15minutes)


athul:
The current UI implementation is straightforward and creates entries as `s` servers. Will be moving this to custom domains in the next PR.

PR #824, opened 2023-05-23
fix: Plan step insertion & removal along with validation

umer2001:


rutwikhdev:
This step is required for paid apps on marketplace

PR #830, opened 2023-05-26
fix: agent status left pending

umer2001:
cron expression already has 5 *s and we don't need 0/5


adityahase:
[`* * * * *` means once every minute](https://crontab.guru/#*_*_*_*_*)
[`* * * * * 0/5` means once every 5 seconds](https://github.com/kiorky/croniter#id5)

Can you please provide some details about what you're trying to solve?

rutwikhdev:
Closing as unrelated

PR #838, opened 2023-05-31
feat: MariaDB System Variables

balamurali27:


press/press/doctype/database_server_mariadb_variable/patches/add_unique_constraint.py
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) 2021, Frappe Technologies Pvt. Ltd. and Contributors
+# For license information, please see license.txt
+
+
+import frappe
+
+
+def execute():
```
ankush, line 8: This isn't required AFAIK. On first migrate it will run the `on_doctype_update`

balamurali27: oh, TIL

PR #844, opened 2023-06-01
feat: option to update all sites to latest version

### LINKED ISSUE #558: the ability to update all sites
After updating the bench, it is required to update all the sites too, and doing it manually for 5+ more sites is a tedious task. 
A button or something to update all sites at once would be a good addition to FC.

BreadGenie:
fix #558

image

### DIFF
--- a/dashboard/src/views/bench/Bench.vue
+++ b/dashboard/src/views/bench/Bench.vue
@@ -164,7 +164,7 @@ export default {
 						});
 					}
 				},
-				this.bench.status == 'Active' && {
+				this.bench.status == 'Active' && !this.bench.public && {
 					label: 'Update All Sites to Latest Version',
 					icon: 'arrow-up-circle',
 					handler: async() => {
--- a/dashboard/src/views/bench/Bench.vue
+++ b/dashboard/src/views/bench/Bench.vue
@@ -62,6 +62,14 @@ export default {
 				auto: true,
 				onError: this.$routeTo404PageIfNotFound
 			};
+		},
+		updateAllSitesToLatest() {
+			return {
+				method: 'press.api.bench.update_all_sites_to_latest_version',
+				params: {
+					bench_name: this.benchName
+				},
+			}
 		}
 	},
 	activated() {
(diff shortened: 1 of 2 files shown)


press/api/bench.py
```diff
@@ -615,6 +615,24 @@ def update(name, bench):
 	frappe.get_doc("Bench", bench).update_all_sites()
 
 
+@frappe.whitelist()
```
rutwikhdev, line 619: Instead of writing it again just use press.api.bench.update ?

Maybe do something like. This should be enough

```python
benches = frappe.get_all("Bench", {"group": name})
for bench in benches:
    frappe.get_cached_doc("Bench", bench).update_all_sites()
```

press/api/bench.py
```diff
@@ -615,6 +615,24 @@ def update(name, bench):
...
+@protected("Release Group")
+def update_all_sites_to_latest_version(bench_name):
```
rutwikhdev: Rename the function to update_all_sites

dashboard/src/views/bench/Bench.vue
```diff
@@ -62,6 +62,14 @@ export default {
 				auto: true,
 				onError: this.$routeTo404PageIfNotFound
 			};
+		},
+		updateAllSitesToLatest() {
```
rutwikhdev: naming: updateAllSites

press/api/bench.py
BreadGenie: Should we fetch `Bench` that are not archived (pending and installing too maybe)?

Edit: Fetching only Active benches for updating now

press/api/bench.py
```diff
@@ -615,6 +615,14 @@ def update(name, bench):
 	frappe.get_doc("Bench", bench).update_all_sites()
 
 
+@frappe.whitelist()
+@protected("Release Group")
+def update_all_sites(bench_name):
+	benches = frappe.get_all("Bench", {"group": bench_name})
+	for bench in benches:
```
ankush, line 623: Does this verify that you're allowed to do it?

Public benches for instance can have many sites which are not owned by bench owner.

BreadGenie: I've disabled this in public benches since the chances of screwing things up over there is much higher.

PR #869, opened 2023-06-14
fix: route to /billing while clicking on alert for unpaid invoices

### LINKED ISSUE #805: UX: Customers are unaware of the existence of /dashboard/billing
When visiting dashboard page, customer only sees the following:
image

The blue button takes them to stripe invoice page corresponding to last unpaid invoice directly. Even in cases where there are multiple unpaid invoices. After payment is done, the site is still not active because there are still unpaid invoices from the previous months. This leads them to believe that even after payment is done, they're still being asked to pay.

Ideally, the button should just redirect them to /dashboard/billing where they can see the list of all unpaid invoices and pay from there only.

image

Perhaps the pay now button should not be white. Reconsider ui for this element, maybe. Primary buttons on hover, perhaps.

BreadGenie:
fix #805 

https://github.com/frappe/press/assets/63963181/c60b245b-0825-4b95-b3e8-8b1aea455cd5

### DIFF
--- a/dashboard/src/views/billing/AccountBillingPayments.vue
+++ b/dashboard/src/views/billing/AccountBillingPayments.vue
@@ -1,16 +1,14 @@
 <template>
-	<Card
-		title="Billing history"
-		:subtitle="subtitle"
-		v-if="!invoiceName && $resources.pastInvoices.data"
-	>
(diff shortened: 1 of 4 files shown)


rutwikhdev:
Let's show selected filter instead of "Filter (icon)" ?

rutwikhdev (approved):

PR #873, opened 2023-06-15
style(site-apps): show commit tag beside app name always

### LINKED ISSUE #872: Show version info always
now it's only visible on hover

image

BreadGenie:
fix #872 

image


rutwikhdev:
Remove group-hover and group from parent tag. That is also not needed if you want to always show the tag

PR #895, opened 2023-06-25
Subscription Creation for BYOD servers

athul:
- Earlier logic didn't take in Plan Change doctype, added that
- Update verify page to add SSH key to `authorised_keys`. This was a dumb bug which I had initially put for development which I later forgot 🙈 
- Added a button for enabling/disabling read_write mode for DB
- Fetch RAM from Ansible ping results for better auto-scaler allocation

### DIFF
--- a/dashboard/src/views/server/NewSelfHostedServer.vue
+++ b/dashboard/src/views/server/NewSelfHostedServer.vue
@@ -225,7 +225,9 @@ export default {
 				},
 				onSuccess(data) {
 					this.playOutput = data;
-					this.$resources.setupNginx.submit();
+					if (data) {
+						this.$resources.setupNginx.submit();
+					}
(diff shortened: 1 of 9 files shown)


rutwikhdev (commented):
@athul you will have to add "Self Hosted Server" at line https://github.com/frappe/press/blob/7b68af3c84b05665cd072585dd0dbf037f77c6a8/press/press/doctype/subscription/subscription.py#L174

assuming the plan document_type is "Self Hosted Server"

also try triggering https://github.com/frappe/press/blob/7b68af3c84b05665cd072585dd0dbf037f77c6a8/press/press/doctype/subscription/subscription.py#L144

Once you make the change, check if usage record is created properly for Self Hosted Servers and appended to draft invoice

PR #896, opened 2023-06-26
feat: full text search for list views

### LINKED ISSUE #888: Server side search
#865
> Replace search with server side search or frappe global search. Right now we are using Fuse.js for client side search for dashboard as well as marketplace.

BreadGenie:
fix #888


press/patches/v0_0_4/create_search_index_table.py
BreadGenie, line 1: not sure if this is the right place to store the patch

BreadGenie:
closing the PR since full text search isn't as good as the search we already have

Issue #876, opened 2023-06-16
Allow configuring max worker limits on release group


ankush:
Some sites might genuinely need >24 web workers because of volume of requests that they handle and if they have memory available then we should let setting higher number

Issue #578, opened 2022-11-14
Gunicorn/Background worker allocation logic


ankush:
Currently, there's a script that does this magic to allocate # of web and background workers. 

Some numbers:

- Web workers consume anywhere from 100-200MB typically and might need a lot more (assumig +100-500MB) in a burst while processing *SOME* request that needs to fetch a lot of data in memory.
- Background jobs consume  around 50-100 MB while idle but can burst and require 100-1000MB more for some jobs (think reports, reposts, bulk processing, POS etc)

Striking right balance is kinda hard without knowing each bench + site's profile and usage. But we can start with some assumptions and keep tweaking till everything is "stable".

**Some observations**
1. Some background workers are underutilized and sit idle most of the time just consuming RAM and doing nothing. More data on this might help. Refer https://github.com/frappe/frappe/pull/18868 (only does for one bench, we need data over time to answer this) 
2. Web workers count more than peak RPS observed on bench (from gunicorn logs). We can safely assume that mosts requests are subsecond, so this means the remaining web workers are sort of unutilized.  

None of this is a "problem" until we start hitting the memory limit and swap. Then having unused workers is detrimental to overall performance because if numbers were lower it would fit in RAM and still be just as useful if not more. 

Action plan:
- Get some data on utilization vs bg worker count, tweak bg worker number logic based on it.
- Get some data on peak RPS vs web workers and tweak web worker logic based on it.

PR #922, opened 2023-07-11
(WIP): BYOD Updates and Patches

### LINKED ISSUE #917: worker allocation logic 0.75 too little for 4GB ram servers
https://github.com/frappe/press/blob/0906828a59a38bc0ba0686fbbf027c4e4990a9fd/press/press/doctype/server/server.py#L813-L816

RAM | keep free RAM
-- | --
<4 | 1.5
<8 | 2
\>8 | 3

athul:
- [ ] Check for Access every 6hrs
- [x] Add vendor information to Self Hosted Doctype
- [x] Get Private IP from Ping
- [ ] ~Worker Allocation updates. Closes #917~
- [x] UI Updates(forecasted)

### DIFF
--- a/press/press/doctype/server/server.py
+++ b/press/press/doctype/server/server.py
@@ -819,8 +819,12 @@ class Server(BaseServer):
 			usable_ram = max(
 				self.ram - 3000, self.ram * 0.75
 			)  # in MB (leaving some for disk cache + others)
-		if self.is_self_hosted and (self.name == self.database_server):
-			usable_ram -= 2000  # More ram for DB
+		if self.is_self_hosted and (
+			self.name == self.database_server
+		):  # if app and database server are on same machine
+			usable_ram = (
+				self.ram - round(self.ram * 0.685)
+			) * 0.75  # [Current RAM - InnoDB Bufferpool Size] More ram for DB, use 75% of remaining ram for app
 		usable_ram_for_gunicorn = 0.6 * usable_ram  # 60% of usable ram
 		usable_ram_for_bg = 0.4 * usable_ram  # 40% of usable ram
 		max_gunicorn_workers = (
--- a/press/press/doctype/server/server.py
+++ b/press/press/doctype/server/server.py
@@ -335,7 +335,11 @@ class BaseServer(Document):
 			frappe.throw(_("Cannot archive server with benches"))
 		self.status = "Pending"
 		self.save()
-		frappe.enqueue_doc(self.doctype, self.name, "_archive", queue="long")
+		if self.is_self_hosted:
+			self.status = "Archived"
+			self.save()
+		else:
+			frappe.enqueue_doc(self.doctype, self.name, "_archive", queue="long")
 		self.disable_subscription()
 
 	def _archive(self):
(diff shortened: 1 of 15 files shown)


press/press/doctype/server/server.py
```diff
@@ -811,9 +811,16 @@ def auto_scale_workers(self):
...
+		if self.ram < 8000:
+			usable_ram = self.ram - 2000  # in MB (leaving some for disk cache + others)
+		elif self.ram < 4000:
+			usable_ram = self.ram - 1500  # in MB (leaving some for disk cache + others)
+		else:
+			usable_ram = max(
+				self.ram - 3000, self.ram * 0.75
+			)  # in MB (leaving some for disk cache + others)
+		if self.is_self_hosted and (self.name == self.database_server):
+			usable_ram -= 2000  # More ram for DB
```
ankush: This is probably not enough for large DBs. 

This should be roughly equal to bufferpool size that you intend to set. 

If DB starts swapping it will be very bad for performance. 


Is there no way to hard partition memory and "reserve" (not limit) portion of it for a process?

athul: Setting to Bufferpool size is doable. But I don't think we can hard partition for a process.

press/press/doctype/server/server.py
```diff
@@ -811,9 +815,20 @@ def auto_scale_workers(self):
...
+		else:
+			usable_ram = max(
+				self.ram - 3000, self.ram * 0.75
+			)  # in MB (leaving some for disk cache + others)
+		if self.is_self_hosted and (
+			self.name == self.database_server
+		):  # if app and database server are on same machine
+			usable_ram = (
+				self.ram - round(self.ram * 0.685)
+			) * 0.75  # [Current RAM - InnoDB Bufferpool Size] More ram for DB, use 75% of remaining ram for app
```
ankush: Bufferpool shouldn't have 69% of all RAM when it's on app server, that balance is quite off. Maybe you'll have to fix mariadb config in case of single server db+app. 


Bufferpool is supposed to hold "working set of database" in memory. So if you have 1GB of database out of which 50% is frequently read/written then 500MB is good enough size for bufferpool. If you assume 69% and only 10% of it is actually in use then you'll end up wasting memory.

On BYOD where most common case is a single site running on server it doesn't need this much memory. The ratio 69% makes sense for a dedicated DB server because there it's JUST used for DB and nothing else, so whatever is theoretical max can be used.

---

IMO safe enough numbers to begin with:
- Compute usable ram first by subtracting with same old logic to preserve some headroom for memory-intensive tasks and other processes.
- 30-50% of usable RAM for DB
- 30-50% of usable RAM for app server


Also, some configs over these things will help a lot. (not server scripts :woozy_face: but overrides) 

Overrides are required because this math will mostly work fine but when it won't we will have to manually tweak it and make the tweaked numbers stick (right now they go away after bench update, which is very pointless configuration)

ankush (changes requested):

press/press/doctype/server/server.py
ankush: related: https://github.com/frappe/press/issues/876 https://github.com/frappe/press/issues/578

athul: I'm moving Worker Allocation Logic updates to Newer PR

PR #930, opened 2023-07-13
fix: Remove spacing new bench form fields

michellealva:


dashboard/src/views/bench/NewBench.vue
```diff
@@ -20,14 +20,14 @@
 						Name your bench based on its purpose. For e.g., Personal Websites,
 						Staging Bench, etc.
 					</p>
-					<Input class="mt-4" type="text" v-model="benchTitle" />
+					<Input class="" type="text" v-model="benchTitle" />
```
BreadGenie:
```suggestion
					<Input type="text" v-model="benchTitle" />
```
we don't need class if it's empty

dashboard/src/views/bench/NewBench.vue
```diff
@@ -41,7 +41,7 @@
 						Select a Frappe version for your bench.
 					</p>
 					<Input
-						class="mt-4"
+						class=""
```
BreadGenie:
```suggestion
```
same here

BreadGenie:
On 2nd thought let's stick with mt-2 instead of removing them all together

PR #942, opened 2023-07-17
feat: bench config

### LINKED ISSUE #652: Feature to set bench variables from dashboard
Users should be able to set options for bench from dashboard. Eg:
- No. of workers
- HTTP timeout
- commonsiteconfig variables

BreadGenie:
fix #652 

image

Only http_timeout is configurable in bench_config.


dashboard/src/views/bench/BenchConfig.vue
```diff
@@ -0,0 +1,429 @@
+<template>
```
ankush, line 1: This file is duplicating quite a lot from `SiteConfig.vue`.

Maybe refactor the common component out `JsonConfig.vue` or something like that?

ankush: My hunch: you'll need this in few more places as we add/expose more backend features.

BreadGenie: #960

PR #949, opened 2023-07-19
feat(bench-list): add update available and deploy in progress status

### LINKED ISSUE #567: More information on Bench list
By looking at the Bench list, I should be able to figure out
- update available in one or more apps inside the bench
- deploy in progress

image

BreadGenie:
fix #567

- [x] Introduce Update Available and Deploy in Progress status
- [x] Add Filter for the above status
- [x] Tests

### Screenshot
image

image

image

### DIFF
--- a/dashboard/src/components/AlertBenchUpdate.vue
+++ b/dashboard/src/components/AlertBenchUpdate.vue
@@ -1,9 +1,9 @@
 <template>
-	<Alert :title="alertTitle" v-if="show">
+	<Alert :title="alertTitle" v-if="deployInformation?.update_available">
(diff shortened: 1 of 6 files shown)


dashboard/src/components/AlertBenchUpdate.vue
```diff
@@ -110,14 +110,6 @@ export default {
 		}
 	},
 	computed: {
-		show() {
```
BreadGenie, line 113: removing this since the status of bench will always be one among Active and Awaiting Deploy so using `deployInformation?.update_available` instead

PR #953, opened 2023-07-20
feat: Frappe spaces

rutwikhdev:
Web based vscode server for Frappe Cloud. Develop, build, deploy on Frappe Cloud 🚀 

Agent: https://github.com/frappe/agent/pull/58


press/api/spaces.py
```diff
@@ -0,0 +1,59 @@
...
+@protected("Code Server")
+def start_code_server(name):
+	frappe.get_doc("Code Server", name).start()
+
+
+@frappe.whitelist()
+@protected("Code Server")
+def code_server_password(name):
+	code_server = frappe.get_doc("Code Server", name)
+	return code_server.get_password("password")
```
NagariaHussain: get_doc is heavy and seems unnecessary here, can be replaced with get_decrypted_password instead?

NagariaHussain (changes requested):

press/api/spaces.py
rutwikhdev: shoot

PR #969, opened 2023-07-26
chore: clean-up get_currrent_team

saurabh6790:


press/utils/__init__.py
```diff
@@ -19,65 +19,81 @@ def log_error(title, **kwargs):
...
-	# get team passed via request header
 	team = frappe.get_request_header("X-Press-Team")
-	user_is_press_admin = frappe.db.exists(
-		"Has Role", {"parent": frappe.session.user, "role": "Press Admin"}
-	)
 
-	if (
-		not team
-		and user_is_press_admin
-		and frappe.db.exists("Team", {"user": frappe.session.user})
```
shadrak98, line 43: Removing this check will throw error if team doesn't exists with given user?

press/press/doctype/team/team.py
```diff
@@ -240,10 +240,14 @@ def on_update(self):
 		self.validate_payment_mode()
 		self.update_draft_invoice_payment_mode()
 
-		if not self.is_new() and self.billing_name and not frappe.conf.allow_tests:
+		if not self.is_new() and self.billing_name \
+			and not frappe.conf.allow_tests and not self.is_team_owned_by_admin():
```
ankush, line 244: this conf key should never be used in code? :thinking: It just means if frappe can run tests or not.

PR #1008, opened 2023-08-22
Support ARM Machines and Builds

athul:
- [x] Support ARM Builds
- [x] Support ARM setups
- [x] Support ARM machines in Public benches


adityahase (changes requested):
Don't set this implicitly. Catch this when people add an ARM server to the Release Group.

athul:
Basically the default wkhtmltopdf is 0.12.6 as per 3c8d525c8aeb07fa060f1a1e0baa98014f4fbe38

press/press/doctype/release_group/release_group.py
```diff
@@ -257,19 +257,39 @@ def create_deploy_candidate(self, apps_to_ignore=None) -> "DeployCandidate":
...
-				"dependencies": dependencies,
-				"packages": packages,
-				"environment_variables": environment_variables,
-			}
-		).insert()
-
-		return candidate
+		platforms = []
+		for server in set(server.server for server in self.servers):
+			platforms.append(frappe.get_value("Server", server, "architecture"))
```
ankush:
```suggestion
			platforms.append(frappe.db.get_value("Server", server, "architecture"))
```

press/api/bench.py
```diff
@@ -563,10 +563,11 @@ def deploy(name, apps_to_ignore=[]):
...
 		frappe.throw("A deploy for this bench is already in progress")
 
-	candidate = rg.create_deploy_candidate(apps_to_ignore)
-	candidate.deploy_to_production()
+	candidates = rg.create_deploy_candidate(apps_to_ignore)
+	for candidate in candidates:
+		candidate.deploy_to_production()
 
-	return candidate.name
+	return candidates[0].name
```
ankush: why not return all of them? This feels like a weird return type to me.

press/api/bench.py
```diff
@@ -563,10 +563,11 @@ def deploy(name, apps_to_ignore=[]):
 	if rg.deploy_in_progress:
 		frappe.throw("A deploy for this bench is already in progress")
 
-	candidate = rg.create_deploy_candidate(apps_to_ignore)
-	candidate.deploy_to_production()
+	candidates = rg.create_deploy_candidate(apps_to_ignore)
```
ankush:
```suggestion
	candidates = rg.create_deploy_candidates(apps_to_ignore)
```

It should be renamed to indicate current state now.

press/api/bench.py
```diff
@@ -607,8 +608,7 @@ def deploy_and_update(name, apps_to_ignore=[], sites=[]):
 def create_deploy_candidate(name, apps_to_ignore=[]):
 	rg: ReleaseGroup = frappe.get_doc("Release Group", name)
 	candidate = rg.create_deploy_candidate(apps_to_ignore)
-
-	return candidate
+	return candidate[0]
```
ankush, line 639: This should ideally return all candidates and consumer code should be updated to handle it.

press/playbooks/roles/mysqld_exporter/tasks/main.yml
```diff
@@ -10,6 +10,14 @@
     src: https://github.com/prometheus/mysqld_exporter/releases/download/v0.12.1/mysqld_exporter-0.12.1.linux-amd64.tar.gz
     dest: /tmp
     remote_src: yes
+  when: ansible_architecture == "x86_64"
+
+- name: Download MySQL Exporter Archive
+  unarchive:
+    src: https://github.com/prometheus/mysqld_exporter/releases/download/v0.12.1/mysqld_exporter-0.12.1.linux-arm64.tar.gz
```
ankush: These are 3 years old! we should update these at some point.

press/press/doctype/bench_update/bench_update.py
```diff
@@ -36,11 +36,12 @@ def deploy(self, apps_to_ignore):
...
 		rg: ReleaseGroup = frappe.get_doc("Release Group", self.group)
-		candidate = rg.create_deploy_candidate(apps_to_ignore)
-		candidate.deploy_to_production()
+		candidates = rg.create_deploy_candidate(apps_to_ignore)
+		for candidate in candidates:
+			candidate.deploy_to_production()
 
 		self.status = "Running"
-		self.candidate = candidate.name
+		self.candidate = candidates[0].name
```
ankush: This will be very confusing when there can be more than 1 candidate but it will always link to first one.

Bad data structure to represent this IMO.

press/press/doctype/deploy/deploy.py
```diff
@@ -23,24 +23,26 @@ def create_benches(self):
...
-					"server": bench.server,
-					"group": self.group,
-					"candidate": self.candidate,
-					"workers": 1,
-					"staging": self.staging,
-					"environment_variables": environment_variables,
-				}
-			).insert()
-			bench.bench = new.name
+			server_arch = frappe.get_value("Server", bench.server, "architecture")
```
ankush:
```suggestion
			server_arch = frappe.db.get_value("Server", bench.server, "architecture")
```

press/press/doctype/deploy/deploy.py
```diff
@@ -23,24 +23,26 @@ def create_benches(self):
...
+			if server_arch == candidate.architecture:
+				new = frappe.get_doc(
```
ankush:
```suggestion
				new_bench = frappe.get_doc(
```

bad variable naming

press/press/doctype/deploy_candidate/deploy_candidate.py
```diff
@@ -403,7 +406,9 @@ def _run_docker_build(self, no_cache=False):
 			f"{settings.docker_registry_url}/{namespace}/{self.group}"
 		)
 
-		self.docker_image_tag = self.name
+		self.docker_image_tag = (
+			self.name + "-arm" if "arm" in self.architecture else self.name
```
ankush, line 410: Why not add suffix regardless of arch? General cases are always better than special cases. There might be 3rd arch someday.

ankush (changes requested):

press/api/bench.py
athul: This is only used for the frontend to go to the newer route. Ref: 

https://github.com/frappe/press/blob/cfd3285be1709974efe49c7a4bf0f4dc090703d3/dashboard/src/components/AlertBenchUpdate.vue#L112-L129

ankush: Frontend can pick first?

Backend should ideall return all candidates that were created right?

PR #1054, opened 2023-09-23
fix: changes for new partner portal

shadrak98:
This PR introduces the following changes:

- Frappe Partnership Date and Customer Partnership Date in Team and Invoice Dt.
- Customers can now see their approval status and edit their Partnership Date from the dashboard.
- Upon enabling partner privileges, a new invoice will be created from that date which will be used to track partner achievement


ankush:
Be more descriptive than `new changes`, you'll have to search for these changes someday :smile:

PR #1084, opened 2023-10-09
Marketplace Billing Refactor

rutwikhdev:
- [x] Remove prepaid billing
- [x] Save card and change plan instead of prepaid payment
- [x] Misc fixes

This PR should also fix the backward compatibility issue with previous smb plans (starter/small-business)

## After

bill3


adityahase:
Should've written tests :frowning:,  we all agreed! :angry:.

PR #1130, opened 2023-10-30
fix: Serialize agent job callbacks without deadlock

balamurali27:
Problem:
For activities that have multiple agent jobs on different servers (Rename Site), when processing the respective callbacks parallelly can leave the doc in a pending state. This happens as the select isn't done with for_update.

Solution:
- ~~Use for_update for every select in the callback transaction.~~ (2nd point ensures serializability as callback will run only after commit of the other)
- Get lock on resource (common to both jobs) at the beginning of the transaction (before the first set_value actually) to avoid deadlock

Essentially simulate [this](https://mariadb.com/kb/en/set-transaction/#serializable)

### DIFF
--- a/press/press/doctype/agent_job/agent_job.py
+++ b/press/press/doctype/agent_job/agent_job.py
@@ -244,6 +244,7 @@ def poll_pending_jobs_server(server):
 			# Update Job Status
 			# If it is worthy of an update
 			if job.status != polled_job["status"]:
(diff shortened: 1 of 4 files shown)


balamurali27:
Alternative is to simulate READ COMMITTED by doing for_update, commit, for_update , but that wouldn't nicely cleanup jobs when we messup the callback.

PR #1138, opened 2023-11-02
fix: Prevent Gunicorn child workers hangup

ankush:
Problem:

- Prerequisite: long running reqeust is going on.
- `bench restart` is requested.
- supervisord sends sigterm to gunicorn master process.
- gunicorn passes it to all child workers and waits for graceful shutdown (30 seconds)
- supervisord has no chill, and sends sigkill to master process in 10 seconds
- gunicorn master proesss is dead, so now child workers will keep running until they complete request which can take a really long time.
- This entire time the sites will be down.

Fix:
- Explicitly encode default graceful_timeout in config - 30 seconds.
- Make supervisor wait 10 more seconds for gunicorn to do its thing, then only send sigkill.

Detailed analysis: https://gameplan.frappe.cloud/g/engineering/projects/152/discussion/2523

### DIFF
--- a/press/docker/config/supervisor.conf
+++ b/press/docker/config/supervisor.conf
@@ -5,11 +5,13 @@ environment={% for v in doc.environment_variables %}{{v.key}}="{{v.value}}",{% e
 {% endif %}
 
 [program:frappe-bench-frappe-web]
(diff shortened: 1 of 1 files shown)


ankush:
> All modified and coverable lines are covered by tests ✅

:clown_face:

PR #1147, opened 2023-11-07
fix: set `payment_mode` as card if `default_payment_method` is set

nextchamp-saqib:
Some teams had payment mode empty while their default payment method was set
So this is a potential fix


shadrak98 (approved):
LGTM

PR #1164, opened 2023-11-13
feat(Site): alerts for version upgrades and site migrations

BreadGenie:
image

image


BreadGenie:
Might stick with just alerts for scheduled and running upgrades/migrations. Failed upgrades/migrations can be a notification and maybe same goes for successful upgrades/migrations

michellealva:
@BreadGenie use the same font case for both.

BreadGenie:
> @BreadGenie use the same font case for both.

Both the alerts have the default font though. Might be the length of the text that's making it look different?

michellealva:
I mean use consistent capitalisation of words

NagariaHussain:
The first one is sentence case (only T is capital), while the second is mixed case.

PR #1175, opened 2023-11-16
fix: Handle stripe webhook timeout

rutwikhdev:
Problem:
The methods that handles invoice creation for prepaid credit times out. But we do create an invoice and allocate amount. 
https://github.com/frappe/press/blob/fd266750425d827f207e5563f3a2724b821ca7ed/press/press/doctype/team/team.py#L994
After a while stripe tries to resend webhook until it gets `200 success response`. This creates multiple allocations for the same payment attempt.

This change doesn't crate a duplicate invoice/allocation and simply returns if the amount has been allocated in previous timed out webhook for the same payment_intent/attempt

### DIFF
--- a/press/press/doctype/team/team.py
+++ b/press/press/doctype/team/team.py
@@ -968,6 +968,12 @@ def process_stripe_webhook(doc, method):
 		process_micro_debit_test_charge(event)
 		return
 
+	if frappe.db.exists(
+		"Invoice", {"stripe_payment_intent_id": payment_intent["id"], "status": "Paid"}
+	):
+		# ignore creating if already allocated
+		return
+
 	team: Team = frappe.get_doc("Team", {"stripe_customer_id": payment_intent["customer"]})
 	amount = payment_intent["amount"] / 100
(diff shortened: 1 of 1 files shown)


press/press/doctype/team/team.py
```diff
@@ -968,6 +968,12 @@ def process_stripe_webhook(doc, method):
 		process_micro_debit_test_charge(event)
 		return
 
+	if frappe.db.exists(
```
ankush, line 971: This is fine but you should add unique validation so this doesn't happen anywhere else in future.

Is there ANY use case for duplicate intent id?

rutwikhdev: > Is there ANY use case for duplicate intent id?

Not really, only reason I didn't add this earlier because we have non-unique records and we need to return 200 even if succeeded webhook is resent. Also stripe_payment_intent_id is only set in prepaid flow.

Added a custom validation anyway if somebody tries to mess up from desk https://github.com/frappe/press/commit/9f39b9a46ee6c58eca22cb3d786f871584c8ac5f

PR #1178, opened 2023-11-17
fix: don't create invoice if new customer

shadrak98:
Enabled partner privileges fails for new customers (with no invoice)


shadrak98:
ignoring the failing test as it seems unrelated.

PR #1182, opened 2023-11-18
fix: negative balance in credit transfer cases

nextchamp-saqib:
Problem:
- Purchase 10$ credit (+ve balance transaction)
- Transfer 5$ (-ve balance transaction)
- Create and finalize a 10$ invoice
- Invoice gets finalized even though the team had only 5$ credit left
- Also balance becomes -5$ after the invoice is finalized


shadrak98 (approved):
LGTM!

Issue #1002, opened 2023-08-17
Support for `gthread` worker


ankush:
- [x] audit potential thread unsafe code in framework. 
- [x] feature flag for changing worker type to gthread
- [x] field for specifying threads per worker.
- [ ] benchmarks with real usage data.

PRs:
- https://github.com/frappe/frappe/pull/22088
- https://github.com/frappe/frappe/pull/22086
- https://github.com/frappe/frappe/pull/22097

PR #1218, opened 2023-12-03
feat: configurable gunicorn worker types

ankush:
Allow configuring worker type on release group:
- Sync or unspecified - Synchronous workers handling 1 request per
  process at a time.
- gthread - Each process having specified numbers of threads. Each
  process can handle same number of requests as thread at same time.
  This worker type requires configuring thread count per worker. IMO 2-4
  threads per worker seem to give best outcome.
  Note: Python GIL limitation applies so concurrency is ONLY for I/O
  and non-python CPU work.
- gevent - Green threads which patches low level IO libraries to
  magically implement cooperative scheduling.

In general:
- Sync is safe and well tested.
- Gthread is largely safe but there can be some thread unsafe code which
  has possibility of breaking
- Gevent is completely untested and requires some modification before it
  can be used in prod safely.

agent: https://github.com/frappe/agent/pull/67

---

:warning: I don't have test setup that builds and runs images. I checked the generated configuration and visually it seems fine.

~~It would be great if our docker build is actually validated in CI. @balamurali27 :grimacing:~~ Added a test here https://github.com/frappe/press/pull/1219

Closes https://github.com/frappe/press/issues/1002

### DIFF
--- a/press/docker/config/supervisor.conf
+++ b/press/docker/config/supervisor.conf
@@ -5,7 +5,8 @@ environment={% for v in doc.environment_variables %}{{v.key}}="{{v.value}}",{% e
 {% endif %}
 
 [program:frappe-bench-frappe-web]
-command=/home/frappe/frappe-bench/env/bin/gunicorn --bind 0.0.0.0:8000 --workers 2 --timeout 120 --graceful-timeout 30 --worker-tmp-dir /dev/shm frappe.app:application --preload --max-requests 5000 --max-requests-jitter 1000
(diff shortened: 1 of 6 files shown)


18alantom:
Additional context regarding use of `gevent`: https://github.com/gevent/gevent/issues/1407

`frappe.local` uses [`werkzeug.local.Local`](https://werkzeug.palletsprojects.com/en/3.0.x/local/#werkzeug.local.Local) which is a wrapper around Python [`contextvars.ContextVar`](https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar). To use `gevent` a custom `frappe.local` implementation wrapping [`gevent.contextvars.ContextVar`](https://www.gevent.org/api/gevent.contextvars.html#gevent.contextvars.ContextVar) would have to be used.

PR #1240, opened 2023-12-07
fix: do not update billing address while app is in installation

saurabh6790:


shadrak98:
Let's remove frappeio dependency entirely? ... We don't need it for Press on Premise and local setup.

PR #1255, opened 2023-12-13
refactor: set threads as bench config

18alantom:
Threads was previously readonly on Bench. Changing it from the Release Group caused updates to supervisor.conf and subsequently the Docker image.

Since agent has its own supervisor.conf, it's better to just update worker count there since that's what's used in prod. This allows on the fly changing of thread count (similar to worker count) using an agent job.

In colab with https://github.com/frappe/agent/pull/69

### DIFF
--- a/press/press/doctype/release_group/release_group.json
+++ b/press/press/doctype/release_group/release_group.json
@@ -43,9 +43,6 @@
   "column_break_9efq",
   "merge_all_rq_queues",
   "merge_default_and_short_rq_queues",
-  "gunicorn_config_section",
-  "gunicorn_worker_type",
-  "gunicorn_threads_per_worker",
   "saas_tab",
   "saas_bench",
   "column_break_26",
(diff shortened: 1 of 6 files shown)


press/press/doctype/release_group/release_group.json
```diff
@@ -302,28 +299,10 @@
...
-  {
-   "default": "sync",
-   "fieldname": "gunicorn_worker_type",
-   "fieldtype": "Select",
-   "label": "Gunicorn Worker Type",
-   "options": "\nsync\ngthread"
-  },
-  {
-   "description": "Setting this to non-zero value will convert sync to gthread worker. This doesn't apply to gevent. ",
-   "fieldname": "gunicorn_threads_per_worker",
```
ankush, line 320: Umm, we still need this on release group... We can't configure individual benches for experimentation. 

I needed it so I can enable it on internal benches permanently and eventually some public benches.

18alantom: https://github.com/frappe/press/pull/1278

PR #1262, opened 2023-12-14
fix: disable code-server by default

18alantom:
Enabling adds 30s and 500MB to image build.

https://github.com/frappe/agent/pull/70


ankush:
@rutwikhdev review this once? :eyes:

rutwikhdev:
@18alantom looks good. Maybe you could've added the flags on team and bench instead of release group, deploy candidate and bench

PR #1263, opened 2023-12-14
feat: refactor permissions

nextchamp-saqib:
- Better permission structure that replaces multiple permission documents (created for each doctype, doc, method) into a single JSON field
- Bulk restrict all documents
- Allows setting only group-wise permission and not user-wise (`Press User Permission` will be removed)
- Doesn't break the older permission system
- Permissions are applied in a `permissive` manner (if any perm-group allows a method, then it is allowed regardless of any restriction in other perm-group)
- Use `controller` methods instead of custom API methods (that are mostly redundant)
- Check controller method permissions without `@protected` decorator
 
Preview:

CleanShot 2023-12-16 at 14 02 02

CleanShot 2023-12-16 at 14 01 54

CleanShot 2023-12-16 at 14 01 34


press/api/site.py
```diff
@@ -69,25 +69,19 @@ def wrapper(wrapped, instance, args, kwargs):
...
-					)
-					name = frappe.db.get_value(doctype, name, "group") if doctype == "Bench" else name
-					doctype = "Release Group" if doctype == "Bench" else doctype
-					restricted_method = frappe.db.exists(
+					is_method_restrictable = frappe.db.exists(
 						"Press Method Permission", {"method": request_path}
 					)
-					has_permission_set = frappe.db.exists(
-						"Press User Permission", {"user": frappe.session.user}
-					)
```
netchampfaris, line 86: does this change the behaviour of `@protected` in old dashboard?

nextchamp-saqib: No, it won't. I have cross checked the logic multiple times

PR #1277, opened 2023-12-19
fix(DatabaseServer): Config changes to mitigate oom

balamurali27:
- chore(MariaDBVariable): Add max_connections and max_user_connections
- fix(DatabaseServer): Reduce buffer pool size (3%)

### DIFF
--- a/press/press/doctype/database_server/database_server.py
+++ b/press/press/doctype/database_server/database_server.py
@@ -603,7 +603,7 @@ class DatabaseServer(BaseServer):
 		self.add_mariadb_variable(
 			"innodb_buffer_pool_size",
 			"value_int",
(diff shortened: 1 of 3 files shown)


press/press/doctype/database_server/database_server.py
```diff
@@ -603,7 +603,7 @@ def adjust_memory_config(self):
 		self.add_mariadb_variable(
 			"innodb_buffer_pool_size",
 			"value_int",
-			int(self.ram_for_mariadb * 0.685),  # will be rounded up based on chunk_size
+			int(self.ram_for_mariadb * 0.65),  # will be rounded up based on chunk_size
```
ankush, line 606: We can go even lower for safety (0.6-0.62 IMO). On most servers bufferpool miss ratio is <2% which is good enough.

Maybe monitor it for few days after change?

PR #1284, opened 2023-12-20
feat: apply team filters in `get_list`

nextchamp-saqib:


BreadGenie:
closing since I don't think we need this anymore

PR #1286, opened 2023-12-21
perf: git-pull simple app updates if possible

18alantom:
Depending on the update diff of a Frappe app, it might only require a git pull and restart for the new code to work properly.

PR checks the diff and updates image build such that for git-pullable updates previously deployed image cache is used and new changes are pulled.

As opposed to running get-app with new hash and causing a layer change causing all downstream layers to be rebuilt.

---

### Notes

**Trade-off**: Locally, computing file diffs between old and new repos takes ~0.5s if both repos have been previously cloned.

Diffing repos (or hashes), afaik cannot be performed on remote and so the source must be present locally.

If pull update is possible then both repos are used for the build, and so they don't need to be re-cloned. If it isn't then old repo is not used for the DC being built.

**What's going on:**
1. Before deploy candidate build (`pre_build`):
    1. Previously deployed hashes are fetched if pull updateable (`_update_app_releases`).
        1. Last deployed hashes are fetched from last deployed bench (`get_pull_update_dict`). 
        2. File diff is fetched between current and last deployed hash (`get_changed_files_between_hashes`).
            1. Repos are cloned for both hashes if not present.
        3. Diff is checked if pull update is possible (`can_pull_update`).
    2. `dc.apps`'s `release` and `hash` are set to previously deployed.
    3. New hash and release values are set on `pullable_hash` and `pullable_release`.
2. Additional build steps are added if pullable updates present (`add_build_steps`).
3. Pullable repos are copied to build context (`_prepare_build_context`).
4. Additional lines are added to Dockerfile to fetch and checkout from pullable repo (`generate_dockerfile`).

(written cause I returned after a break with paltry recollection of what's going on)

⚠️ **Help Needed**: The [`can_pull_update`](https://github.com/frappe/press/pull/1286/files#diff-106107a6838035a917100bbcf5971d4841f7719914ed84d3b4ee04014fc99cffR1003-R1023) function checks changed file list to determine if just a git-pull will suffice. This function could be more accurate.

---

Todo:
- [x] Diff against already cloned repository if it is present (fetching every time bad).
- [x] Update DC apps with old hashes.
- [x] Add git pull to Dockerfile with new hashes.
- [x] Add feature flag to enable this.
- [x] See if it works.
    - [x] Have zero diff between mounted repos of subsequent runs.
- [x] Invert `can_pull_update` check.


press/press/doctype/deploy_candidate/deploy_candidate.py
```diff
@@ -778,6 +780,78 @@ def get_dependency_version(self, dependency):
...
+			return False
+
+		# Requires updated dependencies
+		elif file.endswith("pyproject.toml"):
+			return False
+
+		# Potential change in post get-app fs
+		elif file.endswith("setup.py"):
+			return False
+	return True
```
ankush: Just thinking out loud:

1. Python files - no problems. All apps are installed in editable mode. 
2. pyprojtect/setup.py change - must rebulid image from scratch. Dependencies updated.
3. package.json/yarn.local - must rebuild image from scratch. Dependencies updated
4. css/js/html/less/saas etc - will require `bench build` at least to ensure assets are correct. Maybe a bench build should be issued regardless of what files are changed to ensure we don't miss asset changes (?) Apps also have completely separate asset bundling script. So we have NO way of knowing if assets need rebuild.

ankush (changes requested):
IMO this check should be reversed.

- Assumes image needs to be rebuilt.
- Identify if image doesn't need rebuilding - just pull only changes are done - python file changes or just asset changes in JS/CSS/SASS/LESS/HTML


Also introduce this as a feature flag first? "Delta Builds"

ankush:
@18alantom 

similar logic here: https://github.com/frappe/press/blob/404a6d03b17f8df3f961d909285d1f808725154b/press/press/doctype/app_release_difference/app_release_difference.py#L63-L85

press/press/doctype/deploy_candidate/deploy_candidate.py
```diff
@@ -778,6 +780,78 @@ def get_dependency_version(self, dependency):
...
+
+			pull_update[app_name] = {"old": release.hash, "new": update_hash}
+		return pull_update
+
+
+def can_pull_update(changed_files: list[str]) -> bool:
+	for file in changed_files:
+		# Requires database migration
+		if file.endswith(".json") and "doctype" in file:
+			return False
```
ankush: Something that might require database migration doesn't really require image rebuild?

18alantom: Right, I conflated new site with bench init.

press/press/doctype/deploy_candidate/deploy_candidate.py
18alantom: > So we have NO way of knowing if assets need rebuild.

I was thinking of blacklisting most JS cause of this.

We need to standardize where the to-be-built front end code lies. Right now we ship all of `node_modules` and to-be-built code in the image which is something I want to get rid of. For all of (our) apps using a build step, these are not required.

press/press/doctype/deploy_candidate/deploy_candidate.py
```diff
@@ -778,6 +782,87 @@ def get_dependency_version(self, dependency):
...
+			return {}
+
+		bench_name = benches[0]["name"]
+
+		# db.sql used cause get_list wasn't working as expected
+		bench_apps = frappe.db.sql(
+			"SELECT `app`, `source`, `release`, `hash` FROM `tabBench App` WHERE parent=%s",
+			(bench_name),
+			as_dict=1,
+		)
```
NagariaHussain: Try `db.get_all`?

18alantom:
Failing Black test is unrelated.

18alantom:
@adityahase I can only auto merge, you (or someone else with perms) will have to merge it.

PR #1341, opened 2024-01-12
feat: add simple vimrc to image

18alantom:
Adds a bare-bones `.vimrc`. This enables a few niceties. Some of them:
- sensible tab behavior and display
- line numbers, cursor position display
- syntax highlighting

Some it is from: https://github.com/tpope/vim-sensible/blob/master/plugin/sensible.vim
And some of it has been suggested by Ankush.


ankush:
looks good :+1:

PR #1345, opened 2024-01-12
feat: MariaDB Physical Backups

adityahase:
Reference: [Partial Backup and Restore with Mariabackup](https://mariadb.com/kb/en/partial-backup-and-restore-with-mariabackup/).

Needs https://github.com/frappe/agent/pull/76.

Backup is easy. [Restore is tricky](https://mariadb.com/kb/en/innodb-file-per-table-tablespaces/#importing-transportable-tablespaces-for-non-partitioned-tables).

### DIFF
--- a/press/agent.py
+++ b/press/agent.py
@@ -406,6 +406,21 @@ class Agent:
 			site=site.name,
 		)
 
+	def backup_schema(self, site, offsite=False):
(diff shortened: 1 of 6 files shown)


sagarvora:
Does this guide help?
https://mariadb.com/kb/en/individual-database-restores-with-mariabackup-from-full-backup/#single-node

PR #1378, opened 2024-01-19
fix: restrict fields based on doctype-fields allow list

netchampfaris:
This is my attempt at setting up a permission system for Frappe Cloud which uses Allow approach rather than Restrict.

Every doctype that needs to be fetched on the client must be added in the `ALLOWED_DOCTYPES` list and for each of them a `dashboard_fields` attribute must be added in the doctype controller.

Every doctype action that needs to be available from the client must be added in the `dashboard_actions` attribute in the doctype controller.

This config is used by `press.api.client.*` methods for restricting fields and actions.

Apart from this `press.api.client.get_list` automatically applies `team = $current_team` filter for all select queries if the doctype contains a team field. For doctypes that don't have this field, custom `get_list_query` function is written to add necessary filters.

Also, `press.api.client.insert` sets `team = $current_team` in the doc if the doctype has a team field.

### DIFF
--- a/press/api/client.py
+++ b/press/api/client.py
@@ -106,11 +106,11 @@ def get(doctype, name):
 
 	if not frappe.local.system_user() and frappe.get_meta(doctype).has_field("team"):
 		if doc.team != frappe.local.team().name:
-			frappe.throw("Not permitted", frappe.PermissionError)
+			raise_not_permitted()
 
 	fields = list(default_fields)
-	if hasattr(doc, "whitelisted_fields"):
-		fields += doc.whitelisted_fields
+	if hasattr(doc, "dashboard_fields"):
+		fields += doc.dashboard_fields
 
 	_doc = frappe._dict()
 	for fieldname in fields:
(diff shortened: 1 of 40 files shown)


press/api/client.py
```diff
@@ -4,12 +4,38 @@
 from __future__ import unicode_literals
 import frappe
 import inspect
+from frappe.utils import cstr
 from pypika.queries import QueryBuilder
 from frappe.model.base_document import get_controller
 from frappe.model import default_fields, child_table_fields
 from frappe import is_whitelisted
 from frappe.handler import get_attr, run_doc_method as _run_doc_method
```
ankush, line 12: If we are allowing run_doc_method then it will start allowing all currently whitelisted functions and a lot of them are whitelisted for Frappe team via Desk. Either we need to audit them all or start a new list of whitelisted methods specifically for FC end users like `@press.whitelist()`

ankush (changes requested):

Issue #723, opened 2023-02-23
MariaDB performance schema based reports


adityahase:
[MariaDB performance schema](https://mariadb.com/kb/en/performance-schema-overview/) provides a way to monitor/profile MariaDB runtime performance at the query, transaction, and event level.

The existing monitoring infrastructure doesn't seem a good fit to store this data. All DB Servers run `agent` anyway. Using `agent` to collect this data and store it locally with SQLite seems like a good idea.

Refer to [Sys schema](https://dev.mysql.com/doc/refman/8.0/en/sys-schema-views.html) and [MySQL Workbench performance schema reports](https://dev.mysql.com/doc/workbench/en/wb-performance-reports.html
) for creating useful reports out of the raw event data.

PR #1395, opened 2024-01-29
feat: add support to record performance report of database

tanmoysrt:
Related #723 

**Required** agent PR  https://github.com/frappe/agent/pull/129

**Changes -**
- Added `fetch_performance_report` method in `Database Server` doctype
- Added `Performance Report` doctype which will contain performance schema reports
- Added all the reports (available in mysql workbench) as child table of `Performance Report`
- Anyone can enable/disable specific performance report from `Database Server > Mariadb Settings"
- Added cronjob to fetch reports every 15 minutes

**Preview**

https://github.com/frappe/press/assets/57363826/29ff4bed-484f-4f3e-aedd-0079d419dc63


ankush:
@tanmoysrt awesome stuff!


I've some queries and suggestions. 

1. How much time this takes to fetch from DB and roughly what's size of data per day (assuming 15 min interval)
2. Does this capture `events_statemens_summary_by_digest` or equivalent digest summaries? (found it to be most useful replacement for slow query logs)
3. If you implement summary by digest you'll need to keep truncating the stats and regenerate it on our end using historical data. The digest has some hard limit (10K queries) after which it stops capturing details and stores them against `NULL` digest. 
4. Can we setup log clearing so this doctype and all child tables get cleared at certain interval. Ref: https://github.com/frappe/frappe/pull/17159 see how it's done for email queue. We will need fast clearing in raw SQL without touching ORM.

tanmoysrt:
1. It takes 5~6 seconds to fetch from DB and then couple of seconds to fetch from server to press end. Although it may vary for large database as I was testing on just development setup. Although if we start check Innodb buffer stats, it can slowdown perf schema query + normal queries according to [docs](https://dev.mysql.com/doc/refman/8.0/en/sys-innodb-buffer-stats-by-schema.html) 
2. We are not collecting `events_statements_summary_by_digest` but we can add that. just a little change required. Check this [PR](https://github.com/frappe/agent/pull/80/files#diff-600a2f59df9b34108c5a91387a624c218de01108054837725a0ad6951c3167a0R255-R285), you will find out the sql statements of available reports, that we are fetching from DB
3. Need to check
4. Yup, we can setup log clearing, will check the PR.

tanmoysrt:
After getting some more context on the perf schema, understood that we don't need all the information actually.

Valuable data are available on -
 - [events_statements_summary_by_digest](https://mariadb.com/kb/en/performance-schema-events_statements_summary_by_digest-table/) Most of the views are derived from this table.
     **Some views that might be helpful**
     - https://dev.mysql.com/doc/refman/8.4/en/sys-statement-analysis.html
        SQL : https://mariadb.com/docs/server/ref/mdb/sys/statement_analysis/
     - https://dev.mysql.com/doc/refman/8.4/en/sys-statements-with-full-table-scans.html
        SQL : https://github.com/mysql/mysql-sys/blob/master/views/p_s/statements_with_full_table_scans.sql
 - [table_io_waits_summary_by_index_usage](https://mariadb.com/kb/en/performance-schema-table_io_waits_summary_by_index_usage-table/)
     **Some views that might be helpful**
     - https://dev.mysql.com/doc/refman/8.4/en/sys-schema-unused-indexes.html
        SQL : https://github.com/mysql/mysql-sys/blob/master/views/p_s/schema_unused_indexes.sql

- Also, schema_redundant_index can be helpful to find out those redundant indexes.
   https://github.com/mysql/mysql-sys/blob/master/views/i_s/schema_redundant_indexes.sql

tanmoysrt:
Closing this PR.
Continuing the work in performance insights https://github.com/frappe/press/pull/2361.

PR #1396, opened 2024-01-29
feat(Build): use bench get-app cache 

18alantom:
Use bench get-app cache when running `bench get-app` during build. Related PRs:
- https://github.com/frappe/bench/pull/1519
- https://github.com/frappe/frappe/pull/24412
- https://github.com/frappe/bench/pull/1525
- https://github.com/frappe/bench/pull/1526
- https://github.com/frappe/bench/pull/1527

Since this is bench dependent it will be restricted to bench versions 5.21.3 and up only.

**TODO:**
- [x] Ability to check contents of mounted cache
- [x] Update Dockerfile with new bench get-app command
- [x] Prevent git clone if get-app cache is present
- [x] ~Update build steps~ not required since clone will be marked as cached
- [x] Feature flag on Release Group(?)
- [x] Delete docker image after app cache list check
- [x] Utils to check and delete cache
- [x] If previous deploy present as layer cache, skip app cache
- [x] Add Compress artifacts (hard coded default, else massive tarfiles most above 200MB)
- [x] Layer cache miss cause bind mount has changed (handling of repos in bind mount)
- [x] Does it work?
- [x] Fix CI

# Notes

## Further Improvements
- Caching `bench init` artifacts mainly `frappe` and its dependencies.
- Caching app Python dependencies
- **Reducible Inefficiencies:**
    - Due to bench not handling clone:
        - ~15s to check app cache for every build
        - few seconds to clone repo even if app cache present (else potential layer cache miss)
    - Due to partial build code being in frappe repo
        - node_modules have to be included in cache
        - tar creation and extraction times increase cause of node_modules
            - eg hrms: 305ms to 11,429ms
        - cache space requirement increases cause of node_modules
            - eg hrms: 31MB in 1744 entries to 71MB in 32,198 entries
        - with node_modules compression may be required cause size diff without compression

## Current Caveats
- Due to dependence on `frappe>15.12.0` for build `node_modules` have to be included in the cache.
- Compressing and decompressing the cache takes a lot of time due to inclusion of `node_modules`.
- Python dependencies are not cached and so pip install has to be run, this too is very time consuming.

## Get App Cache in Action

<details>
<summary><b>First Run</b> (cache empty): time to build 9 apps 1780s (~30 minutes)</summary>

- Apps are cached by bench after being built
- Build times include caching times.
- The cache times are high due to inclusion of `node_modules` and compressing cached assets.

one
</details>

<details>
<summary><b>Second Run</b> (cache used): time to build 9 apps 697s (~12 minutes)</summary>

- Wiki added to app list to cause docker layer cache miss.
- Bench get-app cache is used for apps that were present in the first run.
- Install times include pip installation which cannot be cached and decompressing the cache.

two
</details>

Note: without compression, build times are:
- Build 1 (cache empty): 1443s (~24 minutes)
- Build 2 (cache used): 606s (~10 minutes)

## Bench Get App Cache Virtual Doctype

Added to allow viewing cached artifacts, and checking report view. Delete also has been added but since each delete requires building a docker image (~15s) it's preferable to just use docker prune to free space.

Additional functionality can be added later on using [`run_command_in_docker_cache`](https://github.com/frappe/press/pull/1396/files#diff-c0d2c2a8dbae1662aa70e2c9d94525880f190bafe233cc87122931bc04ed5ef2R20)

<details>
<summary><b>Images</b></summary>

Screenshot 2024-02-02 at 18 55 10

Screenshot 2024-02-02 at 18 55 23

</details>


18alantom:
**⚠️ Roadblock ⚠️**
1. Install fails if app fetched from cache due to absent `node_modules`.
2. Bench needs to not delete `node_modules` if app depends on `frappe<15.12.0`.
3. Apps need to be updated with required `frappe` version.
4. Bench needs to be updated to account for points 2. and 3.
5. Need to know version documentation method for 4.

Will be putting it on hold until we've decided a way for apps to document minimum required frappe version.

@ankush

PR #1425, opened 2024-02-07
feat: validate frappe-dependencies

ankush:
Depends on latest bench to actually work: https://github.com/frappe/bench/pull/1534

In absence of command, nothing will be done.

### DIFF
--- a/press/docker/Dockerfile
+++ b/press/docker/Dockerfile
@@ -239,7 +239,7 @@ RUN --mount=type=bind,source=app_updates/{{ app.app }},target=/home/frappe/conte
 
 
 # Validate dependencies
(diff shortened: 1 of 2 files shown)


18alantom:
If command not present:
Screenshot 2024-02-09 at 12 47 21

Valid version:
Screenshot 2024-02-09 at 12 45 45

Invalid version:
Screenshot 2024-02-09 at 12 46 17

- Redirect to `dev/null` doesn't seem to work
- Build doesn't fail if version mismatch (in this case the check is redundant)
- Kinda pollutes build output, but build output is already pretty polluted, so ehh.

18alantom:
Screenshot 2024-02-09 at 13 53 46

`sh` compatible line seems to also not return the exit code, so build still isn't failing

18alantom:
Works as expected. (will merge on CI success)

18alantom:
Merging it, codecov's drunk that line isn't in a conditional and should be covered in a test.

Issue #722, opened 2023-02-22
Update bench version from 5.15.2 to latest


ankush:
diff: https://github.com/frappe/bench/compare/v5.15.2...v5.16.1

Main changes:
- Fixes failures where `__version__` is missing in apps -  https://github.com/frappe/bench/pull/1409
- chdir monkey patch is done earlier for clearing command cache - https://github.com/frappe/bench/pull/1421

ankush:
TODO:

- add new one as option in allowed bench versions
- update all internal RGs
- update saas RG
- update public bench RGs
- make it default in code and update all private bench RGs which are untouched.

ankush:
This update is now even more required because of 2 new features: 
- cache in get-app
- validate deps

18alantom:
Enabled Release Groups on Version 15 and Nightly have been bumped to 5.22.0. They should be using 5.22.0 from the next deploy.

Default bench versions have been set to 5.22.0 in #1430.

PR #1430, opened 2024-02-09
fix: add flags, validations, error messages

18alantom:
- fix: prevent no len of None
- fix: add sensible message if missing apps
- chore: bump bench version for frappe nightly and v15 (addresses https://github.com/frappe/press/issues/722)
- feat: press settings flags for app cache


18alantom:
Merging. Failing Server (1) is due to no space left in GActions Cache. Ignoring CodeCov.

PR #1432, opened 2024-02-11
feat: add support to edit env variables from dashboard

### LINKED ISSUE #1212: Allow configuring environment variables from dashboard UI
- Just like site config, envvars should be configurable here.
- There should be system for setting private env variable not visible on desk but visible on desk (like blacklisted config keys)

image

tanmoysrt:
Fixes #1212

**Changes -**
- Added `internal` checkbox to `Release Group Variable` to hide it from user dashboard
- Added a tab in `bench details` page to configure environment variables. [Added in `dashboard-beta`]
- `update_environment_variable` and `delete_environment_variable` function added to `Release Group` 
- [x] Tests added

**Preview -**

[Screencast from 11-02-24 03:26:34 PM IST.webm](https://github.com/frappe/press/assets/57363826/c8ce4fb8-00b7-4d66-ba5b-e37aa5e0a00f)


dashboard/src2/components/EnvironmentVariablePreviewDialog.vue
BreadGenie, line 1: We don't need a preview for env variables since a preview make sense for bench config as it's equivalent for bench_config.json or common_site_config.json. The same is not the case for environment variables afaik.

dashboard/src2/objects/bench.js
```diff
@@ -534,6 +536,133 @@ export default {
...
+					doctype: 'Release Group Variable',
+					filters: releaseGroup => {
+						return {
+							parenttype: 'Release Group',
+							parent: releaseGroup.name
+						};
+					},
+					orderBy: 'creation desc',
+					fields: ['name'],
+					pageLength: 999,
```
BreadGenie:
```suggestion
```
I've set pageLength to 999 in config tab to avoid an issue where the preview was not complete. We don't need a preview here as mentioned above so let's remove the line.

BreadGenie (changes requested):

dashboard/src2/components/EnvironmentVariablePreviewDialog.vue
tanmoysrt: The required changes has been completed.

BreadGenie (changes requested):
Noticed that internal env variables are added. We should
- show that the env variable is internal in a separate column (a check mark like in Domains tab in Site Details)
- hide the options for edit and delete

BreadGenie:
Or maybe hide the whole thing like we do for config

Also what was the reason for implementing internal enviornment variables?

press/api/tests/test_release_group.py
BreadGenie, line 1: the tests can be moved to the file press/press/doctype/release_group/release_group/test_release_group.py

tanmoysrt: testcases moved to `press/press/doctype/release_group/release_group/test_release_group.py`

press/press/doctype/release_group/test_release_group.py
```diff
@@ -239,3 +240,94 @@ def test_cant_set_min_greater_than_max_workers(self):
...
+		self.assertEqual(len(fetched_environment_variable_list), 2)
+		internal_environment_variables_keys = [
+			env["key"] for env in environment_variables if env["internal"]
+		]
+		non_internal_environment_variables_keys = [
+			env["key"] for env in environment_variables if not env["internal"]
+		]
+		for env in fetched_environment_variable_list:
+			self.assertNotIn(env.key, internal_environment_variables_keys)
+			self.assertIn(env.key, non_internal_environment_variables_keys)
```
tanmoysrt: `get_list` returning 0 records in CI, but when testing locally working correctly .

tanmoysrt: Fixed !
It was due to changee of team in `setUp` method

press/press/doctype/release_group/test_release_group.py
```diff
@@ -239,3 +240,94 @@ def test_cant_set_min_greater_than_max_workers(self):
 	def test_update_available_shows_for_first_deploy(self):
 		rg = create_test_release_group([create_test_app()])
 		self.assertEqual(deploy_information(rg.name).get("update_available"), True)
+
+	def test_fetch_environment_variables(self):
+		rg = create_test_release_group([create_test_app()])
+		frappe.local.team = lambda: frappe.get_doc("Team", rg.team)
```
BreadGenie: instead of overriding the team here we can specify `user=self.team.user` in `create_test_release_group` itself to assign the release group under the test team

tanmoysrt: I tried out to use the user of created team, but while running in CI it's failing, it seems that `self.team` and `frappe.local.team` is different. And the `press.api.client.get_list` use that `frappe.local.team`. as a result `get_list` returning blank list.

Locally working fine !

In CI
image

Locally
image

BreadGenie: are you running tests on your developer instance or a separate test instance?

edit: instance as in site

tanmoysrt: I am running tests with this (ig as in site)
```
bench --site fc.local run-tests --doctype "Release Group"
```

BreadGenie: can you create a new site `fc.test` with press installed and run tests on it? (don't configure anything else, use it purely for testing)

tanmoysrt: ok sure

tanmoysrt: Run in a test site, that also passes all the testcase, team name also same.

image

tanmoysrt: I was testing for this specific doctype, that's why it was not failing. It's failing when we run all the tests together.

Found out the actual issue, in a specific testcase, the `frappe.local.team` method was overwritten, that's why the new testcase was failing.

https://github.com/frappe/press/blob/84637d706b0ea76d367156ffb753124a1c2ad104/press/press/doctype/press_permission_group/test_press_permission_group.py#L125

edit: the issue has been fixed. instead of overriding `frappe.local.team` method, we can override `frappe.local._current_team` variable and make it None at tearDown stage.

BreadGenie (approved):
LGTM

Issue #1437, opened 2024-02-13
UI: random duration for running deploy in deploys list view


NagariaHussain:
![CleanShot 2024-02-13 at 13 14 35@2x](https://github.com/frappe/press/assets/34810212/a6ef4cc7-35e0-4b5c-8f49-1ae36ae5c7d0)

Issue #1413, opened 2024-02-06
d2: Show who triggered deploy and jobs


ankush:
![image](https://github.com/frappe/press/assets/9079960/0ad6bd77-1455-40f4-83b1-1c96932efef4)

PR #1449, opened 2024-02-19
misc fixes on dashboard-2

shadrak98:
- feat: show deployed by columns in deploys list (fixes https://github.com/frappe/press/issues/1413)
- show created_by column in jobs list for both sites and benches (fixes https://github.com/frappe/press/issues/1413)
- fixes https://github.com/frappe/press/issues/1437

### Screenshots:
1.
image

2.
image

3.
image

### DIFF
--- a/press/api/client.py
+++ b/press/api/client.py
@@ -61,7 +61,7 @@ def get_list(
 	if meta.istable and not (filters.get("parenttype") and filters.get("parent")):
 		frappe.throw("parenttype and parent are required to get child records")
 
-	if meta.has_field("team") and not frappe.local.system_user():
+	if meta.has_field("team"):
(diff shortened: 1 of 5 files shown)


press/api/client.py
```diff
@@ -331,6 +331,9 @@ def is_owned_by_team(doctype, docname, raise_exception=True):
 	if not frappe.local.team():
 		return False
 
+	if frappe.local.system_user():
+		return True
```
netchampfaris: this makes the function name return a "wrong" value, why is this needed?

netchampfaris (changes requested):

press/api/client.py
shadrak98: Removed it and added impersonate button instead.

Issue #1305, opened 2024-01-01
feat: hotfix patches on running bench


ankush:
If I have a private bench and there's urgent fix that needs to be deployed I might wanna skip entire update and apply specific fix. 

This is both good and bad. 

The good: Urgent issues get addressed instantly. 
The bad: sanctity of image representing the exact code that's released is gone. So debugging future issue becomes harder if they are related to applied changes. 

Possible impl:

- A dialog for select app and applying patch. Accept a URL e.g. https://github.com/frappe/press/pull/1302.patch 
- `wget URL -o file && git apply patch file`. This can fail if there are conflicts, bail completely in that case as there's nothing we can do about it. Hotfixes should be small and should be applied on relatively recent code (ideally just 1-2 minor version behind)
- bench restart, bench build to apply changes
- if migrate required then let user do it
- store history of actions and maybe display it somewhere that bench isn't "pure" anymore and has patches applied to it. 
- let users undo patches. `git reset --hard original_hash` and bench build/restart.

Note: This is done on running bench (instance) not on release group or DC or any image.

michellealva:
only on desk

PR #1464, opened 2024-02-22
feat: git patch app

18alantom:
TODO:
- [x] Patch App Dialog and option under apps
- [x] Patches list under bench tabs
- [x] Agent job that patches the app
- [x] Add failed status
- [x] Revert single patch
- [ ] Revert all patches
- [ ] Tests
- [ ] Check if patch is applied and sync state
- [x] Does it work
- [x] Agent: https://github.com/frappe/agent/pull/84
- [x] Docs: https://frappecloud.com/docs/benches/app-patches

Adds ability to apply patches to individual apps in a deploy under a release groups.

Details: https://github.com/frappe/press/issues/1305


18alantom:
> ig it mostly will be (because of doctypes) but haven't yet seen the diff in depth.

@ankush There'll be a Doctype called **App Patch** that you can create using [`create_app_patch`](https://github.com/frappe/press/pull/1464/files#diff-58bcf57c4bfdc3d7e383c661ece24c4946890fadb8ddb9268ddf9b33b8e05b9fR76)

**App Patch** itself will have a few methods to apply and revert the patches. Lmk if you need any additional functionality to be there. Scripting it should be fairly straight forward.

18alantom:
I just realized that I'd unknowingly formatted `agent_job_type.json`, hence the div. Forgive me person from the future who's gonna regenerate the fixture.

PR #1472, opened 2024-02-23
feat(config): password/secret support

### LINKED ISSUE #1469: Share encryption key only from backup of site config or on demand
image

This behaviour of showing it by default doesn't make sense.

BreadGenie:
fix #1469

Introduces `Password` type to both Site Config and Common Site Config
Users will be able to set, edit (won't be able to see the value) and delete the key

image
image
image

### Todo
- [ ] create a password field for storing passwords only
- [x] patch to store all keys of type password in password field

### DIFF
--- a/press/patches/v0_7_0/set_password_config_type.py
+++ b/press/patches/v0_7_0/set_password_config_type.py
@@ -1,0 +1,25 @@
+# Copyright (c) 2024, Frappe Technologies Pvt. Ltd. and Contributors
+# For license information, please see license.txt
+
+
+import frappe
+
+
+def execute():
+	secret_keys = frappe.get_all(
+		"Site Config Key", filters={"type": "Password"}, pluck="key"
+	)
+
+	site_config_keys_that_should_be_secret = frappe.get_all(
+		"Site Config", filters={"key": ("in", secret_keys)}, pluck="name"
+	)
+
+	for key in site_config_keys_that_should_be_secret:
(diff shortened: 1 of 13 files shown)


press/patches/v0_7_0/set_password_config_type.py
```diff
@@ -0,0 +1,25 @@
...
+	secret_keys = frappe.get_all(
+		"Site Config Key", filters={"type": "Password"}, pluck="key"
+	)
+
+	site_config_keys_that_should_be_secret = frappe.get_all(
+		"Site Config", filters={"key": ("in", secret_keys)}, pluck="name"
+	)
+
+	for key in site_config_keys_that_should_be_secret:
+		frappe.db.set_value("Site Config", key, "type", "Password")
```
ankush, line 18: This can be done in single query.

press/patches/v0_7_0/set_password_config_type.py
```diff
@@ -0,0 +1,25 @@
...
+
+	common_site_config_keys_that_should_be_secret = frappe.get_all(
+		"Common Site Config", filters={"key": ("in", secret_keys)}, pluck="name"
+	)
+
+	for key in common_site_config_keys_that_should_be_secret:
+		frappe.db.set_value("Common Site Config", key, "type", "Password")
```
ankush, line 25: same with this too

PR #1474, opened 2024-02-23
fix: bump bench to 5.22.1

18alantom:
- fix: bump base bench to 5.22.1
- remove temporary fix
- chore(test): change hrms to raven


18alantom:
Tested build locally with non `frappe/*` app. Test is failing cause GA cache is out of space.

PR #1475, opened 2024-02-23
feat: remote docker image build support

### LINKED ISSUE #1328: Separate build server for benches

tanmoysrt:
Fixes #1328 

Required to merge this PR for agent https://github.com/frappe/agent/pull/83

**Changes -**
- [x] In `Press Settings`, a field to choose remote docker builder server added
- [x] Build process is async and updates based on AgentJob update
- [x] New status tag `Preparing` has been added. In `Preparing` stage it will prepare build context, send that to agent and create the agent job for build.
- [x] if global remote builder configured or release group has remote builder, after `Suspending Builds` new build requests will not be marked as scheduled.
- [x] Option to configure seperate build server (instead of using global one) for a specific release group.

**Preview -**

https://github.com/frappe/press/assets/57363826/a450b50a-abb1-4db1-9d46-091b0633dcb7

**Configuration -**
1. We can configure remote build server from `Press Settings` or from `Release Group`, If we set remote build server for a specific `Release Group` that will be prioritized for that group.

**Flow -**
1. Somone trigerred build.
2. Mark Status as `Preparing`.
3. Prepare build context.
4. Tar (with gzip compression) build directory and upload it to agent.
5. Create a agentjob to start building image.
6. Mark Status as `Running`.
7. On agent job update, build infos will be updated as well.
8. After build finish, agent will push the image in registry.

- On job success, it will check if deploy was requested. if requested it will create the deploy as well
- On job success (on agent side), uploaded build file will be removed as well.


press/press/doctype/deploy_candidate/deploy_candidate.py
```diff
@@ -189,54 +196,190 @@ def _deploy(self, staging=False):
...
+		with open(build_context_archive_filepath, "rb") as f:
+			uploaded_filename = agent.upload_build_context_for_docker_build(f)
+		if not uploaded_filename:
+			raise Exception("Failed to upload build context to remote docker builder")
+
+		settings = self._fetch_registry_settings()
+		agent.build_docker_image(
+			{
+				"deploy_candidate": self.name,
+				"deploy_after_build": deploy_after_build,
```
18alantom, line 300: Some of these flags that are sent to agent don't seem to be used by it. Is this mentioned here for future usage?

For instance this PR removes `_deploy` call from `_build_and_deploy` and passes the `deploy_after_build` flag but deploy is not handled.

So calling build and deploy just builds the image but doesn't deploy it even if remote builder is not being used. 😅

18alantom: > For instance this PR removes `_deploy` call from `_build_and_deploy` and passes the deploy_after_build flag but deploy is not handled.

My bad, this is called on success if not using remote build server.

18alantom (approved):
Only a few minor superficial change requirements, such as function sizes, naming, etc. These have been fixed by me post merging master.

Everything else works and looks good, will merge it soon.

press/press/doctype/deploy_candidate/deploy_candidate.py
tanmoysrt: I need to decouple the parts else need to duplicate the same part 😅

18alantom:
**Server (1)** failing cause of "no space left", I've tested it locally and works, hence merging.

PR #1479, opened 2024-02-23
ci: Make mysql transactions faster

balamurali27:
Sacrifices durability for speed
- **ci: Disable double write buffer**
- **ci: No flushing on commit**
- **ci: Set mariadb buffer pool size to 4G**

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -71,7 +71,7 @@ jobs:
           MYSQL_ROOT_PASSWORD: root
         ports:
           - 3306:3306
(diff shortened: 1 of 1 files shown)


ankush:
https://github.com/frappe/erpnext/pull/25332

I tried this 2 years ago, Don't think it made much difference. 

Do share numbers if the changes help. 😬

ankush:
Ig the main reason was low amount of data and fast enough disks. IIRC the difference was few seconds, attributable to just randomness.

balamurali27:
> Ig the main reason was low amount of data and fast enough disks. IIRC the difference was few seconds, attributable to just randomness.

Yeah, no difference + or - 2 mins

PR #1487, opened 2024-02-26
feat(dashboard-beta): Marketplace

rutwikhdev:


dashboard/src2/objects/marketplace.js
```diff
@@ -0,0 +1,548 @@
...
+			{
+				label: 'Overview',
+				icon: icon('home'),
+				route: 'overview',
+				type: 'Component',
+				component: defineAsyncComponent(() =>
+					import('../components/marketplace/MarketplaceAppAnalytics.vue')
+				),
+				props: app => {
+					return { app: { app: app.doc.name, title: app.doc.title } };
```
netchampfaris: Example of a component that is not well designed. Props that are passed should have a simple structure, preferably primitive values like `String` or `Number`. If an object needs to be passed, try to keep it at one level.

press/press/doctype/marketplace_app/marketplace_app.py
```diff
@@ -382,6 +406,27 @@ def get_payout_amount(self, status: str = "", total_for: str = "net_amount"):
...
+			frappe.qb.from_(site)
+			.left_join(team)
+			.on(team.name == site.team)
+			.left_outer_join(site_app)
+			.on(site.name == site_app.parent)
+			.select(site.name, site.plan, team.user)
+			.where(
+				(site.status == "Active")
+				& (site_app.app == self.app)
+				& (site.plan != "Frappe Team")
```
netchampfaris: 🤦🏻

rutwikhdev: 😆 Resolved

PR #1514, opened 2024-03-05
fix: add global flags, improve error log, etc

18alantom:
- fix: toggle use_cached with use_app_cache
- fix: add global delta build flag
- Also remove success check before deploy (more info? commit message!)


18alantom:
Ignoring `codecov`, cause mostly minor fixes.

Issue #1366, opened 2024-01-16
Site getting served from incorrect bench


ankush:
This happens randomly but sometimes after migration site is still being served from old container. Not sure what's the problem here.

ankush:
This happens when cross-cluster migrate fails and site is not cleared. Happened just now with me.

balamurali27:
> This happens when cross-cluster migrate fails and site dir is not cleared. Happened just now with me.

This case is fixed with [this](https://github.com/frappe/press/issues/1353).

Need to also do https://github.com/frappe/press/issues/910

balamurali27:
fixed by #1537

PR #1537, opened 2024-03-11
fix(SiteMigration): Cleanup site on destination on failure

balamurali27:
But not when site already dropped on source


balamurali27:
fixes #1366

balamurali27:
This makes step 3 (archive in case of retry redundant) Can be removed later

PR #1566, opened 2024-03-14
feat(PrometheusAlertReaction): Add doctype for recording reaction for alert

balamurali27:
- Add tab in Alert Rule for defining rule

### DIFF
--- a/press/press/doctype/prometheus_alert_rule/prometheus_alert_rule.json
+++ b/press/press/doctype/prometheus_alert_rule/prometheus_alert_rule.json
@@ -25,7 +25,13 @@
   "preview_section",
   "alert_preview",
   "column_break_21",
-  "route_preview"
+  "route_preview",
+  "reaction_tab",
+  "reaction_section",
+  "reaction_time",
+  "reaction_method",
+  "column_break_ijmn",
+  "disable_reactions"
  ],
  "fields": [
   {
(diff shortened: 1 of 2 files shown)


press/press/doctype/prometheus_alert_reaction/prometheus_alert_reaction.json
```diff
@@ -0,0 +1,87 @@
...
+   "label": "Alert",
+   "options": "Prometheus Alert Rule"
+  }
+ ],
+ "index_web_pages_for_search": 1,
+ "links": [],
+ "modified": "2024-03-14 17:50:32.653434",
+ "modified_by": "Administrator",
+ "module": "Press",
+ "name": "Prometheus Alert Reaction",
```
rmehta: Should this be tied in to Prometheus - this should be "Monitor Alert Rule"?

ankush:
@balamurali27 don't see any safeguard for how many times this should be allowed to run. 

IMO it should not be more than 1.5-2x the actual size on dedicated servers. On our servers we can keep the threshold slightly higher before manual intervention is required.

press/press/doctype/alertmanager_webhook_log/alertmanager_webhook_log.py
```diff
@@ -69,6 +69,53 @@ def after_insert(self):
 				job_id=f"validate_and_create_incident:{self.incident_scope}:{self.alert}",
 				deduplicate=True,
 			)
+		if frappe.db.get_value("Prometheus Alert Rule", self.alert, "enable_reactions"):
```
saurabh6790: can be cached!

press/press/doctype/prometheus_alert_reaction/prometheus_alert_reaction.json
balamurali27: I've moved parts of this to Prometheus Alert Rule doctype itself. Felt redundant with Press Job doctype

PR #1573, opened 2024-03-15
refactor: create deploy candidate apps

18alantom:
fix issue that causes DC to not update
apps source even if RG app source has changed

occurs if theres a previous deploy with an
old app source


18alantom:
I don't think I'm gonna get back to this.

PR #1584, opened 2024-03-18
fix: reconnect on build_failed error

18alantom:
Probably the lazy way to go about it. On checking mysql error logs, no particular pattern was found that would help debugging the `InterfaceError` that occurs when build status is set to _Failure_ after a long running build failing.

I'd previously suspected that the connection is killed off, but the wait time out is 8 hours, and so this seems implausible. 

Looking at the build in question, it was failed just cause it was long running (by the server script).

### DIFF
--- a/press/utils/__init__.py
+++ b/press/utils/__init__.py
@@ -11,8 +11,10 @@ from urllib.parse import urljoin
 import frappe
 import pytz
 import requests
+import wrapt
 from frappe.utils import get_datetime, get_system_timezone
 from frappe.utils.caching import site_cache
+from pymysql.err import InterfaceError
 
 
 def log_error(title, **kwargs):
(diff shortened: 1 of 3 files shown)


press/utils/__init__.py
```diff
@@ -543,3 +545,15 @@ def poly_get_doctype(doctypes, name):
 		if frappe.db.exists(doctype, name):
 			return doctype
 	return doctypes[-1]
+
+
+def reconnect_on_failure():
+	@wrapt.decorator
```
ankush, line 551: Where are you getting `wrapt` from? :laughing: It's not defined in req.txt or setup.py

press/utils/__init__.py
```diff
@@ -543,3 +545,15 @@ def poly_get_doctype(doctypes, name):
...
+	def wrapper(wrapped, instance, args, kwargs):
+		try:
+			return wrapped(*args, **kwargs)
+		except InterfaceError:
+			frappe.db.connect()
```
ankush, line 556: Also you should add a retry counter here, this can cause infinite spam to DB :sweat_smile:

18alantom: I'm not sure what you mean, this quits after a single retry. If the retry throws, it isn't handled. Am I missing something? 🤔

Do you mean after multiple invocations of the wrapper across different functions? Even then, given that only one of them would run at a time, only two tries would occur.

ankush: Got it, I thought it was recursively calling same function with `except` block

PR #1590, opened 2024-03-19
ci: use frappe/frappe develop branch

BreadGenie:
since we are already using the same in prod


BreadGenie:
what how did it auto merge when the tests failed :0

PR #1597, opened 2024-03-20
feat(minor): option to disable warning banner when init-ing ssh session

phot0n:
image

Can later be extended to add custom message.


phot0n:
i kinda don't like all the `when`'s. is there a better way to do it? something like which can tell if all the tasks need to be run or not? there is an option to add `when` directly in the main playbook (server/database.yml)

ankush:
Why not just allow configuring org name instead (?)

phot0n:
Tbh there is no point of showing that message. The party ssh-ing just wont quit things there after seeing that message right?

PR #1601, opened 2024-03-21
fix: log frappeio conn exception to error log

phot0n:
if the customer renaming failed, the exception was anyway getting logged in error log


phot0n:
closing in favour of https://github.com/frappe/press/pull/1750

PR #1603, opened 2024-03-21
feat: logs for site and bench

### LINKED ISSUE #1593: Logs on new dashboard
Ticket reference 11760

BreadGenie:
fix #1593 

image

image

image

image

### DIFF
--- a/dashboard/src2/components/bench/BenchLogsDialog.vue
+++ b/dashboard/src2/components/bench/BenchLogsDialog.vue
@@ -1,0 +1,109 @@
+<template>
+	<Dialog
+		:options="{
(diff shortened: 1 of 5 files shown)


18alantom:
@BreadGenie in the `BenchLogsDialog` component, the container that actually shows the log (black `pre` region) is unconstrained. This causes the whole component to have larger than viewport dimensions thereby causing full page scroll.

A better way to handle this would be to manage scrolling only in the log section (it'll prevent loss of context w.r.t the dialog too).

BreadGenie:
Makes sense. Resolved in https://github.com/frappe/press/commit/4e021c5b58127c37dc2ac45f73b4389ef81330e4

PR #1616, opened 2024-03-22
chore: set default save_output to 0

18alantom:
- Add boldly worded message to dissuade unless needed


18alantom:
https://github.com/frappe/frappe/pull/21600

PR #1622, opened 2024-03-23
perf: drop modified index to reduce contention

ankush:
Dropped modified index from `Agent Job` and `Agent Job Step`, we don't need these indexes and since these are two most busy tables the index causes unnecessary lock contention. 

TODO:
- [x] mark these as ignored
- [x] Observe and drop (merging this PR should be enough)


ankush:
TODO:
- [ ] mark these as ignored
- [ ] Observe
- [ ] actually drop

PR #1631, opened 2024-03-26
refactor: remote builder

18alantom:
This PR is to be a refactor of https://github.com/frappe/press/pull/1475.

While the previous implementation works, it has a few issues that make it unsuitable for production and long term:
- Improper error handling, if `docker build` fails, press mostly just receives a `CalledProcessError` with irrelevant traceback. One needs to dive into agent logs to figure out what went wrong.
- Divergence from press build code, https://github.com/frappe/agent/pull/83 is almost entirely a duplicate of `deploy_candidate.py` this is not a feasible approach, the code between the two has already diverged.

---

Agent side: https://github.com/frappe/agent/pull/94

Note: this PR is incomplete, remote builder still isn't entirely ready, but since this PR has grown to above 1.2k line changes, I'm merging it as checkpoint.

TODO (will be addressed in a future PR):
- [ ] Proper error handling on agent side build failures (need to test it)
- [ ] Method to clear bench app cache on remote builds
- [ ] Higher freq agent polling for build jobs
- [ ] Report to run command in docker cache
- [ ] Checking if it all works properly

Things completed:
- [x] Refactor handling of build and push output parsing
- [x] Move output parsing from agent to press (prevent duplication and divergence)
- [x] Add remote build related steps
- [x] Improve debugging DX by linking Error Logs on Agent Jobs and Deploy Candidates
- [x] Break remote build into 3 steps (previously it was only 1)
- [x] Ensure that press build works properly


phot0n:
>  Improve debugging DX by linking Error Logs on Agent Jobs and Deploy Candidates

thanks for this :heart: - been so annoyed by this that i had also started doing this.

18alantom:
I am 75% sure the failing tests are unrelated to my PR (they're failing elsewhere too). And so I'm merging it, this really shouldn't be FC workflow, but it is. Will address it soon.

PR #1649, opened 2024-04-01
refactor(NewSite): improved flow

BreadGenie:
image

- Apps are selected first
- Version is auto selected to latest stable version which all selected apps supports
- Region is auto selected to closest region
- rest is the same


NagariaHussain:
"Version Information of Selected Apps" sounds odd to me. Maybe just "Version Compatibility" or similar?

PR #1650, opened 2024-04-01
refactor: refac create_release to suck less

18alantom:
Create release adds a comment on failing and doesn't handle non success error codes correctly. Also the function itself is a bit unwieldy.

This PR:
- mentions app source on the error log
- shows error logs on the app source
- factorizes `create_release` into sensible chunks
- removes creating a comment on errors
- handles non success GitHub API call


18alantom:
Failing test seems to be unrelated, I'm merging it.

PR #1661, opened 2024-04-03
feat: make dashboard-beta default

BreadGenie:
~old dashboard will be available in `/dashboard-old`~ old dashboard won't be accessible due to permissions being different on both the dashboards.


dashboard/vite.config.ts
```diff
@@ -40,7 +40,7 @@ export default defineConfig({
 		rollupOptions: {
 			input: {
 				main: path.resolve(__dirname, 'index.html'),
-				dashboard_beta: path.resolve(__dirname, 'dashboard-beta.html')
+				dashboard_beta: path.resolve(__dirname, 'dashboard.html')
```
netchampfaris: this line can be removed altogether, along with the file "dashboard-beta.html" in dashboard

dashboard/dashboard.html
netchampfaris, line 1: this file is not required

PR #1692, opened 2024-04-13
refactor: `dashboard_whitelist` decorator instead of `dashboard_actions`

BreadGenie:
Replaces `dashboard_actions` list with a `dashboard_whitelist` decorator.

#### Normal methods

Use `@dashboard_whitelist()` decorator instead of `@frappe.whitelist()`

```diff
- dashboard_actions = ['method']

- @frappe.whitelist()
+ @dashboard_whitelist()
  def method():
        ...
```

#### `delete`

Add a dashboard whitelisted delete method

```diff
- dashboard_actions = ['delete']

+ @dashboard_whitelist()
+  def delete():
+    # can add dashboard specific before_delete logic
+    super().delete()
```

#### `set_value`

All fields that are specified in the `dashboard_fields` list can be modified through `set_value`

```diff
+ dashboard_fields = ['field']
```


press/api/client.py
```diff
@@ -289,15 +299,6 @@ def check_team_access(doctype: str, name: str):
 	raise_not_permitted()
 
 
-def check_dashboard_actions(doctype, method):
```
netchampfaris, line 292: should have put the `is method dashboard whitelisted` logic in this function instead of duplicating the logic at multiple places

netchampfaris (changes requested):

PR #1695, opened 2024-04-15
fix: consider before tax total for card invoices

shadrak98:


press/press/doctype/invoice/invoice.py
```diff
@@ -209,6 +209,30 @@ def finalize_invoice(self):
 			self.apply_partner_credits()
 			return
 
+		if self.payment_mode == "Card":
+			try:
+				self.create_stripe_invoice()
```
netchampfaris, line 214: Charging full invoice amount even though the customer has some wallet balance is not a standard behaviour. If this is required for some use-cases it needs to be configurable and not default for everyone.

PR #1707, opened 2024-04-22
Revert "chore(bench-test): skip broken test"

18alantom:
This reverts commit 412b9e2ebc8fa23b03360c542f5ab8e189aba521.

Also, tries to fix the broken test.

### DIFF
--- a/press/api/tests/test_bench.py
+++ b/press/api/tests/test_bench.py
@@ -1,42 +1,42 @@
 import json
 import os
 import time
-from unittest import skip
+import requests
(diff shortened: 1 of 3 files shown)


18alantom:
App Patch test seems to be failing with:

```
ERROR: failed to solve: duplicate cache exports [gha]
```

The previously skipped test seems to be passing, so the revert is valid.

18alantom:
Failing tests seem to be cause of other changes https://github.com/frappe/press/actions/runs/8843984566/job/24285082758#step:15:537

PR #1723, opened 2024-04-26
chore: skip sentry log on log_error

18alantom:
`press` uses internal _overload_ of `frappe.log_error`

A `log_error` call is mostly used to log errors that are expected and may have been handled.
If the error is to be thrown then a `raise` is called after the `log_error`

Due to this, sentry log of a log error is not really useful. This PR disables sentry logging when `log_error` is called.

[frappe reference](https://github.com/frappe/frappe/blob/9e8808f06cf1ca6e0c7014e7220a4e3029bd0dcf/frappe/utils/error.py#L71)


adityahase:
Let's do this later. For now, Sentry does much better aggregation.

We'll come back to this once the overall error rates have reduced.

PR #1736, opened 2024-05-01
fix: invalid http_port

ankush:
`webserver_port` is supposed to be port for accessing site from outside, not gunicorn port. 
This should be left unset to assume 80/443 in prod.

refer: https://github.com/frappe/frappe/pull/26267


ankush:
This is a mess (surprise!)

`webserver_port` - It's used as gunicorn port in bench and http port in frappe. 
`http_port` -

PR #1743, opened 2024-05-04
fix: show Pay Now btn only for card inv

shadrak98:


shadrak98:
Merging this for now. Will add the alternative solution later.

PR #1750, opened 2024-05-07
fix: provision to disable frappe.io auth dependency

saurabh6790:


shadrak98 (approved):

phot0n:
Don't raise exception if the url is not found instead of adding another field?

Also all this logic which connects and does stuff on frappe.io and erpnext.com should probably be moved to another app?

PR #1770, opened 2024-05-13
feat(wip): Simplified Role Permissions

netchampfaris:
cc @BreadGenie


dashboard/src2/components/NavigationItems.vue
```diff
@@ -118,6 +124,7 @@ export default {
 		}
 	},
 	mounted() {
+		this.$session.roles.fetch();
```
netchampfaris: fetching roles in NavigationItems component doesn't feel right, if you have fetch it once then do it in main.js

press/press/doctype/press_role/press_role.py
```diff
@@ -0,0 +1,51 @@
...
+
+		for row in self.users:
+			if row.user == user:
+				self.remove(row)
+				break
+		self.save()
+
+	@dashboard_whitelist()
+	def delete(self):
+		super().delete()
```
netchampfaris, line 74: when role is deleted, the linked role permissions should also get deleted?

dashboard/src2/components/settings/RoleConfigureDialog.vue
```diff
@@ -0,0 +1,140 @@
...
+					</div>
+				</div>
+			</div>
+			<div class="mt-4 rounded border-2 border-dashed p-3 text-base">
+				<div class="mb-2 text-gray-600">Global Permissions</div>
+				<div class="flex flex-col space-y-2">
+					<FormControl
+						type="checkbox"
+						v-model="enableBilling"
+						label="Enable Billing"
```
netchampfaris: "Allow access to Billing"

dashboard/src2/components/settings/RolePermissions.vue
```diff
@@ -81,132 +81,147 @@ function getDropdownOptions(listResource) {
...
+				onClick() {
+					renderDialog(h(RoleConfigureDialog, { roleId: props.roleId }));
+				}
+			},
+			{
+				label: 'Remove',
+				slots: {
+					prefix: icon('trash-2')
+				},
+				disabled: selectedItems.value.size === 0,
```
netchampfaris: Is this button always shown? It should only show when there are selected items

dashboard/src2/components/settings/RolePermissions.vue
```diff
@@ -81,132 +81,147 @@ function getDropdownOptions(listResource) {
...
+				label: 'Configure',
+				slots: {
+					prefix: icon('settings')
+				},
+				onClick() {
+					renderDialog(h(RoleConfigureDialog, { roleId: props.roleId }));
+				}
+			},
+			{
+				label: 'Remove',
```
netchampfaris: "Delete permissions"

dashboard/src2/components/settings/RolePermissions.vue
```diff
@@ -25,54 +28,51 @@
...
+});
+
+const bulkDelete = createResource({
+	url: 'press.api.client.bulk_delete'
+});
+
 const dropdownOptions = [
-	{ label: 'Site Permissions', doctype: 'Site' },
-	{ label: 'Bench Permissions', doctype: 'Release Group' }
+	{ label: 'Allowed Sites', doctype: 'Site', dt: 'site' },
```
netchampfaris: `dt` should be `fieldname`

press/api/client.py
```diff
@@ -115,6 +117,37 @@ def get_list(
...
 
+	if doctype in ["Site", "Release Group", "Server"] and not frappe.local.system_user():
+
+		# check if user has a role and the role has any allowed items
+		if roles := get_permission_roles():
+			if frappe.get_all(
+				"Press Role Permission",
+				fields=["name"],
+				filters={"role": ("in", roles)},
+			):
```
netchampfaris: two queries just to check if filter should be applied or not?

press/api/client.py
```diff
@@ -152,6 +182,34 @@ def get(doctype, name):
...
+			if not any(
+				frappe.get_all(
+					"Press Role",
+					fields=["enable_apps"],
+					filters={"name": ("in", roles)},
+					pluck="enable_apps",
+				)
+			):
+				raise_not_permitted()
+
```
netchampfaris: this block of code is similar to the one in get_list with minor changes, can you figure out a way to commonify?

press/api/client.py
```diff
@@ -229,6 +287,26 @@ def delete(doctype, name):
...
+	if not names:
+		return
+
+	check_permissions(doctype)
+	check_dashboard_actions(doctype, names[0], method)
+
+	permitted_records = []
+	for name in names:
+		try:
+			check_team_access(doctype, name)
```
netchampfaris: should also check if document is allowed via role permission

press/api/client.py
```diff
@@ -115,6 +117,37 @@ def get_list(
 				.where(ParentDocType.team == frappe.local.team().name)
 			)
```
netchampfaris, line 126: also disallow get_list for Team doctype?

press/press/doctype/press_role_permission/press_role_permission.py
```diff
@@ -0,0 +1,43 @@
...
+				"site": self.site,
+				"release_group": self.release_group,
+				"server": self.server,
+			},
+		):
+			frappe.throw("Role Permission already exists")
+
+	@dashboard_whitelist()
+	def delete(self):
+		super().delete()
```
netchampfaris, line 54: should we have a check for who can delete/add role permisions?

press/press/doctype/press_role_permission/test_press_role_permission.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2024, Frappe and Contributors
+# See license.txt
+
+# import frappe
+from frappe.tests.utils import FrappeTestCase
+
+
+class TestPressRolePermission(FrappeTestCase):
+	pass
```
netchampfaris, line 9: let's write some test cases too?

dashboard/src2/pages/Billing.vue
```diff
@@ -5,7 +5,20 @@
 				:items="[{ label: 'Billing', route: { name: 'Billing' } }]"
 			/>
 		</Header>
-		<TabsWithRouter :tabs="tabs" />
+		<TabsWithRouter
+			v-if="
+				$team.doc.is_desk_user ||
+				($session.roles.data || []).some(role => role.enable_billing)
```
netchampfaris: do this computation once in session.js, something like `$session.hasBillingAccess`

netchampfaris (commented):
Let's strive for more polish and completeness. We should have basic tests too.

press/press/doctype/press_role_permission/press_role_permission.py
BreadGenie: Simple approach would be to only allow the team owner. But if we should let others have such admin like privileges maybe we can create special roles for it.

dashboard/src2/components/settings/RolePermissions.vue
BreadGenie: I've went with just "Delete" since adding "permissions" looked a bit cumbersome.

press/api/client.py
BreadGenie: fixed in https://github.com/frappe/press/commit/5b044baa3f5e04f34589f6006c0902434f81faca

press/api/client.py
BreadGenie: it was earlier a single check but it wasn't very readable hence I changed it to this. ref: https://github.com/frappe/press/pull/1770/commits/1d08a8e4e62a699eefc9838f12741db69d9dec0d

press/press/doctype/press_role/test_press_role.py
```diff
@@ -0,0 +1,163 @@
...
+		frappe.set_user("Administrator")
+		site1 = create_test_site(team=self.team.name)
+		site2 = create_test_site(team=self.team.name)
+		self.perm_role.add_user(self.team_user.name)
+		self.perm_role2.add_user(self.team_user.name)
+		frappe.set_user(self.team_user.name)
+
+		# no permissions added should show all records
+		self.assertCountEqual(get_list("Site"), [{"name": site1.name}, {"name": site2.name}])
+		return
```
netchampfaris: unintended return?

BreadGenie: Oops, yes unintended.  Was trying to run that particular case only 😅 .

PR #1772, opened 2024-05-13
feat: Custom oauth logins

rutwikhdev:
Steps:
1. Setup Custom Oauth doctype

image

2. Login page shows button when email domain entered

[Screencast from 2024-05-13 22-07-41.webm](https://github.com/frappe/press/assets/50401596/d6a06d59-fd27-495c-a4d4-bb72b8e535d4)

No UI to setup internally, will add if anyone else requests.


dashboard/src2/data/logins.js
```diff
@@ -0,0 +1,24 @@
+import { createResource } from 'frappe-ui';
```
netchampfaris: file name should be oauth_providers.js?

dashboard/src2/data/logins.js
```diff
@@ -0,0 +1,24 @@
...
+
+let oauthLogins = createResource({
+	url: 'press.api.oauth.get_custom_oauths',
```
netchampfaris: avoid custom endpoints? can this be done using createListResource?

dashboard/src2/pages/LoginSignup.vue
```diff
@@ -165,9 +176,21 @@ export default {
...
-			resetPasswordEmailSent: false
+			resetPasswordEmailSent: false,
+			oauthEmailDomain: null
 		};
 	},
+	watch: {
+		email: {
+			handler: debounce(function (value) {
+				let domain = value.split('@').pop();
+				this.oauthEmailDomain =
```
netchampfaris: oauthEmailDomain can be a computed value

dashboard/src2/pages/LoginSignup.vue
```diff
@@ -218,6 +249,12 @@ export default {
 		}
 	},
 	methods: {
+		isOauthLogin() {
+			return this.oauthEmailDomain === null;
```
netchampfaris: this function name is `isOauthLogin` but it returns `true` only if there is no oauth email domain?

press/press/doctype/custom_oauth/custom_oauth.js
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2024, Frappe and contributors
+// For license information, please see license.txt
+
+// frappe.ui.form.on("Custom OAuth", {
```
netchampfaris: Doctype name should be OAuth Domain Mapping?

dashboard/src2/pages/LoginSignup.vue
```diff
@@ -218,6 +249,12 @@ export default {
...
+		},
+		oauthProvider(emailDomain) {
+			return oauthProviders()[emailDomain].provider_name;
```
netchampfaris: Maybe a better way is to make a helper function in `logins.js` called `getOauthProvider(email)`

netchampfaris (changes requested):
Final implementation is good. Naming of variables, function names and doctype can be improved.

dashboard/src2/pages/LoginSignup.vue
rutwikhdev: fixed

dashboard/src2/data/logins.js
rutwikhdev: @netchampfaris I tried with createListResource initially. But client.get_list is not allowed for guest request. Also the result for this would be same for every user so it's better we fetch from some shared server side cache instead of making db requests? I'm not sure if we can do this with resources api.

netchampfaris: @rutwikhdev In that case, this should be done in LoginSignup.vue. There is already an API call `press.api.account.signup_settings` that fetches data required for login page. Maybe add this there. I don't think it's going to be used anywhere so doesn't make sense to keep it in separate file.

rutwikhdev: Alright. Will get rid of logins.js in that case.

dashboard/src2/pages/LoginSignup.vue
rutwikhdev: @netchampfaris Didn't understand this. I'm trying to set the value of oauthEmailDomain and computed are readonly? 🤔

netchampfaris: The watcher is extracting the email domain from the user inputted `email` value right? This is the perfect use-case for computed value.

dashboard/src2/pages/LoginSignup.vue
```diff
@@ -165,9 +172,20 @@ export default {
...
 			signupEmailSent: false,
-			resetPasswordEmailSent: false
+			resetPasswordEmailSent: false,
+			emailDomain: null
 		};
 	},
+	watch: {
+		email: {
+			handler: debounce(function (value) {
+				this.emailDomain = value.split('@').pop();
```
netchampfaris: emailDomain can also be computed?

rutwikhdev: done

dashboard/src2/pages/LoginSignup.vue
```diff
@@ -259,6 +283,40 @@ export default {
 		hasForgotPassword() {
 			return this.$route.name == 'Login' && this.$route.query.forgot;
 		},
+		emailDomain() {
+			return this.email?.split('@').pop();
```
netchampfaris: this should return `null` or `''` if there's no `@` in `email`

PR #1791, opened 2024-05-20
fix: grammar on site update dialog

nikkothari22:
Could also change this to: "If the update fails, rollback will not occur"

### DIFF
--- a/dashboard/src2/components/SiteUpdateDialog.vue
+++ b/dashboard/src2/components/SiteUpdateDialog.vue
@@ -18,7 +18,7 @@
 							v-model="skipFailingPatches"
 						/>
 						<FormControl
-							label="Skip taking backup for this update (If the update fails, rollback will not be happen)"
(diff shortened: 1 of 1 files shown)


dashboard/src2/components/SiteUpdateDialog.vue
```diff
@@ -18,7 +18,7 @@
 							v-model="skipFailingPatches"
 						/>
 						<FormControl
-							label="Skip taking backup for this update (If the update fails, rollback will not be happen)"
+							label="Skip taking backup for this update (If the update fails, rollback will not happen)"
```
BreadGenie:
```suggestion
							label="Skip taking backup for this update (If the update fails, rollback will not occur)"
```
(adding suggestion and committing, so I can avoid pulling your PR to my local)

BreadGenie: Also thanks for catching the mistake 😅

PR #1806, opened 2024-05-24
fix: add build retries

18alantom:
- [x] add retry handling
- [x] add retry cases
- [x] add dashboard banner informing about retry


18alantom:
Semgrep breaking for unrelated reasons

PR #1816, opened 2024-05-28
fix: give read access to sites for support agents

BreadGenie:
- also add `has_role` method to check if role exists

### DIFF
--- a/press/api/client.py
+++ b/press/api/client.py
@@ -14,6 +14,7 @@ from frappe.model import child_table_fields, default_fields
 from frappe.model.base_document import get_controller
 from frappe.utils import cstr
 from pypika.queries import QueryBuilder
+from press.utils import has_role
 
 ALLOWED_DOCTYPES = [
 	"Site",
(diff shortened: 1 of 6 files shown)


press/utils/__init__.py
```diff
@@ -137,6 +137,13 @@ def _system_user():
 	)
 
 
+def has_role(role, user=None):
+	if not user:
+		user = frappe.session.user
+
+	return frappe.db.exists("Has Role", {"parent": user, "role": role})
```
netchampfaris:
```suggestion
	return frappe.db.exists("Has Role", {"parenttype": "User", "parent": user, "role": role})
```

Has Role is a child table for multiple doctypes. (`/app/doctype?fieldtype=Table&options=Has+Role`)

press/api/client.py
```diff
@@ -299,6 +304,9 @@ def check_team_access(doctype: str, name: str):
 	if frappe.local.system_user():
 		return
 
+	if has_role("Press Support Agent") and doctype in ALLOWED_DOCTYPES_FOR_SUPPORT:
+		return
+
```
netchampfaris, line 309: this shouldn't be here? it changes the meaning of `check_team_access` function. Or rename the function to `check_document_access`

netchampfaris (changes requested):

PR #1835, opened 2024-06-02
refactor: grouped actions

BreadGenie:
### Site Actions 

image

### Bench Actions (new)

image

### Server Actions

image


muchai:
We could use shutdown server action.

PR #1841, opened 2024-06-05
fix: handle more build failure cases

18alantom:
- add retry on redis refused (error 111)
- add yarn build failed UA
- fix pull update failure on dirty repos
- add retry on background job timeout
- retry or fail stuck build and deploy jobs
- catch all UA for repository not found issues
- retry on fail to compute cache key


18alantom:
Failing test appears to be unrelated to this PR, will fix it later.

PR #1883, opened 2024-06-18
fix: handle more build failure cases

18alantom:
Check commit messages.


18alantom:
Failing tests unrelated

PR #1887, opened 2024-06-19
feat: Introduce stripe mandate for indian customers

shadrak98:
### Issue:
Lots of Indian customer on Frappe Cloud were facing the issue that their card were not being auto deducted for the Invoice payment even after having sufficient balance. Turns out most of these card errors were due to **missing mandate**.

#### What is a Mandate?
Mandates are consent given by customers to the service provider to auto deduct their card for the authorised invoices.
It will allow service providers to deduct the customers card for the generated invoice without customers intervention. However, there is a threshold of 15000 INR upto which service provider can charge the card without customer's intervention. Any invoice above 15k Inr should require an extra authorisation mechanism only after which the card will be charged.

### How does this PR help:
For new Indian customers, mandates will be created automatically henceforth whenever they add a card. For existing customers, users have to remove and add their card again.

image

image

image

- [Stripe Docs about Mandate](https://support.stripe.com/questions/background-on-indian-government-regulations-affecting-card-payments)
- [RBI Notice](https://www.rbi.org.in/Scripts/NotificationUser.aspx?Id=11668)


press/utils/billing.py
```diff
@@ -126,16 +126,35 @@ def get_publishable_key():
...
+		hash = random_string(10)
 		intent = stripe.SetupIntent.create(
 			customer=customer_id,
 			payment_method_types=["card"],
-			payment_method_options={"card": {"request_three_d_secure": "automatic"}},
+			payment_method_options={
+				"card": {
+					"request_three_d_secure": "automatic",
+					"mandate_options": {
+						"reference": f"Mandate-team:{team}-{hash}",
```
shadrak98, line 145: reference name can be better. I don't like this much. Need suggestion.

netchampfaris: according to docs, `reference` should be unique string to identity this mandate. so random hash is not useful.
we could just use team_id with amount?

https://docs.stripe.com/api/setup_intents/create#create_setup_intent-payment_method_options-card-mandate_options-reference

dashboard/src2/components/AlertMandateInfo.vue
```diff
@@ -0,0 +1,34 @@
...
+				class="ml-auto shrink-0"
+				@click="showAddCardDialog = true"
+				variant="outline"
+			>
+				Update Card
+			</Button>
+		</AlertBanner>
+		<StripeCardDialog
+			v-model="showAddCardDialog"
+			message="Due to RBI regulations, we need to collect mandate/consent from Indian customers to process auto debit from your card. Please add your card again. Contact support@frappe.io for further details."
```
netchampfaris:
```suggestion
			message="Due to RBI regulations, we need to collect mandate/consent from Indian customers to process auto debit from your card. Please add your card again. Contact support for further details."
```

netchampfaris:
@shadrak98 we should also show mandate information in the billing UI

press/utils/billing.py
shadrak98: since the mandate is being created on `SetupIntent` and not during invoice creation we cannot use amount as it is not determined and the maximum amount limit would be same for all the teams.

PR #1900, opened 2024-06-25
feat(Server): add-on storage

BreadGenie:
Screenshot 2024-06-25 at 12 08 13 PM

https://github.com/frappe/press/assets/63963181/364d7644-befa-40ae-97b2-8fd5262628e3

- [x] Increasing disk size from UI
- [x] Billing
- [x] Patch to add subscription for existing users
- [x] Invoice view

### DIFF
--- a/press/press/doctype/server/server.py
+++ b/press/press/doctype/server/server.py
@@ -77,6 +77,13 @@ class BaseServer(Document, TagHelpers):
 
 		return doc
 
+	@dashboard_whitelist()
+	def increase_disk_size_for_server(self, server: str, increment: int) -> None:
+		if server == self.name:
+			self.increase_disk_size(increment)
+		else:
+			frappe.get_doc("Database Server", server).increase_disk_size(increment)
+
 	@staticmethod
 	def on_not_found(name):
 		# If name is of a db server then redirect to the app server
--- a/press/press/doctype/server/server.py
+++ b/press/press/doctype/server/server.py
@@ -64,6 +64,12 @@ class BaseServer(Document, TagHelpers):
 		from press.api.server import usage
 
 		doc.current_plan = get("Server Plan", self.plan) if self.plan else None
+		doc.storage_plan = frappe.db.get_value(
+			"Server Storage Plan",
+			{"enabled": 1},
+			["price_inr", "price_usd"],
+			as_dict=True,
+		)
 		doc.usage = usage(self.name)
 		doc.actions = self.get_actions()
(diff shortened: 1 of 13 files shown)


press/press/doctype/server/server.py
```diff
@@ -497,10 +507,63 @@ def _create_initial_plan_change(self, plan):
...
+			},
+			["name", "additional_storage"],
+			as_dict=True,
+		):
+			frappe.db.set_value(
+				"Subscription",
+				existing_subscription.name,
+				"additional_storage",
+				increment + int(existing_subscription.additional_storage),
+			)
```
shadrak98, line 559: @BreadGenie why not create a new Subscription record for new purchase?

The problem here in updating the existing subscription record is the plan remains the same. You are not updating the plan that means everytime someone buys additional storage he will be charged the same initial amount.
Correct me if my understanding is incorrect.

press/press/doctype/server/server.py
```diff
@@ -1008,12 +1089,62 @@ def update_subscription(self):
...
+			},
+		):
+			add_on_storage_subscription = (
+				frappe.get_doc("Subscription", add_on_storage_subscription)
+				if add_on_storage_subscription
+				else None
+			)
+
+		if add_on_storage_subscription and add_on_storage_subscription.team != self.team:
+			add_on_storage_subscription.disable()
```
shadrak98: If we need `frappe.get_doc` for disable method then better to call it just before this line.

press/press/doctype/server/server.py
BreadGenie: There'll be only one active storage plan at a time, and it'll only hold the per gb per month rate. The extra disk value is stored in subscription and the rate is computed with that as well

See https://github.com/frappe/press/blob/3a9f707eec9ff55ebda1e48db42ad51141bbd30f/press/press/doctype/subscription/subscription.py#L148-L149

shadrak98 (approved):

rmehta:
@BreadGenie do show the increased amount per day as well. This is an important parameter to make this decision.

Also always get a UI review on such designs. (Happy to review)

BreadGenie:
We do show the info in the plan section (per month though)

Screenshot 2024-06-29 at 11 42 54 AM

Maybe we can show the add-on plan details explicitly as well (without hover). Went this way because it looked cluttered when I tried to show all the details in that small box.

rmehta:
Might be a good idea to show it when then are updating as well.

PR #1910, opened 2024-06-28
fix(ui): adjust columns widths

shadrak98:
### Before:
image

### After:
image


dashboard/src2/pages/BillingInvoices.vue
```diff
@@ -109,13 +109,13 @@ export default {
...
 						},
-						width: 0.5
+						width: 0.8
 					},
 					{
 						label: 'Status',
 						fieldname: 'status',
 						type: 'Badge',
-						width: '150px'
+						width: 0.5
```
BreadGenie: These are better off with a fixed width since badge has a capped width (the longest badge)

PR #1911, opened 2024-06-28
feat: allow invoices download as csv

shadrak98:


NagariaHussain (changes requested):
Why not generate the CSV in the backend and download it in frontend? It is a better approach IMO. Also, API friendly.

PR #1925, opened 2024-07-04
fix: permission conflict between fc server and the build server while preparing build context

tanmoysrt:
The issue arrised while using remote docker builder with a local dev setup on any mac device.

photo_2024-07-05 00 05 42

**Root Cause**
In Remote Builder, we assumed that the non-root user always has uid and gid mapping of 1000:1000. In most of the case, first non-root user has mapping of 1000:1000. And In the dockerfile ([ref](https://github.com/frappe/press/blob/master/press/docker/Dockerfile#L158)) the user got changes to `frappe` user before permforming next operations. So, all the files, folders inside build context should have 1000:1000.

In case of Mac, non-root user has different uid and gid, it writes files/folder with `uid 501` and `gid 21`. So that's the issue here. This same error can be possible for linux system also if it has multiple non-root users in same system

**Solution**
While preparing build context, change ownership of contents to 1000:1000


18alantom (approved):
PR looks fine, you can squash-merge and update prod.

If there's any significant performance impact then you can guard it using `frappe.conf.developer_mode`.

tanmoysrt:
> PR looks fine, you can squash-merge and update prod.
> 
> If there's any significant performance impact then you can guard it using `frappe.conf.developer_mode`.

There is no performance issue while running locally, so merged

PR #1935, opened 2024-07-09
feat: tiny branch

tanmoysrt:
**Goals**

- [x] Only Mumbai, India region (configuration)
- [x] Need to ensure that tiny bench's site doesn't get scheduled only on the benches where larger plan's site is deployed
- [x] HTTP request timeout for 30 seconds (Manually)
- [x] Scheduler tick interval 10 minutes (Manually)
- [x] No `skip auto update`
- [x] Restrict the apps to only frappe apps
- [x] Need to ensure that other plan’s app doesn’t get scheduled to these bench’s server
- [x] In change plan, people can’t downgrade to 5 dollar one
- [x] Redesign the site creation page or maybe create a new page
> [!NOTE]  
> Currently, if user selects a non-allowed app or different region, the plan becomes non-selectable instead of completely removed from UI. It will be better to show the reason to user, why the plan is non-selectable/disabled.
- [x] Add testcases
  - [x] Validation of downgrading to restricted site plan
  - [x] Validation of restriction of disabling auto update for sites in public bench
  - [x] Validation of specific allowed apps for site plan
  - [x] Validation to check if larger plan's site doesn't get deployed to this plan's specific benches/servers
  - [x] Validation to check if these specific plan (tiny plan)'s bench get deployed to only to the selected release group / bench specified in Site Plan

> [!TIP]
> CPU time constraints for background job _**(For Later Version)**_
>    - Track the CPU usage
>    - Custom dequeue strategy to stop picking someone’s task if they hit quota


press/api/site.py
```diff
@@ -706,6 +706,13 @@ def get_site_plans():
 		filters={"document_type": "Site"},
 	)
 
+	for plan in plans:
+		# if it's blank, allow to deploy the site on any cluster
+		release_groups = frappe.db.get_all("Site Plan Release Group", pluck="release_group", filters={"parenttype": "Site Plan", "parentfield": "release_groups", "parent": plan.name})
+		plan.clusters = frappe.db.get_all("Bench", pluck="cluster", filters={"group": ("in", release_groups)})
+
+	return plans
```
NagariaHussain: This means 2 db calls for each plan (O(n)), you can optimise this by using the query builder to join.

press/press/doctype/site/site.py
```diff
@@ -370,6 +378,68 @@ def validate_auto_update_fields(self):
...
+								"parent": self.name,
+							},
+						)
+					)
+				)
+				for app in selected_apps:
+					if app not in allowed_apps:
+						frappe.throw(
+							"In {0}, you can't deploy site with {1} app".format(self.subscription_plan, app)
+						)
```
NagariaHussain: You might want to use an early exist here to make this more readable. Ref: https://gist.github.com/18alantom/d9f0565c0f42d6a71311d4a3093a1331#file-code_readability-md

press/api/site.py
tanmoysrt: Yes, I will optimize the queries after ensuring the flow is working perfectly

press/press/doctype/site/site.py
tanmoysrt: In the code, all the parts should execute based on different condition. We need to check for `allowed_apps` and `release_groups` independently. I didn't find any base case to do early exit.

There are some, where we can. Used frappe.throw instead of returning

dashboard/src2/pages/NewSite.vue
```diff
@@ -557,7 +568,7 @@ export default {
 			let pingTime = 999999;
 			try {
 				let t1 = new Date().getTime();
-				let r = await fetch(`https://${server}`);
+				// let r = await fetch(`https://${server}`); TODO: uncomment before merging
```
BreadGenie, line 571: reminder
this is required for fetching nearby server for the site

press/press/doctype/site_plan_change/site_plan_change.py
```diff
@@ -33,6 +33,14 @@ def validate(self):
 			to_plan_value = frappe.db.get_value("Site Plan", self.to_plan, "price_usd")
 			self.type = "Downgrade" if from_plan_value > to_plan_value else "Upgrade"
 
+		if self.from_plan and self.to_plan and self.type == "Downgrade":
+			if not frappe.db.get_value(
+				"Site Plan", self.to_plan, "allow_downgrading_from_other_plan"
+			):
+				frappe.throw(
+					"Sorry, you cannot downgrade to {} from {}".format(self.to_plan, self.from_plan)
```
BreadGenie, line 41:
```suggestion
					f"Sorry, you cannot downgrade to {self.to_plan} from {self.from_plan}"
```

minor nit

dashboard/src2/pages/NewSite.vue
tanmoysrt: shit! forget to remove.

i actually needed multiple servers but it was a bit tough to setup multiple clusters, that's why created those logically. And as the server doesn't exists in real life , in the frontend it was throwing error, that's why commented temporarily.

press/press/doctype/site_plan_change/site_plan_change.py
tanmoysrt: okay

dashboard/src2/pages/NewSite.vue
tanmoysrt: Fixed https://github.com/frappe/press/commit/385660ea20c6024d851f906de8933cea6667a7d4

PR #1951, opened 2024-07-15
fix: handle 'NoneType' object has no attribute 'strip'

saurabh6790:
Sentry issue id: 18739

### DIFF
--- a/press/press/doctype/team/team.py
+++ b/press/press/doctype/team/team.py
@@ -612,7 +612,7 @@ class Team(Document):
 				"address_line1": billing_details.address,
 				"city": billing_details.city,
 				"state": billing_details.state,
-				"pincode": (billing_details.postal_code).strip().replace(" ", ""),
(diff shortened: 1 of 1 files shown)


press/press/doctype/team/team.py
```diff
@@ -607,12 +607,14 @@ def update_billing_details(self, billing_details):
...
 
+		pincode = billing_details.get("postal_code", "").strip().replace(" ", "")
+
 		address_doc.update(
 			{
 				"address_line1": billing_details.address,
 				"city": billing_details.city,
 				"state": billing_details.state,
-				"pincode": (billing_details.postal_code).strip().replace(" ", ""),
+				"pincode": pincode,
```
shadrak98: no need to create another variable for this since it won't be used again.
```suggestion
				"pincode": billing_details.get("postal_code", "").strip().replace(" ", ""),
```

shadrak98 (changes requested):

PR #1952, opened 2024-07-15
feat : Migrate script external files

arunmathaisk:
In addition to local site migration, you can pass external backup file paths..
That way you do not need to restore the site, to migrate it to FC.

The accepted files are validated using mime-types similar to FC frontend

### DIFF
--- a/press/scripts/migrate.py
+++ b/press/scripts/migrate.py
@@ -270,16 +270,18 @@ def select_site():
 			).strip()
 			if selected_site in available_sites:
 				site_data = available_sites[selected_site]
-				downgrade = is_downgrade(sites_info[selected_site])
-
-				if (not downgrade) or (
-					downgrade
-					and click.confirm(
-						"Downgrading may lead to a broken site. Are you sure you want to do this?"
-					)
-				):
+				global has_external_files
+				if not has_external_files:
+					downgrade = is_downgrade(sites_info[selected_site])
+					if (not downgrade) or (
+						downgrade
+						and click.confirm(
+							"Downgrading may lead to a broken site. Are you sure you want to do this?"
+						)
+					):
+						return site_data
+				else:
 					return site_data
-
 			else:
 				print("Site {} does not exist. Try again ❌".format(selected_site))
 	else:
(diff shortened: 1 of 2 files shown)


press/scripts/migrate.py
```diff
@@ -270,16 +272,18 @@ def select_site():
...
-				if (not downgrade) or (
-					downgrade
-					and click.confirm(
-						"Downgrading may lead to a broken site. Are you sure you want to do this?"
-					)
-				):
+				global has_external_files
+				if not has_external_files:
+					downgrade = is_downgrade(sites_info[selected_site])
+					if (not downgrade) or (
```
18alantom, line 278: This probably can be a single `if` instead of being nested. Expression in the nested `if` is too long. Expressions should not span multiple lines.

Unnecessary indentation could be avoided by early exiting; shifting the else clause above the if.

https://gist.github.com/18alantom/d9f0565c0f42d6a71311d4a3093a1331#code-blocks

press/scripts/migrate.py
```diff
@@ -321,42 +325,84 @@ def is_valid_subdomain(subdomain):
...
-	# take backup
-	files_uploaded = {}
-	print("Taking backup for site {}".format(local_site))
+def take_backup(local_site):
+	print(f"Taking backup for site {local_site}")
 	odb = frappe.utils.backups.new_backup(ignore_files=False, force=True)
+	return [
+		(
+			"config",
+			getattr(odb, "site_config_backup_path", None)
```
18alantom, line 334: use a separate variable for this, if your line is folding it's too long to be an expression

press/scripts/migrate.py
```diff
@@ -321,42 +325,84 @@ def is_valid_subdomain(subdomain):
...
-			),
-			("database", odb.backup_path_db),
-			("public", odb.backup_path_files),
-			("private", odb.backup_path_private_files),
-		]
-	):
-		file_name = file_path.split(os.sep)[-1]
 
+@add_line_after
+def upload_files(files):
```
18alantom, line 344: Since you're authoring the function, add typing

press/scripts/migrate.py
```diff
@@ -321,42 +325,84 @@ def is_valid_subdomain(subdomain):
...
 			print("Cannot create site on Frappe Cloud without all site backup files uploaded.")
-			print("Exitting...")
+			print("Exiting...")
 			sys.exit(1)
-
 	print("Uploaded backup files! ✅")
+	return files_uploaded
+
+
+def external_file_checker(file_path, file_type):
```
18alantom, line 360: This entire function is terrible and looks like the work of a college student.

Like the previous function, add typing in the function signature

press/scripts/migrate.py
```diff
@@ -321,42 +325,84 @@ def is_valid_subdomain(subdomain):
...
+	file_name = os.path.basename(file_path)
+	mime_type, _ = mimetypes.guess_type(file_path)
+	print(mime_type)
```
18alantom, line 363: I suppose your forgot to remove this.

press/scripts/migrate.py
```diff
@@ -321,42 +325,84 @@ def is_valid_subdomain(subdomain):
...
+	if file_type == "database":
+		if not file_name.endswith((".sql.gz", ".sql")) and not file_name.endswith(
```
18alantom, line 365: Too long a conditional, unnecessary nesting.

press/scripts/migrate.py
```diff
@@ -321,42 +325,84 @@ def is_valid_subdomain(subdomain):
...
+			"application/x-gzip",
+			"application/x-sql",
+			"application/gzip",
+			"application/sql",
+		]:
+			raise ValueError("Invalid database backup file")
+
+	elif file_type in ["public", "private"]:
+		if mime_type != "application/x-tar":
+			raise ValueError(f"Invalid {file_type} files backup file")
```
18alantom, line 381: This and the conditional below it don't need nesting, they could've been a single line

press/scripts/migrate.py
```diff
@@ -321,42 +325,84 @@ def is_valid_subdomain(subdomain):
...
+
+	elif file_type == "config":
+		if mime_type != "application/json":
+			raise ValueError("Invalid config files backup file")
 
+
+@add_line_after
+def upload_backup(local_site):
+	files_uploaded = {}
+	if has_external_files:
```
18alantom, line 391: Where is this variable?

18alantom: https://gist.github.com/18alantom/d9f0565c0f42d6a71311d4a3093a1331#avoid-globals

press/scripts/migrate.py
```diff
@@ -368,7 +414,11 @@ def restore_site(local_site):
 	)
 
 	# backup site
-	files_uploaded = upload_backup(local_site)
+	try:
+		files_uploaded = upload_backup(local_site)
+	except Exception as e:
+		print(f"{e}")
```
18alantom, line 420: Why the format string?

press/scripts/migrate.py
```diff
@@ -419,6 +469,7 @@ def create_session():
 def frappecloud_migrator(local_site, frappe_provider):
 	global login_url, upload_url, remote_link_url, register_remote_url, options_url, site_exists_url, site_info_url, restore_site_url, account_details_url, all_site_url, finish_multipart_url
```
18alantom, line 470: Man this all just sucks

press/scripts/migrate.py
```diff
@@ -473,30 +524,60 @@ def executed_from_temp_dir():
...
-	except Exception:
-		frappe_provider = "frappecloud.com"
+	restore_choice = click.prompt(
+		"Do you want to restore from external files? (yes/no)", default="no"
+	)
+	if restore_choice.lower() in ["yes", "y"]:
+		has_external_files = True
+		try:
+			external_db_path = click.prompt("Enter full path to the external database file")
+			external_file_checker(external_db_path, "database")
```
18alantom, line 546: You are using exception handling as a control flow, and as a means to pass validation message. A better way to do it would be to just pass values, you'd even remove a level of indentation that way.

press/scripts/migrate.py
```diff
@@ -473,30 +524,60 @@ def executed_from_temp_dir():
...
 	try:
-		frappe.init(site=local_site)
-		frappe.connect()
-		frappecloud_migrator(local_site, frappe_provider)
+		if not has_external_files:
+			frappe.init(site=local_site)
+			frappe.connect()
+			frappecloud_migrator(local_site, frappe_provider)
+		else:
+			print("heyyyy")
```
18alantom, line 570: you forgot this

PR #1961, opened 2024-07-16
Revert "fix: upgrade to usd 5 from other plan's not allowed"

tanmoysrt:
Reverts frappe/press#1958


tanmoysrt:
Moving the logic to Server Script.

PR #1967, opened 2024-07-17
Dynamic app logos

saurabh6790:


dashboard/src/views/partials/LoginBox.vue
```diff
@@ -27,22 +27,20 @@
 			</div>
 		</div>
 		<div class="absolute bottom-4 z-[1] flex w-full justify-center">
-			<FrappeLogo class="h-4" />
+			<BrandLogo :footer="1" />
```
BreadGenie: not very explicit with using 1
```suggestion
			<BrandLogo :footer="true" />
```

dashboard/src2/components/BrandLogo.vue
```diff
@@ -0,0 +1,39 @@
...
+import FrappeLogo from '@/components/icons/FrappeLogo.vue';
+import { getBrandInfo } from '../data/branding';
+
+export default {
+	name: 'BrandLogo',
+	components: {
+		FCLogo,
+		FrappeLogo
+	},
+	props: ['header', 'footer'],
```
BreadGenie: on 2nd thought this should've been a type prop with options being header and footer. Take a look at [AlertBanner.vue](https://github.com/frappe/press/blob/master/dashboard/src2/components/AlertBanner.vue) file

Then we can do
```
<BrandLogo type="header" />
```

BreadGenie (changes requested):
A minor change needed. Otherwise looks good.

rmehta:
PR description, screenshots? @saurabh6790 - This is not 2014!

PR #1972, opened 2024-07-19
chore: revamp onboarding flow

tanmoysrt:
https://github.com/user-attachments/assets/59c5b0ba-ec72-4570-9b04-0ae5c99d7058


dashboard/src/components/UpdateBillingDetailsForm.vue
```diff
@@ -0,0 +1,105 @@
...
+			@click="$resources.updateBillingInformation.submit"
+		>
+			Update Billing Details
+		</Button>
+	</div>
+</template>
+
+<script>
+import AddressForm from '@/components/AddressForm.vue';
+import { notify } from '@/utils/toast';
```
BreadGenie: We are using toast from vue-sonner instead of the old toast. Can you change this?

BreadGenie: On 2nd look this file seems to be in src folder can you move this to src2?
Having 2 src folders must've confused you. I'll remove the older files soon.

tanmoysrt: okay

tanmoysrt: Moved those pages to src2

PR #1986, opened 2024-07-29
feat: auto complete details in setup wizard

tanmoysrt:
FC will create an additional System User which will have email address and full name of team user. This action will be only performed for SaaS sites.

https://github.com/user-attachments/assets/797a113b-02c7-4940-bfb0-ee0110e333b4

Agent PR - https://github.com/frappe/agent/pull/114

### DIFF
--- a/dashboard/src2/components/Onboarding.vue
+++ b/dashboard/src2/components/Onboarding.vue
@@ -287,7 +287,7 @@ export default {
 		TextInsideCircle
 	},
 	mounted() {
-		if (window.posthog) {
+		if (window.posthog?.__loaded) {
 			window.posthog.identify(this.$team.doc.user, {
 				app: 'frappe_cloud',
 				action: 'onboarding'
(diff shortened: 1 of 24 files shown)


NehaSan16:
This is great. The issue of duplicate fields has been around for quite some time and finally, we fixed it. @tanmoysrt 

Some questions
1. What happens when an old user (old to FC) tries to create a new account through a different app? Do we make them verify their account again in the step 1?
Screenshot 2024-07-30 at 10 17 04 PM

2. Can we create an app login flow, where a user first adds the site name, and then adds their username and password to log in to the site? Adding screenshots below

Step 1: Click on login (on app-level landing page) 
Screenshot 2024-07-30 at 10 18 23 PM

Step 2: Add their site name (also give a reminder that they must have received their app login in an email) 
Screenshot 2024-07-30 at 10 19 31 PM


If the site is archived/suspended. We can show them this message and redirect them to Frappe Cloud
Screenshot 2024-07-30 at 10 20 19 PM

This login flow can be used by all the application users. It simplifies returning user journey.

tanmoysrt:
1. Everytime someone comes through the product trial page, we create a account request and user need to verify his e-mail id every time. In new flow by Faris, only once the user create a FC account and it doesn't ask for user creds again.
2. We can do that flow in new one.

cc @NehaSan16

NehaSan16:
Great. Thanks for the confirmation.

PR #1992, opened 2024-07-30
fix: posthog start session recording error

tanmoysrt:
Related **DASHBOARD-V2-3W8**

The error may be thrown when posthog js SDK is not correctly loaded for any reason (can be network issue, or posthog server issue). After this changes, posthog will try to do session recording/capturing only if posthog is loaded successfully.

### DIFF
--- a/dashboard/src2/components/Onboarding.vue
+++ b/dashboard/src2/components/Onboarding.vue
@@ -287,7 +287,7 @@ export default {
 		TextInsideCircle
 	},
 	mounted() {
-		if (window.posthog) {
+		if (window.posthog && window.posthog.__loaded) {
 			window.posthog.identify(this.$team.doc.user, {
 				app: 'frappe_cloud',
 				action: 'onboarding'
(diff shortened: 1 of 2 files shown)


dashboard/src2/components/Onboarding.vue
```diff
@@ -287,7 +287,7 @@ export default {
 		TextInsideCircle
 	},
 	mounted() {
-		if (window.posthog) {
+		if (window.posthog && window.posthog.__loaded) {
```
BreadGenie, line 290:
```suggestion
		if (window.posthog?.__loaded) {
```

simpler

tanmoysrt: okay

tanmoysrt: Updated https://github.com/frappe/press/commit/9aaa764a2331b7d6450180f0eb2d5ec641132435

PR #2036, opened 2024-08-08
fix: do not process commented out line

arunmathaisk:
Screenshot from 2024-08-08 13-04-03


press/press/doctype/app_release_approval_request/app_release_approval_request.py
```diff
@@ -186,10 +186,16 @@ def _screen_python_files(self):
...
 	def _screen_python_file(self, filename):
+		def is_commented_line(line):
+			stripped_line = line.strip()
+			return stripped_line.startswith("#")
+
 		with open(filename, "r") as ff:
 			lines = ff.read().splitlines()
 		lines_with_issues = []
 		for index, line in enumerate(lines):
+			if is_commented_line(line):
```
18alantom, line 197: why not just `if line.strip().startswith("#"):` (functions fine too, but just wondering why)

PR #2039, opened 2024-08-08
feat: allow select server during bench clone

shadrak98:
image

### DIFF
--- a/dashboard/src2/components/site/SiteChangeBenchDialog.vue
+++ b/dashboard/src2/components/site/SiteChangeBenchDialog.vue
@@ -71,7 +71,8 @@
 					onClick: () =>
 						$resources.cloneGroup.submit({
 							name: site,
-							new_group_title: newGroupTitle
+							new_group_title: newGroupTitle,
+							server: selectedServer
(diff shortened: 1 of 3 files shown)


dashboard/src2/components/site/SiteChangeBenchDialog.vue
```diff
@@ -165,6 +175,22 @@ export default {
 					});
 				}
 			};
+		},
+		serverOptions() {
+			return {
+				url: 'press.api.server.get_list',
```
BreadGenie: Use `get_list` from client.py for fetching data almost always.

press/api/site.py
```diff
@@ -1963,7 +1963,7 @@ def change_group_options(name):
 
 @frappe.whitelist()
 @protected("Site")
-def clone_group(name, new_group_title):
+def clone_group(name: str = None, new_group_title: str = None, server: str = None):
```
BreadGenie: why is the default `None`? 🤔

BreadGenie (changes requested):

PR #2058, opened 2024-08-14
feat: allocate free credits on signup

tanmoysrt:
https://github.com/user-attachments/assets/6d846ae1-c331-4953-858e-4514ed8c76d6

- Allocate free credits, once user has updated the billing address
- If user has already availed saas trial, don't allocate free credis + don't show the offer on FC dashboard for them


tanmoysrt:
Closing as a different approach has been used.
Check https://github.com/frappe/press/pull/2063 for ref.

PR #2077, opened 2024-08-19
feat: auth mechanism for sass site to FC communication

tanmoysrt:
Check the `press/saas/api/readme.md` of pull request for details.

Additional -
- Add these internal site config keys
   - fc_communication_secret
   - subscription

### DIFF
--- a/press/press/doctype/site/site.json
+++ b/press/press/doctype/site/site.json
@@ -79,7 +79,7 @@
   "standby_for_product",
   "column_break_63",
   "hybrid_saas_pool",
-  "saas_fc_communication_secret",
+  "saas_communication_secret",
   "backups_section",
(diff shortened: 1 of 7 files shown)


press/press/doctype/site/site.json
```diff
@@ -596,6 +597,12 @@
    "fieldname": "additional_system_user_created",
    "fieldtype": "Check",
    "label": "Additional System User Created"
+  },
+  {
+   "description": "This key will be used to validate requests from saas site for billing & configuration purpose",
+   "fieldname": "saas_fc_communication_secret",
```
shadrak98: No need to use FC in name. It could simply be `saas_communication_secret` or `saas_secret` to make it more simple.

Issue #545, opened 2022-10-17
Deploy on same bench


rmehta:
Allow a pull-only deploy for fast updates on same bench where there are no schema migrations

ankush:
This is not worth having at cost of immutable images. 👎 


Aditya already mentioned that they just need to figure out the parts which slow down image builds for such cases. Once that time is around 1-5 minute it won't matter much.

rmehta:
> This is not worth having at cost of immutable images. 👎

Why? Religion?

ankush:
Many reasons:

- This will require special handling:
   - If py change then Restart workers so they see new code
   - if JS change then build and relink symlinks (not just `bench build`)
- Takes away the basic idea behind containers. Images don't represent the running version anymore. Just destroying and recreating containers also doesn't work anymore. You'll get rolled back to the previous version.
- What do we show on the dashboard about which commit it's running commit? 

TL;DR: Immutability of images good. Call it religion or whatever.

rmehta:
#1 is trivial - used to be there in Central
#2 we don't need to destroy and recreate containers anyways - and they can be tagged in the model
#3 the last updated commit for sure

TLDR: Religious thinking is not so good. In this case, saves a lot of wasteful processing (?)

Not an urgent feature, just leaving it out there

ankush:
>  is trivial

It's not.

More cases:

1. In shared hosting - What if there is more than one site? Everyone gets updated cause one site wants to pull update on their container? 
2. If we are doing this then why not do it for non-pull builds also? 🥴

rmehta:
> In shared hosting - What if there is more than one site? Everyone gets updated cause one site wants to pull update on their container?

Isn't that a problem with shared containers in general.

> If we are doing this then why not do it for non-pull builds also? 🥴

In pull-only, you don't need migrations

ankush:
> Isn't that a problem with shared containers in general.

It's not currently a problem because whoever wants update just moves to next container.

adityahase:
We can make pull-only deploys faster by doing `git pull` on top of the existing image, during the build process. 

1. Won't change any current flows
1. Will be cache friendly (Since only the last layer that includes the diff needs to be pushed)
1. Should save a lot of time (No `setup requirements`, no `build`)

rmehta:
Reopening this. We should be able run any deploys on same bench.

Canary deploys make sense for benches with large number of sites. For smaller sites / self managed servers - the need to "fast deploy" is paramount to the development process. 

[...]

PR #2079, opened 2024-08-20
feat: inplace updates

18alantom:
This is a bad idea.

Screenshot 2024-09-12 at 13 29 33

Use `docker container commit` for updates that don't span multiple Benches (containers). This allows skipping `docker build`. More details later.

- [x] happy path, if update doesn't fail it will work
- [x] failure handling, if update fails restore last working image (⚠️ not tested ⚠️)
- [x] UI
- [x] Docs: https://frappecloud.com/docs/in-place-updates
- [x] Final test

https://github.com/frappe/agent/pull/121

depends on https://github.com/frappe/press/pull/2098

### DIFF
--- a/press/press/doctype/release_group/release_group.py
+++ b/press/press/doctype/release_group/release_group.py
@@ -8,6 +8,7 @@ from itertools import chain
 from typing import TYPE_CHECKING, List, Optional, TypedDict
 
 import frappe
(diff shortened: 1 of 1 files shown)


18alantom:
https://github.com/frappe/press/issues/545

PR #2096, opened 2024-08-23
feat(experimental): hide sidebar for first time user

tanmoysrt:
Sidebar will be hidden on this conditions -
- if team has no parent team
- if no sites created
- if no payment mode added
- if team doesn't own any marketplace app
- if team user doesn't part of other team

**App selector re-design:**
- **Desktop View**
Screenshot 2024-08-23 at 12 43 51 PM
- **Mobile View**
Screenshot 2024-08-23 at 12 44 27 PM

### DIFF
--- a/dashboard/src2/App.vue
+++ b/dashboard/src2/App.vue
@@ -3,7 +3,7 @@
 		<div class="h-full flex-1">
 			<div class="flex h-full">
 				<div
-					v-if="!$isMobile"
+					v-if="!$isMobile && !$team?.doc?.hide_sidebar"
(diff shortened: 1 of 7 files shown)


BreadGenie:
I think instead of having multiple install button we can choose one of the app (highlight selection) and then a primary button with a create site would be better?

tanmoysrt:
> I think instead of having multiple install button we can choose one of the app (highlight selection) and then a primary button with a create site would be better?

I was thinking the same to have a sticky primary button section instead of this. Later I just add it on all because user need to click once to get to installation page (don't know if it will make any difference).

PR #2159, opened 2024-09-17
refactor(s3): Don't hardcode endpoint and region

cogk:
# Don't hardcode endpoint and region for S3 buckets

There were many references to the `ap-south-1` region and a global implicit reference to AWS. This PR contains refactors of the S3 bucket systems for "offsite backups", "dashboard uploads" and "Backup Bucket doctype" buckets, to avoid vendor lock-in.

I tried my best to make it backwards compatible with what I assume is the configuration on Frappe Cloud, but I can't check.

Details:

* Added region and endpoint fields in _Press Settings_ for the "offsite backups" and "dashboard uploads" buckets
  * Fields are: `offsite_backups_endpoint`, `remote_uploads_region`, `remote_uploads_endpoint`
* Dropped unused fields: `offsite_backups_provider` (replaced by endpoint), `data_40` (typo probably)
* There might be some _**agent**_ stuff left that is not handled correctly (e.g. [retrieve the ENDPOINT_URL param here](https://github.com/frappe/agent/blob/1ff9c051fc3a20275e9eeb47a7b7fe7fab1a4ef6/agent/site.py#L404))
* I did not add unit tests

cc: @gavindsouza :wave:
re: https://github.com/frappe/press/pull/19 https://github.com/frappe/press/pull/22 

---

## For the Frappe team

* There was a hardcoded `uploads.frappe.cloud` URL somewhere, I just kept it unchanged, but the refactor won't be truly finished until a migration path can be found for you.
* I'm not so sure about the other buckets (Backup Bucket doctype), they already had the Endpoint and Region fields, but these were not used everywhere (e.g. not in [delete_s3_file](https://github.com/frappe/press/blob/d37ee1caabfc06ff94edd7c870d0a873a79cae98/press/press/doctype/remote_file/remote_file.py#L261))

---

### AWS S3 (Fallback/Legacy)
* Region: leave empty for `ap-south-1`
* Endpoint: leave empty for automatic endpoint generation by boto3 = AWS S3 (I believe `https://s3.amazonaws.com`)

### AWS S3
* Region: `ap-south-1`
* Endpoint: `https://s3.ap-south-1.amazonaws.com` (possibly a custom DNS name such as `uploads.example.com`)

### OVH
* Region: `de`
* Endpoint: `https://storage.de.cloud.ovh.net`


press/api/site.py
```diff
@@ -1854,28 +1852,26 @@ def get_upload_link(file, parts=1):
 
 @frappe.whitelist()
 def multipart_exit(file, id, action, parts=None):
+	press_settings: "PressSettings" = frappe.get_single("Press Settings")  # type: ignore
+	bucket_name = "uploads.frappe.cloud"  # TODO: Use press_settings.remote_uploads_bucket
```
cogk, line 1856: :eyes:

cogk, line 1870: If the Frappe team can confirm that the current config on Frappe Cloud is indeed
`remote_uploads_bucket = "uploads.frappe.cloud"`

Which I believe is the case:




```suggestion
	bucket_name = press_settings.remote_uploads_bucket
```

NagariaHussain:
A patch could be written to set the value back in Press Settings if migration needs to be automatic.

cogk:
- [ ] Need to migrate `boto3_offsite_backup_session` for audit

PR #2162, opened 2024-09-18
feat: webhook on fc

tanmoysrt:
WIP

Users can configure webhook for different kind of events on Frappe Cloud. This will be useful for those, who want to  build something on top of frappe cloud.

### DIFF
--- a/press/press/doctype/press_webhook/press_webhook.js
+++ b/press/press/doctype/press_webhook/press_webhook.js
@@ -1,0 +1,8 @@
+// Copyright (c) 2024, Frappe and contributors
+// For license information, please see license.txt
+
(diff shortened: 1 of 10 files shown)


tanmoysrt:
For now, just keeping `Site Status Update`, `Site Plan Change` webhook event. Will add more events later as per requirement.

tanmoysrt:
In few places, I have ignored ruff complexity (#noqa: C901 ).
- `process_new_site_job_update`
- `process_rename_site_job_update`

Need to break the function to reduce complexity. This ruff complexity take both - no of conditional statement + depth of condition statement in consideration. So, if some function has long body and many conditional stmt, it still gives high complexity (atleast 9 or 10)

In next PR, I may fix the complexity of those two function

PR #2176, opened 2024-09-23
chore: use ruff for linting and formatting (pt.2)

18alantom:
First attempt at #2169. CI only runs linter and formatter on
changed files. Ruff does not run only on changed files so anyone
who touches a file will have to fix all the issues in it.

Good thing is 782 of the 1318 errors are fixable with --fix

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -20,12 +20,12 @@ concurrency:
 
 jobs:
   format:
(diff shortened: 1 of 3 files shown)


18alantom:
Failing test is unrelated to this PR, seems to have started after https://github.com/frappe/press/commit/23f051400088729f0ff2d72a13fa6694fa803115 @balamurali27

PR #2180, opened 2024-09-24
chore: rename bench

18alantom:
ToDo:
- [x] UI labels, titles, etc
- [x] File and folder names
- [ ] In code references

_"Bench"_ will be renamed such that:
- All user-facing labels will be called _"Bench Group"_
- All code references will either be `group` or `releaseGroup`.

**Why?**: _"Bench"_ on the dashboard refers to the **Release Group** doctype.
This causes immense confusion because there also exists a **Bench** doctype.

The objects of **Bench** are shown UI side (under the Site list), but they
can't be called _"Bench"_ because that already refers to something else.

This ends up being a hinderance when it comes to support, writing documentation,
etc. It is also a hinderance when it comes to development because in dashboard
code **Release Group** objects are referred to both as _bench_ and as _release
group_. This asinine naming causes conceptual-confusion on all sides and we end
up with a kind of tower-of-babel like problem.

Finally, if we are to show **Bench** analytics on the dashboard, we'd be
worsening the problem by using other terms instead of _"Bench"_. So this is a
necessary pre-req.

_Note: as of now `benches/bench-XXXX` route will remain the same._

Notes:
- Avoid use unnecessary use of "Bench", eg: "Scheduling bench deploy" -> "Schedule deploy"
- Not altered: `CreateSiteForMarketplaceApp.vue`, seems to refer to the Doctype SiteGroupDeploy which has `"Bench Deploy *"` as one of its status, not sure what it refers to here.


18alantom:
Broken test unrelated to PR. Will complete the last to-do (renaming in code references, mainly variables) in the next PR. Merging this.

PR #2210, opened 2024-10-08
fix: incorrect github url

arunmathaisk:


press/press/doctype/marketplace_app/marketplace_app.py
```diff
@@ -547,7 +547,7 @@ def site_installs(self):
 
 	@dashboard_whitelist()
 	def listing_details(self):
-		github_repository_url = frappe.get_value("App Source", {"app": self.app}, "repository_url")
+		github_repository_url = frappe.get_value("App Source", {"app": self.app,"team": self.team}, "repository_url")
```
BreadGenie, line 550: Check if it's public as well. What if the team have a fork for testing?

arunmathaisk: Yep. Thanks.
Did not think about it.

PR #2213, opened 2024-10-10
feat!: Restore site config from backup

cogk:
Restore the site_config.json file in the `Restore Site` agent job. I don't know why it was not done in `Restore Site`, but correctly done in `New Site from Backup`. If anyone knows if it's just an oversight or if it's intentional, I'm eager to know. :)

### DIFF
--- a/press/agent.py
+++ b/press/agent.py
@@ -180,6 +180,9 @@ class Agent:
 			"managed_database_config": self._get_managed_db_config(site),
 		}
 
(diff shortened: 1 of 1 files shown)


cogk:
Should probably drop the following keys from the site config before restoring:

- `monitor`
- `redis_cache`
- `redis_queue`
- `redis_socketio`
- `shallow_clone`
- `socketio_port`
- `webserver_port`

PR #2214, opened 2024-10-12
fix(testcase): broken tests in CI

tanmoysrt:
Tests were broken mainly due to changes in framework https://github.com/frappe/frappe/pull/28044
Also, there were few testcases, which was broken due to some recent changes. Those has been fixed.

Few testcases ([ref](https://github.com/frappe/press/pull/2214/files#diff-913e5b306693e7707b39584a192eeccbfd5325c6826617aeb8e1e3e2b04acf5dL75-R177)) added for this site plan override based on type of server (public or dedicated) ([ref](https://github.com/frappe/press/commit/414eb838e0b37698c6e3513dfea591fbefa0d5e6)).

### DIFF
--- a/press/api/tests/test_bench.py
+++ b/press/api/tests/test_bench.py
@@ -8,7 +8,7 @@ import docker
 import frappe
 import requests
 from frappe.core.utils import find
-from frappe.tests.utils import FrappeTestCase, timeout
+from frappe.tests import UnitTestCase, timeout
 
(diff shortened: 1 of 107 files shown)


tanmoysrt:
The deprecation wouldn't fail CI due to this update in framework https://github.com/frappe/frappe/pull/28103
Closing it for now. Re-open, if required to move to latest version.

PR #2218, opened 2024-10-15
feat: Trials for all first site

shadrak98:
All first time users, who signup directly on FC will get to enjoy the first site on 14 day trial plan. Trial plan should be mentioned in press settings `press_trial_plan`, since we can have multiple trial plans for saas, marketplace etc. 

Free Credits will be removed in another PR, once this becomes stable.


dashboard/src2/pages/InstallApp.vue
```diff
@@ -275,17 +279,29 @@ export default {
...
 
 			return {
 				url: 'press.api.marketplace.create_site_for_app',
 				makeParams() {
+					this.sitePlan = this.selectedPlan
+						? this.selectedPlan.name
+						: this.options.public_site_plan;
+
+					if (!this.$team.doc.onboarding.site_created) {
+						this.sitePlan = this.trialPlan.name;
```
tanmoysrt: The `get_trial_plan` function returns the site plan name. So, `this.trialPlan` is already a string.
So, it should be just `this.sitePlan = this.trialPlan` ig.

Else, the site plan is sent as `null` and trial site plan doesn't getting set on the site.

PR #2224, opened 2024-10-16
fix: handle non-fc payments

shadrak98:
Internal ticket #23435


tanmoysrt (approved):
I have tested locally, razorpay payment flow via webhook (for adding prepaid credit) is working fine.
Not able to simulate the cases, where order_id is not available, but as per code it should work.

PR #2195, opened 2024-10-01
feat: add page for Bench (`/benches`)

18alantom:
- [x] Benches sidebar option
- [ ] Benches tabs:
    - [x] Sites
    - [x] Patches
    - [x] Apps
    - [x] Jobs
    - [x] Processes
    - [x] Logs
    - [ ] Bench Config
    - [ ] Analytics
- [ ] Links to Bench page

PR #2225, opened 2024-10-16
fix: filter functionality

18alantom:
Follow up to #2195


18alantom:
Filters have still not been entirely fixed, few filter options are populated using the values in the `listResource`, this is kinda stupid cause on changing the filter, the filter options change.

PR #2226, opened 2024-10-16
fix: add more to `dashboard/bench`

18alantom:
Continuation of #2195


balamurali27:
Would be nice if number of workers can be displayed. Memory limits also. For dedicated server users only. Ideally users should be able to set min and max workers from release group (bench group)

18alantom:
Closing cause of my role change. Won't get to complete this.

PR #2231, opened 2024-10-18
refactor: group analytics, jobs, logs and reports

BreadGenie:


dashboard/src2/components/site/SiteInsights.vue
```diff
@@ -0,0 +1,96 @@
...
+				},
+				{
+					label: 'Reports',
+					value: 'Site Performance Reports',
+					children: [
+						'Site Performance Slow Queries',
+						'Site Performance Binary Logs',
+						'Site Performance Process List',
+						'Site Performance Request Logs',
+						'Site Performance Deadlock Report'
```
shadrak98, line 67: Maybe don't repeat "Site Performance" in report names

BreadGenie: Agreed but these are routes

PR #2259, opened 2024-11-05
fix(RemoteFile): Don't try to delete unavailable S3 file

cogk:
If I'm not mistaken, unavailable files could be removed (like other Log Types) once they become unavailable


ssiyad:
Hello! Is there any update on this? Can you fix the issues?

PR #2261, opened 2024-11-06
fix(HybridSaaSPool): create standby sites for all hybrid pool rules

tanmoysrt:
- fix(HybridPool): check for specific app as per record instead of erpnext app
- fix(HybridPool): dont stop create pooling only one hybrid pool has been replenished


tanmoysrt:
The failed tests are not related to this, so merged.

PR #2282, opened 2024-11-14
feat: log browser

BreadGenie:
agent: frappe/agent#143 (will do this later on as there are some perf issues with it)


adityahase:
Based on the video in https://gameplan.frappe.cloud/g/frappe-cloud/projects/17/discussion/3210/log-browser-updates

- The columns jump around when you switch log levels
- Timestamps look weirdly aligned. Could a monospace font help?

PR #2301, opened 2024-11-21
feat: use bench python to validate python syntax in builds

saurabh6790:
Screenshot 2024-11-21 at 3 40 30 PM

### DIFF
--- a/press/press/doctype/app_release/app_release.py
+++ b/press/press/doctype/app_release/app_release.py
@@ -508,13 +508,16 @@ def check_python_syntax(dirpath: str) -> str:
 
 
 def _get_python_path() -> str:
-	from frappe.utils import get_bench_path
+	try:
+		from frappe.utils import get_bench_path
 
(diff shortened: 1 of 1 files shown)


press/press/doctype/app_release/app_release.py
```diff
@@ -505,8 +491,8 @@ def check_python_syntax(dirpath: str) -> str:
 	- -q: quiet, only print errors (stdout)
 	- -o: optimize level, 0 is no optimization
 	"""
-
-	command = f"python3 -m compileall -q -o 0 {dirpath}"
+	python_exec = _get_python_executable()
```
gavindsouza: `sys.executable` should suffice if you are trying to fetch the interpreter for Press' current runtime. This is assuming `get_bench_path` returns the path of the bench Press is installed on.

> Warning: I don't have a Press setup; was just browsing this...for fun 😄

PR #2314, opened 2024-11-29
chore: capture payment intent id in webhook log

shadrak98:


shadrak98:
Tests are breaking due to [this](https://github.com/frappe/frappe/commit/9edd44de012a885779e4f40226dc008dbfbb4cbc)

Fixed [here](https://github.com/frappe/frappe/pull/28621) for backward compatibility

PR #2373, opened 2024-12-18
perf: prevent caches from frappe.clear_cache

ankush:
This happens because sites are renamed and rename can't reliably guess what
else got modified via queries so it just drops everything.

AFAIK, new sites are renamed so they don't have anything else linking to them.
So it's fine to just... ignore it.

Risks:
- Old site renames -> Might require manual eviction
- Other valid use of `frappe.clear_cache()` can break, use `frappe.cache.flushall()`

I also identified bunch of keys that need to be "persistent", added them all to separate list.

:warning: UNTESTED CODE. Use with caution and be ready to rollback :smile:


press/hooks.py
```diff
@@ -355,3 +355,61 @@
...
+
+# `frappe.rename_doc` erases all caches, this hook preserves some of them.
+# Note:
+# - These are only "most used" cache keys. This lessens the impact of renames but doesn't eliminate them.
+# - Adding more keys here will slow down `frappe.clear_cache` but it's "rare" enough.
+# - This also means that other "valid" frappe.clear_cache() usage won't clear these keys!
+# - Use frappe.cache.flushall() instead.
+persistent_cache_keys = [
+	*__persistent_cache_keys,
+	"agent_job_step_output",
```
ankush, line 379: This one isn't evicted anywhere btw, it can get huge but we have tons of memory so :shrug:

adityahase:
@balamurali27 Please test and merge when you're back.

PR #2394, opened 2024-12-30
fix(ci): set pr comment permissions at job stage

tanmoysrt:


tanmoysrt:
Looks like the issue with this is not related to misconfigured workflow.
It's working on other repo of mine
https://github.com/tanmoysrt/press/pull/3#issuecomment-2565476648

PR #2400, opened 2025-01-01
perf(build): Reduce gzip compression level in packaging build context

tanmoysrt:
By default, tarfile library in python do gzip compression at level 9 ([ref](https://docs.python.org/3/library/tarfile.html#:~:text=tarfile.open()%20accepts%20the%20keyword%20argument%20compresslevel%20(default%209)%20to%20specify%20the%20compression%20level%20of%20the%20file.)). 

`tar` or `gzip` cli use compress level 6 by default ([ref](https://www.gnu.org/software/gzip/manual/gzip.html#:~:text=The%20default%20compression%20level%20is%20%2D6)).

Anything above 5 or 6 compression level can take significant resource and time.

---

Locally, I have cloned 10 frappe apps (430MB) to compress with different gzip level using [tarfile](https://docs.python.org/3/library/tarfile.html). 

- Result of Gzip compression level 9
  ```
   real	0m13.507s
   user	0m13.171s
   sys	0m0.327s
   ```
  **Compressed File Size** - 235MB
- Result of Gzip compression level 5
  ```
   real	0m8.599s
   user	0m8.275s
   sys	0m0.324s
   ```
  **Compressed File Size** - 239MB

Other benchmarks - https://stephane.lesimple.fr/blog/lzop-vs-compress-vs-gzip-vs-bzip2-vs-lzma-vs-lzma2xz-benchmark-reloaded/


ankush:
@tanmoysrt you might wanna add all existing violations to cspell whitelist in one shot, else it will become a continuous annoyance and won't help much.

tanmoysrt:
> @tanmoysrt you might wanna add all existing violations to cspell whitelist in one shot, else it will become a continuous annoyance and won't help much.

Added those words in whitelist.

PR #2452, opened 2025-01-30
feat(devbox): Bootstrap code.

arunmathaisk:


dashboard/src2/components/NavigationItems.vue
```diff
@@ -143,6 +143,13 @@ export default {
 					),
 					disabled: enforce2FA
 				},
+				{
+					name: 'Dev Boxes',
```
BreadGenie, line 147: maybe have this inside dev tools?

PR #2456, opened 2025-02-03
fix(deploy_candidate): Load redisearch in redis_queue

### LINKED ISSUE #2391: Load redisearch module in redis_queue also
We need a persistent instance for things like search indexes. 

Ref: 
https://github.com/frappe/press/blob/3e14845b5a12cc962d5bcb05015310010c7e2d7a/press/docker/config/redis-cache.conf#L11-L13

This needs to be in other file too

tarunps:
Closes #2391


ankush:
LGTM, test it once locally or on staging before merge.

PR #2530, opened 2025-03-18
feat(bare-metal): press on bare metal

AlenVelocity:


AlenVelocity:
Closing in favor of https://github.com/frappe/press/pull/2540

PR #2553, opened 2025-03-25
feat(database):  Mariadb Configuration

tanmoysrt:


ankush:
@tanmoysrt the lower limit on connection seems quite aggressive. 

Each worker thread could have one open connection. So we should consider that as indicator instead of memory. 

50-100 is a better lower bound IMO. 

Instead of lowering connection, we can lower per connection memory usage for smaller servers.

PR #2563, opened 2025-03-27
fix(prm): Use team instead of partnerteam

shadrak98:


dashboard/src2/components/partners/PartnerCertificates.vue
```diff
@@ -122,7 +122,7 @@ export default {
 					};
 				},
 				filters: {
-					partner_team: this.$team.doc.name,
+					team: this.$team.doc.name,
```
BreadGenie, line 125: this isn't needed
the filtering will be done by get_list automatically

PR #2576, opened 2025-04-01
feat: Added process snapshot doctype

Aradhya-Tripathi:
Created snapshot doctype to trigger snapshot (py-spy) on agent server.
Related:
* https://github.com/frappe/agent/pull/178


tanmoysrt:
@Aradhya-Tripathi Keep the doctype name little bit distinguishable, as it can create confusion with `Virtual Disk Snapshot` 😅

Aradhya-Tripathi:
> @Aradhya-Tripathi Keep the doctype name little bit distinguishable, as it can create confusion with `Virtual Disk Snapshot` 😅

Oh will do didn't know about virtual disk

PR #2581, opened 2025-04-05
fix: Reducing work from web worker

Aradhya-Tripathi:
Reducing web worker load enqueuing job payload processing.

### DIFF
--- a/press/api/callbacks.py
+++ b/press/api/callbacks.py
@@ -37,10 +37,11 @@ def validate_server_request(remote_addr: str):
 
 
 def verify_job_id(server: str, job_id: str):
-	return frappe.db.get_value("Agent Job", {"server": server, "job_id": job_id})
+	return frappe.get_value("Agent Job", {"server": server, "job_id": job_id})
 
 
-def handle_job_updates(server: str, job_id: str):
+def handle_job_updates(server: str, job_identifier: str):
+	job_id = job_identifier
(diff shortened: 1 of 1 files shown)


press/api/callbacks.py
```diff
@@ -90,10 +94,9 @@ def callback(job_id: str):
...
-	job = verify_job_id(server, job_id=job_id)
+	job = verify_job_id(server, job_id)
 	if not job:
 		frappe.throw("Invalid Job Id", frappe.ValidationError)
 
-	handle_job_updates(server=server, job_id=job_id)
-	frappe.set_user("guest")
-	return "Success"
+	frappe.enqueue(handle_job_updates, server=server, job_identifier=job_id)
+	frappe.set_user("Guest")
```
tanmoysrt, line 102: Maybe reset user in a final block of try..catch

Something like this > https://github.com/frappe/press/blob/693cd5f8e124318aef7c9786749a91d4871a7e0f/press/api/saas.py#L57-L89

Aradhya-Tripathi: Does frappe.enqueue raise?

tanmoysrt: Not in usual scenario,  but enqueue also can raise error in certain condition 

Also there is verify_job_id function which do db call that can fail

Also before raising the error by frappe.throw you should unset the session user

PR #2582, opened 2025-04-05
fix: Error handling

Aradhya-Tripathi:
- Try catch block in job processing

### DIFF
--- a/press/api/callbacks.py
+++ b/press/api/callbacks.py
@@ -105,8 +105,6 @@ def callback(job_id: str):
 	remote_addr = frappe.request.environ["HTTP_X_FORWARDED_FOR"]
 	server = validate_server_request(remote_addr)
 
-	# TODO: There should be something better.
-
 	# Request origin not authorized to update job status.
 	if not server:
 		frappe.throw("Not permitted", frappe.ValidationError)
--- a/press/api/callbacks.py
+++ b/press/api/callbacks.py
@@ -3,16 +3,12 @@
 from __future__ import annotations
 
 import ipaddress
-import typing
 
 import frappe
 
+from press.agent import Agent
 from press.press.doctype.agent_job.agent_job import handle_polled_job
 from press.utils import log_error
-
-if typing.TYPE_CHECKING:
-	from press.press.doctype.press_settings.press_settings import PressSettings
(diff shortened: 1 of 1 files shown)


press/api/callbacks.py
```diff
@@ -41,42 +42,50 @@ def verify_job_id(server: str, job_id: str):
...
-	)
-	callback.insert()
-	frappe.db.commit()
-
-	handle_polled_job(polled_job=polled_job, job=job)
+	current_user = frappe.session.user
+	try:
+		frappe.set_user("Administrator")
+		job_id = job_identifier
+		server: Server = frappe.get_doc("Server", server)
```
tanmoysrt: For getting the agent object, just use  `Agent(server_name, server_type)`

Then you don't need get_doc

Something like this > https://github.com/frappe/press/blob/master/press/press/doctype/site/site.py#L554


Also, callback can receive from Server, Database Server, Proxy Server right ?
You need to figure out that somehow , else things might fail

press/api/callbacks.py
```diff
@@ -41,42 +42,50 @@ def verify_job_id(server: str, job_id: str):
...
+		agent = server.agent
+		press_settings: PressSettings = frappe.get_doc("Press Settings")
```
tanmoysrt: Use get_value or get_cached_doc instead

press/api/callbacks.py
```diff
@@ -41,42 +42,50 @@ def verify_job_id(server: str, job_id: str):
...
+		callback = frappe.get_doc(
+			{
+				"doctype": "Agent Job Callback",
+				"job_name": job.job_type,
+				"agent_job": polled_job["agent_job_id"],
+				"status": polled_job["status"],
+			}
+		)
+		callback.insert()
+		frappe.db.commit()
```
tanmoysrt: This `handle_job_updates` func is running in background job.
Avoid manual commit then, because `handle_polled_job` func will run next.

tanmoysrt (changes requested):

press/api/callbacks.py
Aradhya-Tripathi: This request should only originate from a build server so we ideally don't need to check will add disclaimer as per discussion.

Issue #2345, opened 2024-12-10
Handle python 3.13 install


balamurali27:
Fails as no distutils package for this version
![image](https://github.com/user-attachments/assets/fc6e486e-02e4-431a-91e8-c94e20f8fe5b)

ankush:
Just remove distutils for 3.13

ankush:
3.12 doesn't work too

PR #2604, opened 2025-04-11
feat(build): Removed distutils for python >= 3.12

Aradhya-Tripathi:
Fixes: #2345

### DIFF
--- a/press/press/doctype/deploy_candidate/deploy_candidate.py
+++ b/press/press/doctype/deploy_candidate/deploy_candidate.py
@@ -20,6 +20,7 @@ from typing import Any, Literal
 
 import frappe
 import frappe.utils
+import semantic_version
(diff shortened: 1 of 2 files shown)


press/press/doctype/deploy_candidate/deploy_candidate.py
```diff
@@ -1133,16 +1135,35 @@ def _set_container_mounts(self):
 			order_by="idx",
 		)
 
+	def get_python_version(self, version: str):
+		"""
+		Supports major.minor version extraction
+		Also supports ambiguous version information (>=;~=)
+		"""
+		if match := PYTHON_VERSION_REGEX.search(version):
```
ankush: Can we use version parser here :eyes: `semantic_version` should work? Treating them as floats is wild IMO :woozy_face:

Issue #2598, opened 2025-04-10
Add Physical Database & Restore method as Site Backup


tanmoysrt:
- [ ] Provide Physical backup as standard Site Backup
- [ ] Don't show physical backup taken for `Site Update`
- [ ] Delete all physical backups on site archival
- [ ] From site doctype, allow to configure backup frequency and time of backup
  - [ ] Logical Backup
  - [ ] Physical Backup
- [ ] Allow to restore database on the same site or on a site which is in same region of snapshot.

PR #2605, opened 2025-04-11
refactor(site-backup): Add Backup Frequency and Physical Backup

tanmoysrt:
Fixes https://github.com/frappe/press/issues/2598

- [x] Provide Physical backup as standard Site Backup
- [x] Don't show physical backup taken for `Site Update`
- [x] Delete all physical backups on site archival
- [x] Allow user to trigger backup
  - [x] With File
  - [x] Logical Backup
  - [x] Physical Backup 
- [x] From site doctype, allow to configure backup frequency and time of backup
  - [x] Logical Backup
  - [x] Physical Backup
  - [x] Update all jobs to use this new format
- [x] In physical backup, ensure site deactivation
- [x] Fix testcases
- [x] Allow to restore physical backup on the same site
- [x] Write patch to move old backup times in new format

**UI Previews -**

**Manually Schedule Backup -**

image

image

**Backup List -**
image

_We will show backup type only if Physical Backup is enabled_

**Support for Custom Backup Frequency -**
image


tanmoysrt:
@BreadGenie  we can't hide the `Physical Backup` column based on the flag available in `Site` doctype.
😅 We can configure automated physical backup but disallow physical backup trigger by end-user.

Without that flag enabled, there can be physical backups.

Anymore changed to a check icon.

tanmoysrt:
@BreadGenie I am leaving the `Site Restore via Physical Backup` frontend for now, will come back to this later.

PR #2614, opened 2025-04-14
perf: Add innodb_stats_persistent_sample_pages to mariadb variables

ankush:
https://mariadb.com/kb/en/innodb-system-variables/#innodb_stats_persistent_sample_pages

20 is way too low, we can easily 10x this and it will barely have any
real cost for us. So 10x-ing this for now.

ref: https://www.percona.com/blog/correcting-mysql-inaccurate-table-statistics-for-better-execution-plan/

We also need to upgrade to 11.x someday soon... when 11.8 is stable.

### DIFF
--- a/press/fixtures/mariadb_variable.json
+++ b/press/fixtures/mariadb_variable.json
@@ -606,5 +606,19 @@
   "set_on_new_servers": 0,
   "skippable": 0,
   "title": null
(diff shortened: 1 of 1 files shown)


tanmoysrt (approved):
looks good

PR #2619, opened 2025-04-17
fix(app-patch): Marking patch as not applied when bench is archived

Aradhya-Tripathi:
Marking patches on obsolete benches as archived.


BreadGenie (approved):
LGTM. Need to write a patch to fix existing patches though.

PR #2642, opened 2025-04-23
fix(notifications): Build failure notification with correct candidate

Aradhya-Tripathi:
Build failure notifications failing and deploy candidate build being passed instead of deploy candidate


Aradhya-Tripathi:
Merging as no tests written for change.

PR #2648, opened 2025-04-23
fix(notifications): Pass deploy candidate build

Aradhya-Tripathi:
Passing deploy candidate build and fetching candidate from there.


Aradhya-Tripathi:
Fixed notifications in https://github.com/frappe/press/pull/2651

PR #2654, opened 2025-04-25
fix(glob): Loop over files with glob post moving to build

Aradhya-Tripathi:
Moving glob clean up to deploy candidate build.


Aradhya-Tripathi:
Merging since not coverage.

PR #2669, opened 2025-04-29
chore(naming): Auto increment name & median

Aradhya-Tripathi:
Getting median instead of average for durations, and moved to auto increment naming.


Aradhya-Tripathi:
No checks for this change!

PR #2680, opened 2025-05-04
fix(retry): Retry count is increased in a new deploy candidate build

Aradhya-Tripathi:
- Fail the current build doctype instead of setting it's status to scheduled (leads to race condition scheduling)
- In the new deploy candidate build increase the retry count by one, so that after 3 retries we stop.

### DIFF
--- a/press/press/doctype/deploy_candidate_build/deploy_candidate_build.py
+++ b/press/press/doctype/deploy_candidate_build/deploy_candidate_build.py
@@ -661,11 +661,17 @@ class DeployCandidateBuild(Document):
 			log_error("Deploy Candidate Build Exception", doc=self)
 
 	def schedule_build_retry(self):
-		self.retry_count += 1
+		"""
+		Fail the current deploy candidate build and create a new deploy candidate build
+		with an increased `retry_count` and a scheduled time.
+		"""
 		minutes = min(5**self.retry_count, 125)
 		scheduled_time = now() + timedelta(minutes=minutes)
-		self.set_status(Status.SCHEDULED)
-		self.candidate.schedule_build_and_deploy(run_now=False, scheduled_time=scheduled_time)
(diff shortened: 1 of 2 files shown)


tanmoysrt (approved):

press/press/doctype/deploy_candidate_build/deploy_candidate_build.py
```diff
@@ -661,11 +661,17 @@ def handle_build_failure(
...
+		"""
 		minutes = min(5**self.retry_count, 125)
 		scheduled_time = now() + timedelta(minutes=minutes)
-		self.set_status(Status.SCHEDULED)
-		self.candidate.schedule_build_and_deploy(run_now=False, scheduled_time=scheduled_time)
+		self.set_status(Status.FAILURE)
+		# Increase the retry count
+		self.candidate.schedule_build_and_deploy(
+			run_now=False, scheduled_time=scheduled_time, retry_count=self.retry_count + 1
+		)
```
tanmoysrt: Maybe here add some hard limit that if retry_count reach > 3, dont even call anything else.

tanmoysrt: Okay that `should_build_retry` function can be used

tanmoysrt: Ignore, it's already handled

Aradhya-Tripathi: Ok

Aradhya-Tripathi: Fixed in #[f373717](https://github.com/frappe/press/pull/2680/commits/f3737171ddd06ea252ce08e3cf973c703d744bb1)

PR #2690, opened 2025-05-11
fix(env-var): Changed deploy candidate env var field from data to text

Aradhya-Tripathi:
This is to accommodate large text values/API keys in `Deploy Candidate`


Aradhya-Tripathi:
Merging since no coverage.

PR #2693, opened 2025-05-12
Fix:  cannot import name '_check_multicommand'

asaura08:
Summary:
This pull request addresses an issue where the bench CLI fails to initialize a new bench due to an incompatibility with the click package. The _check_multicommand import used in bench is no longer available in newer versions of click.

Reason for Change:
Without this fix, running bench init throws an ImportError, breaking the build and preventing Frappe environments from being initialized properly. #https://github.com/frappe/bench/pull/1632


asaura08:
> Dependencies are on press are fetched from a Frappe Versions DocType.
> 
> Which has already been updated in production.

Yes but by default it point to a version that cannot be deployed due to click package version change the "_check_multicommand" method

Aradhya-Tripathi:
It can be merged without an issue.

PR #2696, opened 2025-05-14
fix for dependent app validation check in hooks_uri, also made it con…

0spinboson:
…ditional w/ error

doesn't seem like it should be impossible to save an app source just because the dependency check fails, esp. at this point. That said, main issue was a self.app invocation which should've been self.repository.


Aradhya-Tripathi:
LGTM ⚡

PR #2698, opened 2025-05-14
feat: Install bench into a virtualenv during init

akhilnarang:
This makes the resolver fetch the correct version of urllib3 (part of request's dependencies), otherwise it uses an old version from dist-packages, which tries to use `six`, which is not installed.


akhilnarang:
Bad approach when I think a bit more, need to figure out a safe way to upgrade urllib3 instead

Issue #2703, opened 2025-05-18
new host and related agent calls no longer trigger Nginx reload by default, so newly added domains never become active


engahmed1190:
![Image](https://github.com/user-attachments/assets/3a911bdc-1bdf-4333-bc22-6209e250890f)

**Problem Description**
A recent refactor flipped the default behavior of all proxy-management methods in the Press agent—including `new_host` and `add_domain_to_upstream`—so that they now skip reloading Nginx unless explicitly told otherwise. As a result, when a new custom domain is created or restored:

* The upstream configuration file is written correctly
* **But** Nginx is never reloaded to pick up that change
* The domain remains unregistered in the live upstream block, causing requests to fail

In practice, users must remember to invoke every host/domain change with a “reload” flag or manually restart Nginx—an easy step to miss that leads to broken sites.

---

**Steps to Reproduce**

1. On current `master` (post-PR #2699), add a new Site Domain via the UI or code.
2. Observe that the proxy config file under `sites/.../proxy.conf` contains the new entry.
3. Check Nginx’s process start time or logs—no reload has occurred.
4. Attempt to access the site over the new domain and see that the upstream is not found.

---

**Expected Behavior**
Adding or restoring a domain should automatically reload Nginx so the new upstream entry becomes active immediately.

---

**Actual Behavior**
By default the agent skips the reload step, leaving the new domain unserved until a manual reload is performed.

---

**Root Cause**
A sweeping change set all agent methods to default to “skip reload,” and critical code paths (like site-migration restore steps and the after-insert hook for Site Domain) never override that default.

---

**Proposed Resolution**
Restore the old default behavior—have domain-adding methods reload Nginx automatically—or ensure that any code path which writes a new upstream entry explicitly triggers a reload immediately afterward.

tanmoysrt:
Hi,
Low priority nginx reload tasks will be deferred.
Update agent on your server and enable Nginx Defer Reload (You will find it under `Proxy Server` actions)

Only important tasks will trigger instant reload > https://github.com/frappe/press/pull/2699

phot0n:
some reload manager was introduced as well? quite possibly after this issue was created - can this be closed?

PR #2704, opened 2025-05-19
fix(site): nginx reload on new site creation

engahmed1190:
Solves issue #2703


tanmoysrt:
This should be fixed by new changes
- https://github.com/frappe/press/pull/2718
- https://github.com/frappe/agent/pull/208

PR #2726, opened 2025-05-27
feat(build-server): Create arm build for rg with arm servers

Aradhya-Tripathi:
Currently we have 1 arm build server, which is ok for now since are converting them slowly.

Eventually we shoudn't need this bit of code hoping all app servers are on arm.

### DIFF
--- a/press/press/doctype/deploy_candidate/utils.py
+++ b/press/press/doctype/deploy_candidate/utils.py
@@ -3,9 +3,12 @@ import re
 from collections import Counter
 from datetime import timedelta
 from pathlib import Path
-from typing import Any, Optional, TypedDict
+from typing import TYPE_CHECKING, Any, Optional, TypedDict
 
 import frappe
+
+if TYPE_CHECKING:
+	from press.press.doctype.release_group.release_group import ReleaseGroup
 
 
 class PackageManagers(TypedDict):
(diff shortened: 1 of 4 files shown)


press/press/doctype/deploy_candidate/utils.py
```diff
@@ -142,16 +145,31 @@ def get_build_server(group: str | None = None) -> str | None:
...
-	if server := get_build_server_with_least_active_builds():
+	if group:
+		release_group: ReleaseGroup = frappe.get_doc("Release Group", group)
+		for server in release_group.servers:
+			server_platform = frappe.get_value("Server", server.server, "platform")
+			if server_platform == "arm64" and (server := get_arm_build_server_with_least_active_builds()):
+				return server
+
+	if server := get_x86_build_server_with_least_active_builds():
 		return server
```
tanmoysrt: There might be one issue.

Assume, you have a public release group and you have multiple servers linked with it (some are from AWS, some from OCI). 

And For now you can only move aws servers to ARM, not the OCI ones.

So, in that case you need to build for both platform ?

Aradhya-Tripathi: That's an issue as well, I was thinking how do we handle this but for now I decided to reduce complexity maybe?

tanmoysrt: But, once you start migrating servers, this thing can start causing issues and need to eventually fix if we have x86 and arm servers listed in same release group.

Aradhya-Tripathi: This can get a little complicated there are 19 release groups with more than 3 servers

tanmoysrt: Yeah mostly would be `public` servers for shared hosting. Those can be ignored then in first phase of migration.

Aradhya-Tripathi: Maybe only release groups with one server should be targeted in the first phase then 🤔

Issue #2735, opened 2025-05-29
Suspend Site on crossing disk/database size


tanmoysrt:
For consecutive 7 days, the site has exceeded the plan's database or site usage, site will be suspended.

Site will be un-suspended on plan up-gradation

PR #2737, opened 2025-05-29
feat(site): Site suspension on exceeding disk / db limit

tanmoysrt:
Resolves https://github.com/frappe/press/issues/2735

- [x] Track First Site Usage Exceed Datetime + last when it's found
- [x] Run a daily job to send alert in email with some guide
- [x] Run daily job to suspend site which are excceeding site usage for last 7 days
  - [x] Send an mail along with the info and next steps to take 
- [x] On plan upgradation, check and un-suspend site
- [x] In between, if the usage comes down unsuspend site and reset state
- [x] Test cases

---
**Rule -**
If any sites on shared server cross it's plan limit, we will alert them daily for 7 days and then site will be suspended.

**What will reset the state -**
- Plan Change
- `Activate` site from desk.
- Disk / Database usage comes down automatically.

**How we can disable this auto suspension ? -**
- From `Press Settings`, keep `enforce_storage_limits` disabed
- At site level, mark the team/site as free.

---

image

image

### DIFF
--- a/press/press/doctype/site/site.py
+++ b/press/press/doctype/site/site.py
@@ -4243,7 +4243,7 @@ def send_warning_mail_regarding_sites_exceeding_disk_usage():
 				as_dict=True,
 			)
 			if not site_info.notify_email or (
-				site_info.current_disk_usage <= 100 and site_info.current_database_usage <= 100
+				site_info.current_disk_usage < 120 and site_info.current_database_usage < 120
 			):
 				# Final check if site is still exceeding limits
 				continue


press/press/doctype/site/site.py
```diff
@@ -2642,6 +2659,54 @@ def get_database_performance_report(self):
 		result["is_performance_schema_enabled"] = is_performance_schema_enabled
 		return result
 
+	def check_if_disk_usage_exceeded(self, save=True):
+		if self.free:
```
shadrak98: this only skips for free sites not free teams. We should consider free teams as well?

tanmoysrt: Free team and free site has been excluded from the checks.
Added few testcases as well.

PR #2745, opened 2025-06-03
refactor(telemetry): Add events in more steps

BreadGenie:


press/saas/doctype/product_trial_request/product_trial_request.py
```diff
@@ -95,6 +95,12 @@ def after_insert(self):
 		self.capture_posthog_event("product_trial_request_created")
 
 	def on_update(self):
+		if self.has_value_changed("site") and self.site:
+			init_telemetry()
+			ph = getattr(frappe.local, "posthog", None)
+			with suppress(Exception):
+				ph and ph.alias(previous_id=self.get_email(), distinct_id=self.site)
```
ankush: This can sometimes be problamatic? Same email will have multiple sites.

Account request and first site is better here? Both will be unique to a person/team.

BreadGenie: I've changed the distinct id to account request. In the current flow, we don't create a new account request for another trial (since we aren't doing a new signup). So for now, I'm tracking the first trial/signup.
Might need to rework how the flow works to track multiple product trials (with account request)

PR #2748, opened 2025-06-05
fix(app): Check for all installed app on site

Aradhya-Tripathi:
In install/uninstall app on site job check for all installed apps currently.


press/press/doctype/site/site.py
```diff
@@ -3638,12 +3638,15 @@ def process_install_app_site_job_update(job):
...
-				site.save()
+			site: Site = frappe.get_doc("Site", job.site)
+			apps = site.app_server_agent.request(
+				"GET", f"/benches/{job.bench}/sites/{job.site}/apps?format=json"
+			).get("data")
+			for app in apps:
+				app_doc = find(site.apps, lambda x: x.app == app)
+				if not app_doc:
+					site.append("apps", {"app": app})
+			site.save()
```
tanmoysrt: Use the existing `sync_apps` function ?

https://github.com/frappe/press/blob/950e053af3ac6245714490c6c670137ec2ec87a4/press/press/doctype/site/site.py#L623-L628

press/press/doctype/site/site.py
```diff
@@ -3638,12 +3638,8 @@ def process_install_app_site_job_update(job):
...
 	if updated_status != site_status:
 		if job.status == "Success":
-			site = frappe.get_doc("Site", job.site)
-			app = json.loads(job.request_data).get("name")
-			app_doc = find(site.apps, lambda x: x.app == app)
-			if not app_doc:
-				site.append("apps", {"app": app})
-				site.save()
+			site: Site = frappe.get_doc("Site", job.site)
+			site.sync_apps()
```
tanmoysrt: Maybe run this check, if agent job fails even.

If any app install fails in post-install phase, the app actually installed on site but on FC it will show not installed.
And next time user tries to install again, that throw error.

PR #2762, opened 2025-06-16
fix: Return `/` if Virtual Machine provider isn't EC2

mngshm:


Aradhya-Tripathi (approved):
LGTM

Aradhya-Tripathi:
This might not be required at all, since we only have arm64 on aws EC2 🤔

press/api/server.py
```diff
@@ -45,9 +45,15 @@ def _get_intel_mount_point(_: Literal["f", "m"]) -> str:
 
 
 def get_mount_point(server: str) -> str:
+	provider = frappe.get_value("Server", server, provider)
```
tanmoysrt:
```suggestion
	provider = frappe.get_value("Server", server, "provider")
```

tanmoysrt (approved):

press/api/server.py
```diff
@@ -45,9 +45,15 @@ def _get_intel_mount_point(_: Literal["f", "m"]) -> str:
 
 
 def get_mount_point(server: str) -> str:
+	provider = frappe.get_value("Server", server, "provider")
+	if provider != "AWS EC2":
+		return _get_intel_mount_point("/")
+
```
tanmoysrt, line 51: @mngshm here is a small mistake.

During fetching the provider. If `server` name starts with `f` check in `Server` doctype and for `m` check in `Database Server` doctype.

PR #2763, opened 2025-06-18
chore: Remove non-exist build column's linking

mngshm:
The build field (column in the DB) was removed some time ago in the commit https://github.com/frappe/press/commit/46dfda5385cd3d5c7e3761711c85df744bf40b60. But it always had the linking to build field inside "Deploy Candidate Build" doctype leading to weird `tabDeploy.build column not found` kind of errors.


Aradhya-Tripathi (approved):
LGTM

PR #2877, opened 2025-08-03
fix(cluster): Create deploy when new region is added

Aradhya-Tripathi:
Currently if update is not available we simply don't add region to release group which does not make sense.


Aradhya-Tripathi:
@MergifyIo backport-master

PR #2912, opened 2025-08-07
feat(server): Incident log

Aradhya-Tripathi:
- [x] Creating incident log for confirmed incidents
- [x] Creating incident resolved log for all resolved incidents on the server
~~Add likely cause post resolution.~~
~~Highlight incidents on the activity log.~~


Aradhya-Tripathi:
@MergifyIo backport master

PR #2913, opened 2025-08-07
chore(incident): Unidentified resource

Aradhya-Tripathi:
In case of unidentified resource fallback to server.


Aradhya-Tripathi:
@MergifyIo backport master

PR #2922, opened 2025-08-08
feat: Hetzner Volumes

mngshm:
Volumes for Hetzner.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -82,6 +82,7 @@
 		"COMPATBILITY",
 		"confs",
 		"Consolas",
+		"Containerised",
 		"cpath",
 		"cpcommerce",
 		"cpus",
--- a/dashboard/src/components/AlertBanner.vue
+++ b/dashboard/src/components/AlertBanner.vue
@@ -3,7 +3,11 @@
 		:class="`flex items-center rounded-md border border-${color}-200 bg-${color}-100 px-3.5 py-2.5`"
 	>
 		<lucide-alert-triangle
-			v-if="showIcon"
+			v-if="showIcon && (type === 'error' || type === 'warning')"
+			:class="`h-4 w-8 text-${color}-600`"
+		/>
+		<lucide-info
+			v-if="showIcon && type === 'info'"
 			:class="`h-4 w-8 text-${color}-600`"
 		/>
 		<div
--- a/dashboard/src/components/NavigationItems.vue
+++ b/dashboard/src/components/NavigationItems.vue
@@ -32,8 +32,6 @@ export default {
 			const routeName = this.$route?.name || '';
 			const onboardingComplete = this.$team.doc.onboarding.complete;
 			const isSaasUser = this.$team.doc.is_saas_user;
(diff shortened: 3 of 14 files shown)


press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -480,9 +487,12 @@ def increase_disk_size(self, volume_id=None, increment=50):
...
 		volume = find(self.volumes, lambda v: v.volume_id == volume_id)
 		volume.size += int(increment)
-		self.disk_size = self.get_data_volume().size
-		self.root_disk_size = self.get_root_volume().size
-		volume.last_updated_at = frappe.utils.now_datetime()
+		if self.cloud_provider != "Hetzner":
+			# These are AWS/OCI Specific checks. Hence, bypassing them
+			self.disk_size = self.get_data_volume().size
+			self.root_disk_size = self.get_root_volume().size
+			volume.last_updated_at = frappe.utils.now_datetime()
```
tanmoysrt: Will be better to override the behaviour  for Hetzner in the volume related functions ?

In `get_data_volume` & `get_root_volume` functions.

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -878,7 +918,7 @@ def get_temporary_volume_ids(self) -> list[str]:
 		tmp_volume_ids = set()
 		tmp_volumes_devices = [x.device for x in self.temporary_volumes]
 
-		def get_volume_id_by_device(device):
+		def gid_by_device(device):
```
tanmoysrt: `get_volume_id_by_device` name was more expressive :)

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -1416,21 +1482,30 @@ def get_next_volume_device_name(self):
...
+				return False
+			self.client().detach_volume(
+				Device=volume.device, InstanceId=self.instance_id, VolumeId=volume.volume_id
+			)
+		elif self.cloud_provider == "Hetzner":
+			for volume in self.volumes:
+				volume = self.client().volumes.get_by_id(volume.volume_id)
+				self.client().volumes.detach(volume=volume)
 		self.sync()
 		return True
```
tanmoysrt: AWS EBS handling missing ?

tanmoysrt (changes requested):
You can ignore the Physical Backup related things for now. That has already [validation](https://github.com/frappe/press/blob/develop/press/press/doctype/site_update/site_update.py#L249-L251) for `AWS EC2` provider,
If I enable that for Hetzner, I will do the required changes regarding that.

press/press/doctype/virtual_machine/virtual_machine.py
tanmoysrt: Also
```
		elif self.cloud_provider == "Hetzner":
			for volume in self.volumes:
				volume = self.client().volumes.get_by_id(volume.volume_id)
				self.client().volumes.detach(volume=volume)
```

This is detaching everything. Instead it should detach only requested one

PR #2997, opened 2025-08-14
fix: Populate vm specs, update servers and get cloud init

mngshm:


mngshm:
@Mergifyio backport master

PR #3039, opened 2025-08-22
fix: Remove syntax issue

mngshm:


mngshm:
@Mergify backport-master

PR #3046, opened 2025-08-24
fix: Correct bench permissions if backup fails

mngshm:


press/press/doctype/site_backup/site_backup.py
```diff
@@ -325,6 +333,22 @@ def create_database_snapshot(self):
...
+		"""
+		Run this whenever a Site Backup fails with the error
+		"[Errno 13]: Permission denied".
+		"""
+		job = frappe.db.get_value("Agent Job", self.job, filters={"status": "Failure"}, fields=["output"])
+		if job and "Permission denied" in job.output:
+			try:
+				Bench.correct_bench_permissions()
+				frappe.logger().info(f"Bench permissions corrected for backup {self.name}")
+				self.start_backup()
```
tanmoysrt: correct_bench_permissions going to enqueue the job - 

https://github.com/frappe/press/blob/6a3b019287d1de58fbaaf7fe2742568555f7e062/press/press/doctype/bench/bench.py#L356-L365

So, `self.start_backup()` is not waiting for that job to succeed.
Maybe, remove the backup trigger ?
Anyway FC will try to backup after some time again.

mngshm: ahh makes sense!

PR #3066, opened 2025-08-27
feat(alertmanager-webhook): Set webhook callback token in config

tanmoysrt:


tanmoysrt:
@Mergifyio backport master

press/api/monitoring.py
```diff
@@ -112,14 +112,14 @@ def targets(token=None):
...
-		if frappe.request.args.get("monitor_token") != monitor_token:
+		if frappe.request.args.get("webhook_token") != webhook_token:
 			raise frappe.AuthenticationError("Invalid credentials")
 
-		monitor_token = frappe.db.get_single_value("Press Settings", "monitor_token", cache=True)
+		webhook_token = frappe.db.get_single_value("Press Settings", "webhook_token", cache=True)
 
-		if frappe.request.args.get("monitor_token") != monitor_token:
+		if frappe.request.args.get("webhook_token") != webhook_token:
 			raise Exception
```
phot0n, line 123: this seems wrong? there is no field `webhook_token` in press settings

and not to mention there are 2 checks for the same thing raising different exceptions

we can just get the monitor server from press settings and then get it's token from it's document? - or store it redundantly in press settings (as it is currently in the form of `monitor_token`) - but add like a fetch on save from the actual monitor doc so that we dont have to add it manually (and also make it read only)

tanmoysrt: This looks like an old commit reference. It has been updated later.
Please check the monitoring.py file in develop or master branch.

https://github.com/frappe/press/blob/ef468599894ce0b140a7e91aa92790a8da874543/press/api/monitoring.py#L112-L123

Monitor Server doctype has the token field now.
You need to set some random token.

Then update any alert rule, that will set the token on monitor server.

phot0n: ah thanks

PR #3134, opened 2025-08-31
chore(server): Use the vm name from press job doctype

tanmoysrt:


tanmoysrt:
@Mergifyio backport master

PR #3152, opened 2025-09-01
fix(signup):  Fix login timeout fallback logic

Bowrna:
Resolve an infinite loading state during login when the WebSocket connection fails.

The initial socket failure correctly triggered the fallback mechanism, but the fallback code itself contained a bug that prevented the login process from completing. This patch corrects the logic in the fallback handler to ensure login proceeds using the alternative HTTP polling method.

The websocket is not pushing updated status as its broken. The get_progress function, which can ensure login using the polling method doesn't have "status" in the JSON response. As status is not sent in response, the page shows "Logging in" message for very longtime without any sign of moving to product setup

### DIFF
--- a/dashboard/src/pages/signup/LoginToSite.vue
+++ b/dashboard/src/pages/signup/LoginToSite.vue
@@ -172,7 +172,7 @@ export default {
 							};
 						},
 						onSuccess: (data) => {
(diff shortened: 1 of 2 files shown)


BreadGenie (approved):
LGTM

PR #3203, opened 2025-09-04
 feat: Don't show unnecessary records in listview 

tanmoysrt:


tanmoysrt:
failed test is different

PR #3261, opened 2025-09-15
chore(build): Obvious failure button & confirmation

Aradhya-Tripathi:
- Took the button outside of the dropdown.
- Showing a verbose confirmation modal

### DIFF
--- a/dashboard/src/pages/DeployCandidate.vue
+++ b/dashboard/src/pages/DeployCandidate.vue
@@ -28,6 +28,13 @@
 				<Badge class="ml-2" :label="deploy.status" />
 				<div class="ml-auto flex items-center space-x-2">
 					<Button
+						@click="failBuild"
+						v-if="deploy && deploy.status === 'Running'"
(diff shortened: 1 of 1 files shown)


dashboard/src/pages/DeployCandidate.vue
```diff
@@ -27,6 +27,13 @@
 				</h2>
 				<Badge class="ml-2" :label="deploy.status" />
 				<div class="ml-auto flex items-center space-x-2">
+					<Button
+						@click="failBuild"
+						v-if="deploy && deploy.status === 'Running'"
+						class="bg-red-50 text-red-600 hover:bg-red-100 border border-red-400"
+					>
+						Fail Build
+					</Button>
```
BreadGenie, line 36: Use `theme="red"` instead of adding custom css. You can add an icon using `iconLeft="x"`. Refer https://feathericons.com to use a different icon.

Also "Stop deploy" sounds like a better label that a normal user would understand.

PR #3286, opened 2025-09-17
fix: Log_error ✅ frappe.throw ❌

mngshm:
`frappe.throw()` blocks agent from proceeding ahead.


mngshm:
@Mergify backport master

PR #3292, opened 2025-09-17
chore: Take consent before performing dangerous action

mngshm:


mngshm:
@Mergify backport master

Issue #3288, opened 2025-09-17
One click elastic IP for server


balamurali27:
Users need static public IP for server. This is because they need to whitelist outbound IP across third party services.

- Only for AWS for now
- Can be made default for all servers

PR #3302, opened 2025-09-18
feat(cluster): Add elastic ip integration to ec2 servers (#3288)

regdocs:
- Add action to allocate an Elastic IP to an EC2 instance and trigger sync from the associated Virtual Machine
 - Add a step in sync under update_servers to update the new `is_static_ip` read-only field in Server if the returned IP is Elastic

### DIFF
--- a/press/press/doctype/server/server.py
+++ b/press/press/doctype/server/server.py
@@ -1909,6 +1909,7 @@ class Server(BaseServer):
 
 	if TYPE_CHECKING:
 		from frappe.types import DF
+
 		from press.press.doctype.resource_tag.resource_tag import ResourceTag
 		from press.press.doctype.server_mount.server_mount import ServerMount
 
--- a/press/press/doctype/server/server.py
+++ b/press/press/doctype/server/server.py
@@ -1909,7 +1909,6 @@ class Server(BaseServer):
 
 	if TYPE_CHECKING:
 		from frappe.types import DF
-
 		from press.press.doctype.resource_tag.resource_tag import ResourceTag
 		from press.press.doctype.server_mount.server_mount import ServerMount
 
@@ -1942,7 +1941,7 @@ class Server(BaseServer):
 		is_server_setup: DF.Check
 		is_standalone: DF.Check
 		is_standalone_setup: DF.Check
-		is_static_ip_setup: DF.Check
+		is_static_ip: DF.Check
 		is_upstream_setup: DF.Check
 		keep_files_on_server_in_offsite_backup: DF.Check
 		managed_database_service: DF.Link | None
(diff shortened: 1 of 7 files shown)


press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -966,6 +966,14 @@ def _sync_aws(self, response=None):  # noqa: C901
 			self.private_dns_name = instance.get("PrivateDnsName")
 			self.platform = instance.get("Architecture", "x86_64")
 
+			# Check if EC2 instance has Elastic IP and update all associated "Server" docs
+			servers = frappe.get_all("Server", filters={"virtual_machine": self.name}, pluck="name")
```
tanmoysrt: You can use `server = frappe.db.exists("Server", filters={"virtual_machine": self.name})`

https://docs.frappe.io/framework/user/en/api/database#frappedbexists

Btw, VM doctype gets linked with all kind of server doctypes - `Server`, `Database Server`, `Proxy Server` etc.
You can use the series name to guess the server type - `m` for db, `f` for app, `n` for proxy

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -966,6 +966,14 @@ def _sync_aws(self, response=None):  # noqa: C901
...
+			for server_name in servers:
+				server_doc = frappe.get_doc("Server", server_name)
+				server_doc.is_static_ip_setup = 1
+				server_doc.save(ignore_permissions=True)
```
tanmoysrt: You can use `frappe.db.set_value(....)` instead of fetching whole doc and update.

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -966,6 +966,14 @@ def _sync_aws(self, response=None):  # noqa: C901
...
+			frappe.db.commit()
```
tanmoysrt: Avoid manual commits unless absolutely necessary

press/press/doctype/server/server.py
```diff
@@ -2373,6 +2375,61 @@ def _setup_pyspy(self):
...
+
+			# Initialize EC2 client
+			ec2_client = boto3.client(
+				"ec2",
+				aws_access_key_id=aws_access_key_id,
+				aws_secret_access_key=aws_secret_access_key,
+				region_name=region_name,
+			)
+			print("[AWS] EC2 client initialized successfully")
+
```
tanmoysrt: VM doctype has `def client(self, client_type=None)`
You can use that to get the boto3 client.

https://github.com/frappe/press/blob/develop/press/press/doctype/virtual_machine/virtual_machine.py#L1357

press/press/doctype/server/server.py
```diff
@@ -2373,6 +2375,61 @@ def _setup_pyspy(self):
 		except Exception:
 			log_error("Setup PySpy Exception", server=self.as_dict())
 
+	@frappe.whitelist()
+	def get_aws_static_ip(self):
+		try:
+			vm_doc = frappe.get_doc("Virtual Machine", self.virtual_machine)
+
+			if vm_doc.cloud_provider != "AWS EC2":
+				frappe.throw("Failed to proceed as VM is not AWS EC2")
```
tanmoysrt: Maybe place this validation outside the try...catch to avoid getting logged in error log

press/press/doctype/server/server.py
```diff
@@ -2373,6 +2375,61 @@ def _setup_pyspy(self):
...
+
+			print(f"[AWS] Starting static IP allocation for VM {self.virtual_machine}")
```
tanmoysrt: Before merging, remove the print statements ?

If you need to print in local env, you can do something like this 
```
if frappe.conf.developer_mode:
   print(.....)
```

press/press/doctype/server/server.py
```diff
@@ -2373,6 +2375,61 @@ def _setup_pyspy(self):
...
+
+			# Associate with instance
+			ec2_client.associate_address(InstanceId=instance_id, AllocationId=allocation_id)
+			print(f"[AWS] Associated Elastic IP {public_ip} with Instance {instance_id}")
+
+			# Trigger VM sync
+			vm_doc.sync()
+			print(f"[AWS] Cluster {cluster_doc.name} synced after IP allocation")
+
+			return f"Static IP {public_ip} alloted to the VM (Allocation ID: {allocation_id})"
```
tanmoysrt: Will be better to use `frappe.msgprint` to show the message in frontend instead of returning the message
https://docs.frappe.io/framework/user/en/api/dialog#frappemsgprint

press/press/doctype/virtual_machine/virtual_machine.py
regdocs: Thank you, I'll remember that. I shifted this code to a function called `update_servers` under VM doctype, on Balu's feedback. Could you please take a look once more?

press/press/doctype/virtual_machine/virtual_machine.py
regdocs: Done, thank you.

press/press/doctype/virtual_machine/virtual_machine.py
regdocs: Done

press/press/doctype/server/server.py
regdocs: Oh this is the server doctype. I was looking but couldn't find it in server, however found it in VM.

press/press/doctype/server/server.py
regdocs: Got it, I actually removed the try..except on Balu's feedback since it's a web request and there is an error boundary set up in frappe.

So this condition and the following code is directly under the function's root scope.

press/press/doctype/server/server.py
regdocs: Done, thank you.

press/press/doctype/server/server.py
regdocs: Got it.

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -1047,6 +1043,25 @@ def get_data_volume(self):
 			return volume
 		return frappe._dict({"size": 0})
 
+	def is_elastic_ip_aws(self) -> bool:
+		instance_id = self.instance_id
+		ec2 = self.client()
+
+		reservations = ec2.describe_instances(InstanceIds=[instance_id])["Reservations"]
+		instance = reservations[0]["Instances"][0]
```
ssiyad: Shouldn't there be some length checks here? Blind array indexing does not seem safe.

regdocs: Done, added KeyError and IndexError boundary. If any of these is thrown, function returns False.

PR #3311, opened 2025-09-18
feat(registry): Registry mirror

Aradhya-Tripathi:
- Add missing configuration files for registry container.
- Allow creating of registries on the fly in different clusters using the base registry.
- Pass mirroring configuration on registry.


Aradhya-Tripathi:
Unrelated breaking tests.

PR #3338, opened 2025-09-20
fix: Use Site Partner Lead doctype to store lead

tanmoysrt:


tanmoysrt:
@mergifyio backport master

PR #3356, opened 2025-09-22
fix(signups): Move create subscription part as bg job

Bowrna:
This PR moves the creating subscription process for precreated site from the sign up flow into background job that runs for every 5 mins and take care of it there.

### DIFF
--- a/press/press/doctype/site/site.py
+++ b/press/press/doctype/site/site.py
@@ -4412,3 +4412,35 @@ def suspend_sites_exceeding_disk_usage_for_last_7_days():
 			# Check once again and suspend if still exceeds limits
 			site: Site = frappe.get_doc("Site", site.name)
 			site.suspend(reason="Site Usage Exceeds Plan limits", skip_reload=True)
+
+
+def create_subscription_for_trial_sites():
+	# Get sites that are in "Site Created" status and has no entry in "Site Plan Change"
+	# For those sites, invoke "Create Subscription" that puts entry into "Site Plan Change" and "Subscription"
+	active_sites = frappe.db.sql(
+		"""
+		SELECT trial.site, producttrial.trial_plan
+		FROM `tabProduct Trial Request` trial
+		LEFT JOIN `tabSite Plan Change` siteplanchange
+		ON trial.site = siteplanchange.name
+		LEFT JOIN `tabProduct Trial`  producttrial ON trial.product_trial = producttrial.name WHERE trial.is_subscription_created = 0 AND siteplanchange.name is NULL AND trial.status="Site Created" LIMIT 25;
+		""",
+		as_dict=True,
+	)
+	frappe.log(f"Creating subscription for the sites {active_sites}")
+	for trial_site in active_sites:
+		if has_job_timeout_exceeded():
+			return
+		try:
+			site: Site = frappe.get_doc("Site", trial_site.site)
+			site.create_subscription(trial_site.trial_plan)
+			frappe.db.set_value(
+				"Product Trial Request",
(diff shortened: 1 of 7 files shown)


press/press/doctype/site/site.py
```diff
@@ -4412,3 +4412,35 @@ def suspend_sites_exceeding_disk_usage_for_last_7_days():
...
+def create_subscription_for_trial_sites():
+	# Get sites that are in "Site Created" status and has no entry in "Site Plan Change"
+	# For those sites, invoke "Create Subscription" that puts entry into "Site Plan Change" and "Subscription"
+	active_sites = frappe.db.sql(
+		"""
+		SELECT trial.site, producttrial.trial_plan
+		FROM `tabProduct Trial Request` trial
+		LEFT JOIN `tabSite Plan Change` siteplanchange
+		ON trial.site = siteplanchange.name
+		LEFT JOIN `tabProduct Trial`  producttrial ON trial.product_trial = producttrial.name WHERE trial.is_subscription_created = 0 AND siteplanchange.name is NULL AND trial.status="Site Created";
```
tanmoysrt: maybe add a limit in sql query to only pick ~25 at a time.
and run the job every minute instead of every 5 mins.

press/press/doctype/site/site.py
```diff
@@ -4412,3 +4412,35 @@ def suspend_sites_exceeding_disk_usage_for_last_7_days():
...
+		""",
+		as_dict=True,
+	)
+	frappe.log(f"Creating subscription for the sites {active_sites}")
+	for trial_site in active_sites:
+		if has_job_timeout_exceeded():
+			return
+		try:
+			site: Site = frappe.get_doc("Site", trial_site.site)
+			site.create_subscription(trial_site.trial_plan)
```
tanmoysrt, line 4446: Just for sanity, maybe check here if the `Site` has any active subscription already.
If already one active subscription or `Site Plan Change` exists, assume that subscription has been created.

press/press/doctype/site/site.py
Bowrna: yes i will do this changes. @tanmoysrt

press/press/doctype/site/site.py
Bowrna: @tanmoysrt In the query, i did left join with `Site Plan Change` and checked if it is NULL to confirm that subscription doesn't exist and then have added Subscription. Do you think it would be enough? 

```
SELECT trial.site, producttrial.trial_plan
		FROM `tabProduct Trial Request` trial
		LEFT JOIN `tabSite Plan Change` siteplanchange
		ON trial.site = siteplanchange.name
		LEFT JOIN `tabProduct Trial`  producttrial ON trial.product_trial = producttrial.name WHERE trial.is_subscription_created = 0 AND siteplanchange.name is NULL AND trial.status="Site Created; 
```

tanmoysrt: Ah okay, that's fine then.
I have missed this part.

press/saas/doctype/product_trial/product_trial.py
```diff
@@ -121,7 +121,7 @@ def setup_trial_site(
 			site.signup_time = frappe.utils.now()
 			site.generate_saas_communication_secret(create_agent_job=True, save=False)
 			site.save()  # Save is needed for create_subscription to work TODO: remove this
-			site.create_subscription(plan)
+			# site.create_subscription(plan)
```
tanmoysrt:
```suggestion
```

tanmoysrt (approved):
Looks good

PR #3379, opened 2025-09-24
fix: If job.output ...

mngshm:


mngshm:
@Mergify backport master

PR #3382, opened 2025-09-24
fix: Make Stripe billing updates conditional (#3381)

### LINKED ISSUE #3381: Bug Report: Stripe Error When Updating Billing Details Without Stripe Configuration
**Problem Description**
When attempting to update billing details through the dashboard, users encounter a Stripe configuration error even when they don't intend to use Stripe as their payment method. The system should allow billing details updates without requiring Stripe to be configured.

**Error Message**
Setup stripe via Press Settings before using press.api.billing.get_stripe

**Steps to Reproduce**
- Access the Frappe Cloud dashboard
- Navigate to billing settings/onboarding
- Attempt to update billing details (name, address, etc.)
- Click "Update Billing Details" button
- Error occurs even when Stripe is not configured or intended to be used

**Expected Behavior**
- Billing details should be updatable without Stripe configuration
- Stripe integration should only be called when Stripe is properly configured in Press Settings

**Current Behavior**
- The system unconditionally calls [update_billing_details_on_stripe()]
- This causes the [get_stripe()] function to throw an error when Stripe is not configured
- Users cannot update basic billing information

**Attachments:**

https://github.com/user-attachments/assets/7f395968-0e76-44aa-a313-dc080d899fb7

Image

mayankgongal5:
Only call update_billing_details_on_stripe() when Stripe is configured. Prevents errors when updating billing details with non-Stripe payment methods.

- Add utility functions to check Stripe configuration safely
- Add conditional logic in Team.update_billing_details()
- Add error handling for Stripe operations

Fixes #3381

https://github.com/user-attachments/assets/1d805855-c1d9-4ad4-a942-763c17eb67da

### DIFF
--- a/press/press/doctype/team/team.py
+++ b/press/press/doctype/team/team.py
@@ -660,7 +660,8 @@ class Team(Document):
 		self.save()
 		self.reload()
 
-		self.update_billing_details_on_stripe(address_doc)
+		# Only update Stripe if it's configured
+		self._safely_update_stripe_billing(address_doc)
 		self.update_billing_details_on_frappeio()
 		self.update_billing_details_on_draft_invoices()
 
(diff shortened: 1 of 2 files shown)


mayankgongal5:
any feedback or review ?

press/utils/billing.py
```diff
@@ -153,6 +153,29 @@ def get_setup_intent(team):
...
+		stripe_secret_key = get_decrypted_password(
+			"Press Settings",
+			"Press Settings", 
+			"stripe_secret_key",
+			raise_exception=False,
+		)
+		return bool(stripe_secret_key)
+	except Exception:
+		return False
+
```
tanmoysrt, line 169: Check for `stripe_publishable_key` or have a checkbox on `Press Settings` ?

press/press/doctype/team/team.py
tanmoysrt, line 719: We can just add a guard clause here

```
if not stripe_configured:
	return
```

rather than doing changes at other places.

PR #3404, opened 2025-09-29
fix(signup): Server side validation for subdomain and domain

Bowrna:
Server side validation for subdomain and domain to prevent adding invalid names during the signup process.

### DIFF
--- a/press/press/doctype/root_domain/root_domain.py
+++ b/press/press/doctype/root_domain/root_domain.py
@@ -10,6 +10,7 @@ import boto3
 import frappe
 from frappe.core.utils import find
 from frappe.model.document import Document
+from frappe.utils.caching import redis_cache
 
 from press.utils import log_error
 
@@ -208,3 +209,8 @@ def cleanup_cname_records():
 			continue
 
 		domain.remove_unused_cname_records()
+
+
+@redis_cache(ttl=3600)
+def get_domains():
+	return frappe.get_all("Root Domain", filters={"enabled": ["=", "1"]}, pluck="name")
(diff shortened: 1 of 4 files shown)


press/saas/doctype/product_trial_request/product_trial_request.py
```diff
@@ -224,8 +231,7 @@ def create_site(self, subdomain: str, domain: str):
 		if self.status != "Pending":
 			return
 
-		if not subdomain:
-			frappe.throw("Subdomain is required to create a site.")
+		ProductTrialRequest.validate_subdomain_domain(subdomain, domain)
```
ssiyad: Just use `self.` no?

press/saas/doctype/product_trial_request/product_trial_request.py
```diff
@@ -213,6 +214,12 @@ def get_setup_wizard_payload(self):
 			)
 			frappe.throw(f"Failed to generate payload for Setup Wizard: {e}")
 
+	@staticmethod
+	def validate_subdomain_domain(subdomain: str, domain: str):
+		validate_subdomain(subdomain)
+		if domain not in get_domains():
+			frappe.throw("Domain used is invalid")
```
ssiyad: "Invalid domain" might be better

press/press/doctype/root_domain/root_domain.py
```diff
@@ -208,3 +209,9 @@ def cleanup_cname_records():
 			continue
 
 		domain.remove_unused_cname_records()
+
+
+@redis_cache(ttl=3600)
+def get_domains():
+	domains = frappe.get_all("Root Domain", filters={"enabled": ["=", "1"]}, fields=["name"])
```
ssiyad: Use `pluck` instead of `fields`

press/saas/doctype/product_trial_request/product_trial_request.py
```diff
@@ -213,6 +214,12 @@ def get_setup_wizard_payload(self):
 			)
 			frappe.throw(f"Failed to generate payload for Setup Wizard: {e}")
 
+	@staticmethod
+	def validate_subdomain_domain(subdomain: str, domain: str):
```
ssiyad: Method name is a little confusing 🤷‍♂️

Issue #3041, opened 2025-08-22
Strip out scheme while adding custom domain


tanmoysrt:
During adding custom domain, users can put url in domain field.
It happens normally in browser, while copying any domain from any document or text.

We should extract the domain name from the URL and then proceed to next step.
https://support.frappe.io/helpdesk/tickets/47110

PR #3409, opened 2025-09-30
feat(add-custom-domain): Remove scheme and resource path before submit

regdocs:
Fixes frappe/press #3041

### DIFF
--- a/dashboard/src/components/AddDomainDialog.vue
+++ b/dashboard/src/components/AddDomainDialog.vue
@@ -104,12 +104,7 @@
 				class="mt-2 w-full"
 				variant="solid"
 				:loading="$resources.checkDNS.loading"
-				@click="
-					$resources.checkDNS.submit({
-						name: site.name,
-						domain: newDomain,
-					})
-				"
+				@click="preprocessAndVerifyDns()"
 			>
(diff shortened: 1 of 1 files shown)


dashboard/src/components/AddDomainDialog.vue
```diff
@@ -148,6 +143,20 @@ export default {
 			newDomain: null,
 		};
 	},
+	methods: {
+		preprocessAndVerifyDns() {
+			if (this.newDomain) {
+				// Remove scheme if present
+				this.newDomain = this.newDomain.replace(/(^\w+:|^)\/\//, '');
```
BreadGenie, line 151: it's better to give some examples in comment if we do regex
eg: `https://frappecloud.com -> frappecloud.com`

dashboard/src/components/AddDomainDialog.vue
```diff
@@ -148,6 +143,20 @@ export default {
 			newDomain: null,
 		};
 	},
+	methods: {
+		preprocessAndVerifyDns() {
```
BreadGenie: try to do only one thing with a function/method

BreadGenie (approved):

PR #3392, opened 2025-09-26
feat(dashboard_banner): Make Dashboard Banner not single doctype

regdocs:
This resolves frappe/press issue [#3374](https://github.com/regdocs/press/issues/3374)

PR #3427, opened 2025-09-30
fix(dashboard-banner): Prevent empty IN clauses in query builder

regdocs:
Fixes bug in PR #3392

### DIFF
--- a/press/api/account.py
+++ b/press/api/account.py
@@ -96,7 +96,7 @@ def verify_otp(account_request: str, otp: str) -> str:
 
 
 @frappe.whitelist(allow_guest=True)
-@rate_limit(limit=5, seconds=60)
+@rate_limit(limit=5, seconds=60 * 60)
 def verify_otp_and_login(email: str, otp: str):
 	from frappe.auth import get_login_attempt_tracker
 


press/press/doctype/dashboard_banner/dashboard_banner.py
```diff
@@ -76,19 +76,28 @@ def get_user_banners():
...
-		.where(DashboardBanner.enabled == 1)
-		.where(
-			(DashboardBanner.is_global == 1)
-			| ((DashboardBanner.type_of_scope == "Site") & (DashboardBanner.site.isin(sites)))
-			| ((DashboardBanner.type_of_scope == "Server") & (DashboardBanner.server.isin(servers)))
-			| ((DashboardBanner.type_of_scope == "Team") & (DashboardBanner.team == team))
+	query = frappe.qb.from_(DashboardBanner).select("*").where(DashboardBanner.enabled == 1)
+
+	conditions = [DashboardBanner.is_global == 1]
+
```
tanmoysrt: btw you can do this, instead of appending in a list.


```suggestion
	query = frappe.qb.from_(DashboardBanner).select("*").where(DashboardBanner.enabled == 1)
	
	query = query.where(DashboardBanner.is_global == 1)
```

regdocs: Got it, so where clauses can be chained like that.

But will this also work if the `is_global == 1` is a part of the OR group? That was my original intention here.

tanmoysrt: Ah okay, got it.

You can chain or as well that, but you need to prepare the clause inside where seperately first.

```
filters = (DashboardBanner.is_global == 1)

if something:
	filters = filters | (DashboardBanner.type_of_scope == "Site") & (DashboardBanner.site.isin(sites))
```
Anyway, this above approach kinda same as your approach. Ig it's fine.

---

Also, there is another way to supress raising error if blank list passed to qb.

Check this example -

https://github.com/frappe/press/blob/843ecaacae58fe0d5e1555f5a07a783e7972e70e/press/api/marketplace.py#L205

regdocs: Ah this makes sense, so `[""]` won't affect the original condition, but at the same time it prevents the error because IN doesn't get an empty array.

regdocs: Thank you, I'll try applying this.

PR #3435, opened 2025-10-01
fix(dashboard-banner): Move handler methods to press.api.account

regdocs:


ssiyad (commented):
Looks good but I think `get_list_query` and `run_doc_method` can replace both endpoints. Maybe make those changes in next iteration.

shadrak98 (approved):
LGTM

press/api/account.py
```diff
@@ -1415,6 +1415,68 @@ def reset_2fa_recovery_codes():
...
+			{
+				"user": user,
+				"dismissed_at": frappe.utils.now(),
+				"dashboard_banner": banner_name,
+			},
+		)
+		banner.save(ignore_permissions=True)
+		return True
+	return False
+
```
BreadGenie: let's move this to the controller file and keep it as a method

then you can do this in the frontend to dismiss banner

```
		dismissBanner() {
			return {
				url: 'press.api.client.run_doc_method',
				makeParams() {
					return {
						dt: 'Dashboard Banner',
						dn: <doc_name here>,
						method: 'dismiss_banner',
					};
				},
			};
		},
```

press/api/account.py
```diff
@@ -1415,6 +1415,68 @@ def reset_2fa_recovery_codes():
...
+	if banner and banner.is_dismissible:
+		banner.append(
+			"user_dismissals",
+			{
+				"user": user,
+				"dismissed_at": frappe.utils.now(),
+				"dashboard_banner": banner_name,
+			},
+		)
+		banner.save(ignore_permissions=True)
```
BreadGenie: Using `ignore_permissions` isn't a good idea, I think.

BreadGenie:
> @ssiyad Got it, will do. Do you think the get list method will support a complex query like the one in get_user_banners?

It doesn't 
Fetching banners will have to stay

PR #3444, opened 2025-10-03
fix(login): Request key expired signed up user could not login

Bowrna:
This PR fixes the issue that allows the user who signed up for Frappe account, but didn't continue finishing the setup account process and later came back and logged in. The login started to throw permission issue as the account request key got expired and they were struck in that part forever. This fix regenerates the account key if expired and ensures that user can continue setup of the account process

### DIFF
--- a/press/press/doctype/site/site.py
+++ b/press/press/doctype/site/site.py
@@ -4435,7 +4435,6 @@ def create_subscription_for_trial_sites():
 		""",
 		as_dict=True,
 	)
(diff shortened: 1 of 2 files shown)


press/press/doctype/site/site.py
```diff
@@ -4435,7 +4435,6 @@ def create_subscription_for_trial_sites():
 		""",
 		as_dict=True,
 	)
```
Bowrna, line 4437: I have removed the below because its flooding the logs file.

PR #3464, opened 2025-10-06
fix: Use Ubuntu 22.04 Jammy repos for Docker images

cogk:


Aradhya-Tripathi:
We had to build wkhtmltopdf binaries to support wkhtmltopdf-`0.12.5`

 https://github.com/frappe/press/pull/3478

PR #3487, opened 2025-10-08
chore: add chromium dependencies

iamejaaz:


ankush:
@iamejaaz don't use `chore` for production changes please

PR #3496, opened 2025-10-08
feat(virtual-disk-resize): Add background job to schedule disk resize

regdocs:


press/infrastructure/doctype/virtual_disk_resize/virtual_disk_resize.py
```diff
@@ -77,7 +79,10 @@ def before_insert(self):
 
 	def after_insert(self):
 		"""Enqueue current volume attribute fetch and volume creation"""
-		frappe.enqueue_doc(self.doctype, self.name, "run_prerequisites", queue="long", timeout=2400)
+		if not self.scheduled_time:
+			self.status = "Pending"
+			self.save()
+			frappe.enqueue_doc(self.doctype, self.name, "run_prerequisites", queue="long", timeout=2400)
```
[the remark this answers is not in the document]
regdocs: @Copilot "Scheduled" is default

press/infrastructure/doctype/virtual_disk_resize/virtual_disk_resize.py
```diff
@@ -87,6 +92,8 @@ def run_prerequisites(self):
 		self.create_new_volume()
 		self.status = Status.Ready
 		self.save()
+		if self.scheduled_time:
+			self.execute()
```
tanmoysrt, line 96: @regdocs 

This code block sometimes can break weirdly.

self.create_new_volume calls -> attach_new_volume of vm doctype -> that calls vm.sync() method.

This vm.sync can raise TimestampMismatchError. You need to ignore this error and do manual sync until unless vm doctype gets updated.

----

Also, as a safeguard add a try..catch in run_prerequisites and set status to `Failure` if it failed.

Else, the bg job will keep trying to run same job multiple times and it can cause unknown issues.

press/infrastructure/doctype/virtual_disk_resize/virtual_disk_resize.py
```diff
@@ -686,3 +695,21 @@ class Status(str, Enum):
...
+	for task in resize_tasks:
+		try:
+			frappe.enqueue_doc(
+				task.doctype,
+				task.name,
+				"run_prerequisites",
+				queue="long",
+				timeout=2400,
+			)
+		except Exception as e:
```
tanmoysrt, line 714: Please use `job_id` and `deduplicate=True`, so that scheduler doesn't enqueue it multiple times.

Example - 

https://github.com/frappe/press/blob/63e0f1fbdc9a319a5e23957fc7fdd27a75da19f7/press/press/doctype/virtual_machine/virtual_machine.py#L1635-L1644

tanmoysrt: Also, move the status from `Scheduled` to `Pending` first and then enqueue the `run_prerequisites` job ?

Probably do it in `VirtualDiskResize` class itself

PR #3518, opened 2025-10-09
fix(login): Add session id verification during login

Bowrna:


press/api/site_login.py
```diff
@@ -147,27 +147,30 @@ def verify_otp(email: str, otp: str):
...
 
-@frappe.whitelist(allow_guest=True)
+def verify_session_user():
+	session_id = frappe.local.request.cookies.get("site_user_sid")
+	if not session_id or not isinstance(session_id, str):
+		return None
+	return frappe.db.get_value("Site User Session", {"session_id": session_id, "verified": 1}, "user")
+
+
+@frappe.whitelist(allow_guest=False)
```
BreadGenie:
```suggestion
@frappe.whitelist()
```
allow_guest is by default False

press/api/site_login.py
```diff
@@ -147,27 +147,30 @@ def verify_otp(email: str, otp: str):
 	)
 
 
-@frappe.whitelist(allow_guest=True)
+def verify_session_user():
+	session_id = frappe.local.request.cookies.get("site_user_sid")
+	if not session_id or not isinstance(session_id, str):
+		return None
+	return frappe.db.get_value("Site User Session", {"session_id": session_id, "verified": 1}, "user")
```
BreadGenie: 2 types of users do site login
- A user who has FC account and is logged in won't have `Site User Session` record created
- User who has no FC account or they're not logged in

here we only consider for the user without FC account

ssiyad:
There is a difference is mails sent during signup and login right? When signing up we will be sending a mail saying "This is your OTP to activate the site". But during login, we send an email saying "This is your OTP to login to your site", which may not exist.

press/api/site_login.py
```diff
@@ -97,19 +102,69 @@ def get_product_sites_of_user(user: str):
 	)
 
 
+def verify_user_in_site(site_doc: "Site", user: str) -> bool:
+	try:
+		query = f"""
+			SELECT name, enabled
+			FROM `tabUser`
+			WHERE name = '{user}'
+		"""
```
ssiyad, line 111: Please convert to query builder.

press/api/site_login.py
```diff
@@ -97,19 +102,69 @@ def get_product_sites_of_user(user: str):
...
+		response = site_doc.run_sql_query_in_database(query, False)
+		if (
+			response.get("success", False)
+			and response.get("data")
+			and response["data"][0].get("row_count", 0) == 1
```
ssiyad, line 116: Array length check would be nice.

press/api/site_login.py
```diff
@@ -97,19 +102,69 @@ def get_product_sites_of_user(user: str):
...
+			frappe.qb.from_(Site)
+			.select(Count(Site.name).as_("site_count"))
+			.join(SiteUser)
+			.on(SiteUser.site == Site.name)
+			.where(SiteUser.user == email)
+			.where(SiteUser.enabled == 1)
+			.where(Site.status == "Active")
+		)
+		output = query.run(as_dict=True)
+		return output[0].get("site_count", 0) if output else 0
```
ssiyad, line 141: Array length check would be nice.

press/api/site_login.py
```diff
@@ -97,19 +102,69 @@ def get_product_sites_of_user(user: str):
...
+			response.get("success", False)
+			and response.get("data")
+			and response["data"][0].get("row_count", 0) == 1
+		):
+			_, is_user_enabled = response["data"][0]["output"]["data"][0]
+			return bool(is_user_enabled)
+	except frappe.DoesNotExistError:
+		pass
+	except Exception as e:
+		log_error("Error verifying user in site", data=e)
```
ssiyad, line 123: Why not throw an error?

press/api/site_login.py
```diff
@@ -97,19 +102,69 @@ def get_product_sites_of_user(user: str):
...
+			.on(SiteUser.site == Site.name)
+			.where(SiteUser.user == email)
+			.where(SiteUser.enabled == 1)
+			.where(Site.status == "Active")
+		)
+		output = query.run(as_dict=True)
+		return output[0].get("site_count", 0) if output else 0
+	except Exception as e:
+		log_error("Error fetching site count", data=e)
+		return 0
```
ssiyad, line 144: Why not throw an error?

press/api/site_login.py
```diff
@@ -97,19 +102,69 @@ def get_product_sites_of_user(user: str):
...
-	return session.send_otp()
+		frappe.throw("Please wait for 30 seconds before sending the OTP again")
+	try:
+		session = frappe.get_doc({"doctype": "Site User Session", "user": email}).insert(
+			ignore_permissions=True
+		)
+		return session.send_otp()
+	except Exception as e:
+		log_error("Error sending OTP", data=e)
+		frappe.throw("Sending OTP failed")
```
ssiyad, line 167: Do we need both `log_error` and `frappe.throw`?

Bowrna: @ssiyad Added to create an error log about the issue and also throw relevant error message at frontend. Let me know if you think i could do it better

press/api/site_login.py
Bowrna: @ssiyad @tanmoysrt I have not used this function during login to site  to verify if user is in site. My idea was initially to use it as we want FC to work as auth layer, this will run the function remotely via agent in site to check if the user is present and enabled. 

Currently, I am still using "Site User" doctype that we have at press side as before. As we are syncing site user between site and press. I have not removed and committed this in code to ask your opinion on can we rely on existing "Site User" doc or do you think we need to run query at site site and check?

press/api/site_login.py
ssiyad: [frappe.throw](https://github.com/frappe/frappe/blob/7e2a2f306522f5d6b1d81189d3816a26572e19c3/frappe/utils/messages.py#L126-L154) is a wrapper around msgprint which also raise an exception.

press/api/site_login.py
tanmoysrt: @Bowrna works. Use `shlex.quote` for quotation around user variable.

press/api/site_login.py
```diff
@@ -82,7 +88,6 @@ def get_product_sites_of_user(user: str):
 	sites = frappe.db.get_all(
 		"Site User", filters={"user": user, "enabled": 1}, fields=["site"], pluck="site"
 	)
```
tanmoysrt, line 90: @Bowrna we need to get rid of this api.

First, do a initial verification of user mail, then only proceed to next ?

Or, you can do this as well
- FC doesn't need to give option to user on which site they want to login
- User goes to his own site example.frappe.cloud
- Clicked on `Login with Frappe Cloud`
- Either pass the site name while redirecting to FC or utilize http [referrer](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Referer) header to get the site name
- Then follow the usual flow.

Only one issue is that - People can forget the site url and then become clueless. But, I don't think that's a big issue comparatively.
They can login to FC and see the site names.

press/api/site_login.py
ssiyad: We can also use `.get_sql()` from query builder.

press/api/site_login.py
Bowrna: Here verifying if user exists in "Site User Session" is the initial verification of user mail in this api. Do you see any flaw in that part @tanmoysrt ?

tanmoysrt: We are allowing any user to access this method `allow_guest=True`

```
@frappe.whitelist(allow_guest=True)
@rate_limit(limit=10, seconds=60)
```

That means, anyone can just bruteforce and extract the site names owned by users. The rate limit is also not that hard `@rate_limit(limit=10, seconds=60)`

PR #3547, opened 2025-10-13
fix(virtual-disk-resize): Fix TimestampMismatchErrors

regdocs:
Resolve the TimestampMismatchError when virtual disk resize (a long running op) fails to save the doc because of concurrent background sync tasks trying to save to the same doc

### DIFF
--- a/press/infrastructure/doctype/virtual_disk_resize/virtual_disk_resize.py
+++ b/press/infrastructure/doctype/virtual_disk_resize/virtual_disk_resize.py
@@ -94,6 +94,10 @@ class VirtualDiskResize(Document):
 			self.save()
 			if self.scheduled_time:
 				self.execute()
+		except frappe.QueryTimeoutError:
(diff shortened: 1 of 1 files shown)


press/infrastructure/doctype/virtual_disk_resize/virtual_disk_resize.py
```diff
@@ -85,15 +84,23 @@ def after_insert(self):
...
+			self.save()
+			self.set_filesystem_attributes()
+			self.set_new_volume_attributes()
+			self.create_new_volume()
+			self.status = Status.Ready
+			self.save()
+			if self.scheduled_time:
+				self.execute()
+		except frappe.QueryTimeoutError:
+			frappe.db.rollback()
```
Aradhya-Tripathi, line 98: Maybe release the lock as soon as possible via a commit?

Instead of excepting timeout errors?

PR #3559, opened 2025-10-14
fix(2fa-token-test): Add new 2FA code generator

### LINKED ISSUE #3558: [Test - User 2FA] Recovery code validation test case for `isupper()` failing in rare edge case
https://github.com/frappe/press/actions/runs/18464166057/job/52602123240?pr=3550

regdocs:
Resolves #3558

### DIFF
--- a/press/press/doctype/user_2fa/test_user_2fa.py
+++ b/press/press/doctype/user_2fa/test_user_2fa.py
@@ -12,9 +12,4 @@ class TestUser2FA(FrappeTestCase):
 		recovery_codes = list(User2FA.generate_recovery_codes())
 		self.assertEqual(len(recovery_codes), User2FA.recovery_codes_max)
 		self.assertTrue(all(len(code) == User2FA.recovery_codes_length for code in recovery_codes))
-		self.assertTrue(
-			all(
-				(code.isupper() if any(c.isalpha() for c in code) else code.isnumeric())
-				for code in recovery_codes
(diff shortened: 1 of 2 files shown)


ssiyad:
I guess the output is fine but it runs four loops, which is a bit too much I think.

ssiyad:
Maybe use the current setup with a counter? It might be the easy way.

regdocs:
@ssiyad @tanmoysrt I feel, all of this entirely unnecessary. We're not contributing at all to the entropy of the code by ensuring there is at least one uppercase letter in the string. In fact, this reduces the sample space even further (< 36¹⁶).

The number of hex codes of length 16 that have at least one digit AND at least one uppercase letter is 36¹⁶ - 10¹⁶ - 26¹⁶, lesser than the maximum combinations 36¹⁶.

In fact I think, it is the test case that needs to be rectified here, to account for the digits-only only cases. The hex generator function we currently have is sufficiently secure.

ssiyad:
The math seems incorrect here. Having alphabets is definitely more secure than just numbers. To be honest, I don't think a fix is hard and I don't understand why we are overcomplicating things.

> In fact I think, it is the test case that needs to be rectified here, to account for the digits-only only cases.

No. Don't fix the test by fixing the test. That feels wrong.

regdocs:
Fixed this with @ssiyad in a call.

PR #3583, opened 2025-10-15
Fixes Python installation breakage for Ubuntu Focal.

0spinboson:
Fix Python installation and analytics API for Ubuntu focal. Replace deadsnakes PPA with Python-from-source build to fix Ubuntu Focal compatibility. 
- Add dynamic version fetching for latest Python patch releases, 
- Add comprehensive Python build dependencies for source compilation, 
- Include build caching optimizations for faster subsequent builds.
- Also small fix for analytics.daily_usage function to handle dict returns properly.

This resolves build failures on Ubuntu Focal where deadsnakes PPA no longer provides packages, and fixes 'str has no attribute max' errors in analytics API.

### DIFF
--- a/press/api/analytics.py
+++ b/press/api/analytics.py
@@ -721,8 +721,15 @@ def daily_usage(name, timezone):
 
 	plan = frappe.get_cached_doc("Site", name).plan
 
+	# Handle case where get_usage returns a dict instead of a list
+	# This happens when there's no log server or no aggregations data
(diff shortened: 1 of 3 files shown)


Aradhya-Tripathi:
We are now building on jammy regardless

https://github.com/frappe/press/blob/develop/press/docker/Dockerfile#L2C1-L3C1

PR #3646, opened 2025-10-17
fix(banner): Verifying if the site is a trial plan

Bowrna:
This doesn't rely on site plan changes.

### DIFF
--- a/press/saas/api/site.py
+++ b/press/saas/api/site.py
@@ -10,20 +10,20 @@ from press.saas.api import whitelist_saas_api
 @whitelist_saas_api
 def info():
 	is_fc_user = False
-	site = frappe.get_value("Site", frappe.local.site_name, ["plan", "trial_end_date", "team"], as_dict=True)
+	site = frappe.get_value(
+		"Site", frappe.local.site_name, ["trial_end_date", "standby_for_product", "team"], as_dict=True
+	)
 	site_user = frappe.request.headers.get("x-site-user")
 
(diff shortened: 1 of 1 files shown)


press/saas/api/site.py
```diff
@@ -10,20 +10,20 @@
...
-
+	plan = frappe.db.get_value("Product Trial", site.standby_for_product, "trial_plan")
 	return {
 		"is_fc_user": is_fc_user,
 		"name": frappe.local.site_name,
 		"trial_end_date": site.trial_end_date,
-		"plan": frappe.db.get_value("Site Plan", site.plan, ["is_trial_plan"], as_dict=True)
-		if site.plan
-		else None,
+		"plan": {"is_trial_plan": 1} if plan else None,
```
tanmoysrt, line 26: If the plan is available on site, we should send the data based on that.
Else, look at product_trial_plan

It could be something like this,

```python
plan = site.plan if site.plan else frappe.db.get_value("Product Trial", site.standby_for_product, "trial_plan")
```

And then you can do,
```python
"plan": frappe.db.get_value("Site Plan", plan, ["is_trial_plan"], as_dict=True)
```

Bowrna: i wrongly hit merged before resolving this part. @tanmoysrt let me raise another PR with changes you suggested

tanmoysrt: @Bowrna you can revert this pr then.

PR #3684, opened 2025-10-23
refactor(dashboard-banners): Replace dashboard banners scheduler with a time window query

regdocs:
- Replaced banner scheduler with a time window query, so banners show up exactly when they’re supposed to, if they're `enabled`.

- Removed publish/unpublish logic to avoid unnecessary database updates

- Updated the banner message field to Long Text


ssiyad (approved):
LGTM

PR #3712, opened 2025-10-28
fix(dashboard-banner): Show width 100% in server overview

tanmoysrt:


tanmoysrt:
@mergify backport master

PR #3729, opened 2025-10-29
fix(role): Added role Press Billing Manager to add/edit billing address

Bowrna:
Edit: Add new role "Press Billing Manager". Giving permission across doctypes related to billing. Giving appropriate permissions through settings page.


press/api/account.py
```diff
@@ -841,21 +841,14 @@ def get_billing_information(timezone=None):
...
-		):
-			frappe.throw("Cannot change country after registration")
-		team.update_billing_details(billing_details)
-	except Exception as ex:
-		log_error(
-			"Billing update failing",
-			data=ex,
-			reference_doctype="Team",
-		)
+	billing_details = frappe._dict(billing_details)
```
Bowrna, line 845: I have added this try block part by mistake in develop branch. This will swallow the issue without raising to the user

press/press/doctype/press_role/press_role.py
```diff
@@ -82,20 +82,32 @@ def set_admin_permissions(self):
...
 		user = frappe.get_doc("User", user)
-		user.append_roles("Press Admin")
+		user.append_roles(role)
 		user.save(ignore_permissions=True)
 
-	def remove_press_admin_role(self, user):
+	def add_press_billing_manager_role(self, user):
+		self.add_press_role(user, "Press Billing Manager")
+
+	def add_press_admin_role(self, user):
```
shadrak98, line 86: With Press Billing Manager, do we still need this method? I think we can safely remove it.

Bowrna: @shadrak98 This one is for giving "Admin Access" role. Segregated that role from billing manager role. So user with Press Admin / Press Billing Manager role can edit the billing address.

PR #3734, opened 2025-10-29
fix(dashboard): Change fold step background

ssiyad:
## Before
image

## After
image

### DIFF
--- a/dashboard/src/components/FoldStep.vue
+++ b/dashboard/src/components/FoldStep.vue
@@ -38,7 +38,7 @@
 			</div>
 		</button>
 		<div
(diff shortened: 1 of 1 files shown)


ssiyad:
Me too have a dark background with text text on terminal but that does not blend well with rest of the page here. Maybe we need dark mode for dashboard?

regdocs:
@ssiyad yeah good idea, lets add the dark theme to the backlog

ssiyad:
Merging. Hoping for dark mode soon.

BreadGenie:
While we are at it, can we add a timer to the ongoing step as well?

Issue #3553, opened 2025-10-14
Refactor `Archive Job` to avoid server status to be stuck to `Pending`


tanmoysrt:
- Try not to trigger another `Archive Server` Press job, if one already succeded
- In `Archive Server` press job, first check whether the server actually exists or not. If it already terminated, don't proceed further.

Example: https://cloud.frappe.io/app/press-job/111274

PR #3753, opened 2025-10-31
fix: Refactor Archive Server job

AyshaHakeem:
Do not run Archive Server job if it was recently successfully completed or if the server has already been terminated
#3553

### DIFF
--- a/press/press/doctype/server/server.py
+++ b/press/press/doctype/server/server.py
@@ -1010,6 +1010,25 @@ class BaseServer(Document, TagHelpers):
 
 	@frappe.whitelist()
 	def archive(self):
+		if frappe.db.exists(
+			"Press Job",
+			{
+				"job_type": "Archive Server",
+				"server": self.name,
+				"server_type": self.doctype,
+				"status": "Success",
+			},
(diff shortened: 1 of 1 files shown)


tanmoysrt:
If its Archived actually, you might not need to throw error
Just set the status to Terminated.

press/press/doctype/server/server.py
```diff
@@ -1026,6 +1026,26 @@ def archive(self):
...
+			"Press Job",
+			{
+				"job_type": "Archive Server",
+				"server": self.name,
+				"server_type": self.doctype,
+				"status": "Success",
+				"creation": (">", frappe.utils.add_to_date(None, days=-1)),
+			},
+		):
+			frappe.throw(_("Archive Server job already completed successfully recently."))
```
tanmoysrt: Maybe just show a message instead of throwing error and return.

Doesn't need filter on `creation`. As we don't reuse the record for re-creating servers.

AyshaHakeem: thanks @tanmoysrt, fixed this

PR #3827, opened 2025-11-07
fix(account)!: Disallow temporary email providers

ssiyad:
closes https://github.com/frappe/press/issues/3797


tanmoysrt:
@ssiyad Maybe we can use this [disposable](https://github.com/disposable/disposable) repo's domain list to filter out temp mails instead of maintaining the list ourselves.

There might be some library of it.

Secondly, maybe keep a feature flag at `Press Settings` to disable this ? Useful for any random purpose (development e.g.)

PR #3843, opened 2025-11-10
fix(dashboard-banners): Fix cluster scope permission issue

regdocs:
This is a workaround to prevent the cluster scoped banners' dismissal permission issue where the dismissal was failing for unprivileged Press Admin users but succeeding for System Managers. Need to investigate this further.

### DIFF
--- a/press/api/account.py
+++ b/press/api/account.py
@@ -1490,7 +1490,7 @@ def dismiss_banner(banner_name):
 				"parent": banner_name,
 			},
 		)
-		banner.save()
+		banner.save(ignore_permissions=True)
 		return True
 	return False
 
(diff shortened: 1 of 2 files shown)


press/api/account.py
```diff
@@ -1490,7 +1490,7 @@ def dismiss_banner(banner_name):
 				"parent": banner_name,
 			},
 		)
-		banner.save()
+		banner.save(ignore_permissions=True)
 		return True
 	return False
```
ssiyad, line 1495: I think this entire block should be inside the banner controller.

`ignore_permissions` is a bad way to fix things. Most probably the issue is with permissions on a specific banner, which can be fixed with `has_permission` hook.

regdocs: I agree. I had tried the controller thing initially but was hitting a permission road block. Will attempt it again and let you know how that goes.

regdocs:
Closing this, will create a clean PR with the fix integrated with the feature.

Issue #3864, opened 2025-11-12
Deploy UX - add app versions, redeploy button


netchampfaris:
On the deploy screen, I only have the basic information like deploy time and owner. It will be nice to see app versions too. Will help me decide if I want to do a new deploy with latest version or do a redeploy. Need a "Retry deploy" button too in case the deploy fails for a temporary reason (no space left on device)

Aradhya-Tripathi:
⚡

PR #3891, opened 2025-11-13
feat(deploy): Show app versions & redeploy

Aradhya-Tripathi:
ref: https://github.com/frappe/press/issues/3864

- Removed redundant redeploys
   - Showed `redeploy` when the build failed as a button
   - Showed `fail and redeploy` when the build was running in the drop down

- Shows redeploy
   - Show redeploy when build is in a `terminal` stage after showing app versions
   - Users will need to stop the build using the `stop build` button to now view redeploy options along with the app versions

Preview
Screenshot 2025-11-13 at 4 56 24 PM

Cases when redeploy is allowed:
1. Successful deploy
2. Failed deploy

Any other state only the versions are visible (since users might manually fail)
Screenshot 2025-11-13 at 4 57 40 PM

### DIFF
--- a/press/press/doctype/user_2fa/user_2fa.py
+++ b/press/press/doctype/user_2fa/user_2fa.py
@@ -3,6 +3,9 @@
 
 from __future__ import annotations
 
+import random
+import string
+
 import frappe
 import frappe.utils
 from frappe.model.document import Document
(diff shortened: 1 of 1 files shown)


dashboard/src/dialogs/AppVersionsDialog.vue
```diff
@@ -0,0 +1,139 @@
...
+				class="h-80"
+				:columns="[
+					{
+						label: 'App',
+						key: 'name',
+						width: 2,
+					},
+					{
+						label: 'Commit',
+						key: 'hash',
```
netchampfaris: should be rendered in monospace font?

Aradhya-Tripathi: 👍🏿

PR #3902, opened 2025-11-14
fix(monitor): Support arm64 during prometheus setup

tanmoysrt:


adityahase:
https://github.com/frappe/press/compare/develop...fix-monitor

PR #3904, opened 2025-11-14
fix(build): Remove stale lock files when installing custom packages

Aradhya-Tripathi:


adityahase:
Just a hunch (not supported by any documentation). These might not be stale locks BTW.

Two parallel builds might be doing apt-install at the same time. In that case clearing locks doesn't look like a good idea.

PR #3954, opened 2025-11-18
fix: Add amount validation to create_razorpay_order

AyshaHakeem:


press/api/billing.py
```diff
@@ -648,7 +648,11 @@ def is_paypal_enabled() -> bool:
 
 
 @frappe.whitelist()
-def create_razorpay_order(amount, type=None):
+def create_razorpay_order(amount, type=None) -> dict | None:
+	if amount <= 0:
+		frappe.msgprint(_("Amount should be greater than zero"))
+		return None
```
ssiyad, line 654: Why not throw an error?

AyshaHakeem: throwing an error seemed unnecessary, there is also a client side validation in place

PR #3963, opened 2025-11-18
feat(secondary-server): Start and stop secondary server

Aradhya-Tripathi:
- Start secondary when we attach nfs volume
- Stop secondary when attachment is completed
- Start secondary when we detach nfs volume
- Drop secondary when detachment is compeleted


Aradhya-Tripathi:
@mergifyIo backport master

PR #3966, opened 2025-11-18
feat(deploy): Allow auto deploy via commit markers

Aradhya-Tripathi:
Trigger app auto deploy if we pickup deploy marks on the commit message in the app release;
current deploy marker `fc-deploy` can be used to target release groups as well; `fc-deploy-bench-150`

If no bench group is passed all bench groups containing this app under the team will be deployed.

Deploy conditions
- No active running builds should be present for the bench group
- Bench group should have the app
- Bench group should belong to the same team

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -13,9 +13,7 @@ playbooks/ @adityahase @balamurali27
 site* @balamurali27
 team/ @shadrak98
 
-dashboard/ @breadgenie
 invoice/ @shadrak98
-marketplace* @breadgenie
 stripe* @shadrak98
 razorpay* @shadrak98
 subscription/ @shadrak98


press/press/doctype/app_release/app_release.py
```diff
@@ -101,6 +106,76 @@ def before_save(self):
...
+			return
+
+		if bench_group:
+			self._deploy_bench_group(bench_group)
+
+		else:
+			for bench_group in frappe.db.get_all(
+				"Release Group", {"team": self.team, "enabled": 1}, pluck="name"
+			):
+				self._deploy_bench_group(bench_group)
```
phot0n: can we restrict it via any flag user/team can put on press?

so that we don't deploy everything they have

press/press/doctype/app_release/app_release.py
```diff
@@ -101,6 +106,76 @@ def before_save(self):
 		if self.team in teams or self.app in apps:
 			self.status = "Approved"
 
+	def _has_auto_deploy_marker(self) -> tuple[bool, str | None]:
+		"""<deploy-marker>-<bench-group> | <deploy-marker>"""
+		deploy_marker = frappe.db.get_single_value("Press Settings", "deploy_marker", cache=True)
```
phot0n: make this hardcoded? not sure the implications but anyone can change it whenever they want if it's accessible via a setting on ui

press/press/doctype/app_release/app_release.py
Aradhya-Tripathi: Not sure what this means, restrict all bench group deploy?

phot0n: so somehow put some checkbox or something (like a label mangesh mentioned as well) on dashboard - so that we can only deploy those benches - not all the team/user owns (?)

Aradhya-Tripathi: Only deploying bench groups with the auto-deploy resource tag ([a8a22a6](https://github.com/frappe/press/pull/3966/commits/a8a22a6ebfb87f8d5de5ff304561b10a92337eb0))

press/press/doctype/app_release/app_release.py
Aradhya-Tripathi: Not sure if hardcoding is a good idea here, we can always just disable auto deploys be removing the field from press settings

PR #4012, opened 2025-11-20
fix(marketplace): Payout tracking and commission calculation

### LINKED ISSUE #3818: inconsistent marketplace developer revenue tracking
- Payouts or revenue tracking logic against a payout order in Marketplace App Payment is buggy
Commission and net amoutns per payout order items are added to total amount field inconsistently
- remove mandatory purchase order field - this could go into a custom script since purchase order is an erpnext record
- track commissions in a separate field

AyshaHakeem:
fix #3818

### DIFF
--- a/press/marketplace/doctype/marketplace_app_payment/marketplace_app_payment.py
+++ b/press/marketplace/doctype/marketplace_app_payment/marketplace_app_payment.py
@@ -30,8 +30,8 @@ class MarketplaceAppPayment(Document):
 		"""
 		press_settings = frappe.get_cached_doc("Press Settings")
 		exchange_rate = press_settings.usd_rate or 80
-		threshold = press_settings.threshold or 500
-		commission_rate = press_settings.commission or 0.20
+		threshold = press_settings.threshold
+		commission_rate = press_settings.commission
 
 		# Calculate current total in USD
 		current_total = self.total_usd + (self.total_inr / exchange_rate)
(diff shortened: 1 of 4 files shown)


press/marketplace/doctype/marketplace_app_payment/marketplace_app_payment.py
```diff
@@ -6,17 +6,52 @@
...
+		"""
+		press_settings = frappe.get_cached_doc("Press Settings")
+		exchange_rate = press_settings.usd_rate or 80
+		threshold = press_settings.threshold or 500
+		commission_rate = press_settings.commission or 0.20
+
+		# Calculate current total in USD
+		current_total = self.total_usd + (self.total_inr / exchange_rate)
+
+		if current_total >= threshold:
```
shadrak98:
```suggestion
		if current_total > threshold:
```
We can only check for threshold > total and not equals to since it will be handled later

press/press/doctype/payout_order/payout_order.py
```diff
@@ -57,87 +59,126 @@ class PayoutOrder(Document):
...
-		return query
+		return query.where(PayoutOrder.docstatus != 2)
 
 	def validate(self):
-		self.validate_items()
-		self.validate_net_totals()
-		self.compute_total_amount()
-
-	def validate_items(self):
 		for row in self.items:
```
shadrak98: this shouldn't be under validate but rather some `calculate_net_total`

Nothing is being validated here

Issue #3395, opened 2025-09-26
refactor: Binlog Browser


tanmoysrt:
**Tasks -**

-  Use [systemd-run](https://www.freedesktop.org/software/systemd/man/latest/systemd-run.html) to run the program with limits to throttle io and cpu usage
-  Update the chart with new component from frappe-ui
- On clicking each bar of chart, timerange should auto update
- Restrict binlog indexer from using too much resource (Use systemd or docker)
- Change some logics
  - Make S3 upload as optional
  - In disk keep only 1 days of binlog, If S3 enabled
  - Don't complain if some binlogs are missing (might have purge)

PR #4066, opened 2025-11-27
refactor(database): Refactor binlog browser

tanmoysrt:
Resolves https://github.com/frappe/press/issues/3395

- [x] Use systemd-run for limiting resource usage
   - Without systemd-run, small server can face issue.
- [x] Improved Indexer
   - Now it has low-memory, balanced, high-compression mode to accomodate with system resources  
- [x] Update charts and have data zoom feature from echarts
- [x] Replace the custom date picker with new one
- [x] Ensure to sort out table names in the list [better to have search box]
- [x] Persist state in url
- [x] Event size based filtering
- [x] Store indexed binlog files in data disk
- [x] Add option to disable/enable few stuffs for end-user (For replica we need to keep binlogs, but we can have indexer there as well)
  - [x] Binlog uploading (can be chargeable)
- [x] Make sure to keep binlog purging configuration across everything same 
- [x] Store index of 14 days
- [x] Handle gap in data due to purging binlogs by user / size limit. Need to descriptive in UI as well.
- [x] Increase the searching window and be more ux friendly
	- Provide larger window for dedicated server.
- [x] Show maintenance / upload schedule as well in UI.
  - DuckDB doesn't allow read during write. So, while indexing binlogs  show notification to user that indexer is unavailable for that particular window.
  - There is an workaround as well. Copy main DuckDB and do the modification there and once done, just replace. 
- [x] On-demand indexing (sometimes we need to index some binlogs quickly for investigation regardless of resource usage)
- [x] Add a status window in binlog browser to show binlogs and index statuses

---

https://github.com/user-attachments/assets/06733909-b5bb-4f7e-af09-4857387bac7d

image

image

image

On-demand index binlogs, helpful for investigation where indexer wasn't enabled previously.

image

Agent PR : https://github.com/frappe/agent/pull/383

### DIFF
--- a/dashboard/src/components/server/ServerActionCell.vue
+++ b/dashboard/src/components/server/ServerActionCell.vue
@@ -110,10 +110,15 @@ function onTeardownSecondaryServer() {
 					After teardown, autoscaling will <strong>not trigger</strong> unless you configure a new secondary server.
 				</p>
 
-				<p class="mt-3 text-gray-600">
-					See the docs to learn more about autoscaling:
-					<a href="#" target="_blank" rel="noopener">Secondary Server teardown guide</a>.
+				<p class="mt-3">
+					See the docs to learn more about autoscaling:<br>
+					<a href="https://docs.frappe.io/cloud/application-server-horizontal-scaling#opting-out"
+					target="_blank" rel="noopener" style="text-decoration: underline;">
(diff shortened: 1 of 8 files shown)


regdocs (approved):
Thanks, I'll try to make my charts, and filters resemble your binlog echart for consistency.

PR #4089, opened 2025-12-02
feat(pricing): Auto scale pricing

Aradhya-Tripathi:
- Discount on original app server price (10%)
- Ignore usage record creation in the subscription doctype
- Create usage record on scale down event calculation scaled duration using the last successful scale up record.
- Adjusted quantity in the Invoice as the total usage minutes.

### DIFF
--- a/dashboard/src/components/server/ServerOverview.vue
+++ b/dashboard/src/components/server/ServerOverview.vue
@@ -103,6 +103,7 @@
 											$appServer?.doc?.status === 'Active' &&
 											$appServer?.doc?.secondary_server
 										"
+										:disabled="startedScaleUp"
 										@click="scaleUp()"
 										label="Scale Up"
 									/>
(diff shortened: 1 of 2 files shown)


press/press/doctype/auto_scale_record/auto_scale_record.py
```diff
@@ -357,6 +367,64 @@ def initiate_secondary_shutdown(self, step: "ScaleStep"):
...
+				"secondary_server": self.secondary_server,
+				"status": "Success",
+				"action": "Scale Up",
+			},
+			"modified",
+		)
+
+		auto_scaled_for = frappe.utils.now_datetime() - last_auto_scale_at
+		auto_scaled_for = auto_scaled_for.total_seconds() / 3600
+		secondary_server_plan = frappe.db.get_value("Server", self.secondary_server, "plan")
```
shadrak98: This can be written as 
`secondary_server_plan, secondary_server_team = frappe.db.get_values("Server", self.secondary_server, ("plan", "team"))`

PR #4090, opened 2025-12-02
feat(site): Allow extending trial for 7 days

shadrak98:


dashboard/src/objects/site.js
```diff
@@ -1712,6 +1713,48 @@ export default {
...
+								) {
+									throw new Error('Reason is required');
+								}
+								return site.extendTrial
+									.submit({ reason: values.reason })
+									.then((res) => {
+										console.log(res);
+										hide();
+										toast.success('Trial extended by 7 days');
+									});
```
phot0n, line 1753: if people are going to be able to extend this *without our approval* - might as well just increase the default trial duration?

PR #4102, opened 2025-12-04
chore(job): Pass param to backup site job to determine agent_job_timeout

Bowrna:
@tanmoysrt Do you think we should give each site backup specific timeout or common backup timeout through configuration in Press Settings? Please do let me know if there are other ways that you think that would fit.

### DIFF
--- a/press/agent.py
+++ b/press/agent.py
@@ -278,7 +278,7 @@ class Agent:
 		assert site.config is not None, "Site config is required to restore site from backup"
 
 		data = {
-			"config": json.loads(site.config),
+			"config": json.loads(site.config) if site.config else {},
 			"apps": apps,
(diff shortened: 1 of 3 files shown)


press/agent.py
```diff
@@ -533,7 +533,7 @@ def backup_site(self, site, site_backup: SiteBackup):
 		from press.press.doctype.site_backup.site_backup import get_backup_bucket
 
 		data = {"with_files": site_backup.with_files}
-
+		data.update({"agent_job_timeout": site.backup_timeout})
```
tanmoysrt:
```suggestion

data = {
	"with_files": site_backup.with_files,
	"agent_job_timeout": site.backup_timeout
}
```

PR #4110, opened 2025-12-05
fix(site): Block actions except archive/restore if new-site job fails

AyshaHakeem:
When New Site or New Site From Backup agent job fails block all site actions except restoring from backup or archiving

- Add creation_failed checkbox to site doctype. 
   - Enable it when site creation jobs fail and disable if restored.
   - Update @site_action decorator to block all actions except archive and restore when this is enabled
   - Disable the field if site is restored.
- Display a banner to convey the same to the user
- Add background job to archive sites with failed creation after  7 days
Screenshot 2025-12-10 at 12 04 33 PM

### DIFF
--- a/dashboard/src/components/SiteOverview.vue
+++ b/dashboard/src/components/SiteOverview.vue
@@ -14,7 +14,7 @@
 			v-if="$site?.doc?.creation_failed"
 			class="col-span-1 lg:col-span-2"
 			type="error"
-			title="Site creation failed. You can restore the site from a backup or drop this site to create a new one."
+			:title="`Site creation failed. You can restore the site from a backup or drop this site to create a new one. The site will be automatically dropped after ${$site?.doc?.creation_failure_retention_days} days if not restored.`"
 		>
 		</AlertBanner>
 
--- a/press/press/doctype/site/site.py
+++ b/press/press/doctype/site/site.py
@@ -330,6 +330,7 @@ class Site(Document, TagHelpers):
 		doc.current_usage = self.current_usage
 		doc.current_plan = get("Site Plan", self.plan) if self.plan else None
 		doc.last_updated = self.last_updated
+		doc.creation_failure_retention_days = CREATION_FAILURE_RETENTION_DAYS
 		doc.has_scheduled_updates = bool(
 			frappe.db.exists("Site Update", {"site": self.name, "status": "Scheduled"})
 		)
@@ -4824,7 +4825,7 @@ def archive_creation_failed_sites():
 	filters = [
 		["creation_failed", "!=", None],
 		["creation_failed", "<", creation_failure_retention_date],
-		["status", "!=", "Archived"],
+		["status", "=", "Broken"],
 	]
 
 	failed_sites = frappe.db.get_all("Site", filters=filters, fields=["name"], pluck="name")
@@ -4833,7 +4834,7 @@ def archive_creation_failed_sites():
 		try:
 			site = Site("Site", site)
 			site.archive(
-				reason="Site creation failed and was not restored within 7 days",
+				reason=f"Site creation failed and was not restored within {CREATION_FAILURE_RETENTION_DAYS} days"
 			)
 			frappe.db.commit()
 		except Exception:


press/press/doctype/site/site.json
```diff
@@ -814,8 +822,8 @@
    "link_fieldname": "site"
   }
  ],
- "modified": "2025-10-07 19:14:24.341019",
- "modified_by": "Administrator",
+ "modified": "2025-12-05 13:49:05.307868",
+ "modified_by": "aysha@frappe.io",
```
regdocs: Not really an issue but workspace email in the codebase makes me uneasy 😅

regdocs (approved):

AyshaHakeem:
can you please review this @balamurali27

press/press/doctype/site/site.py
```diff
@@ -4090,6 +4112,8 @@ def process_restore_job_update(job, force=False):
 			site = Site("Site", job.site)
 			process_marketplace_hooks_for_backup_restore(set(apps_from_backup), site)
 			site.set_apps(apps_from_backup)
+			if frappe.db.get_value("Site", job.site, "creation_failed"):
+				frappe.db.set_value("Site", job.site, "creation_failed", 0)
```
balamurali27: The check isn't needed. Update query checks implicitly, I believe.

press/press/doctype/site/site.py
```diff
@@ -4647,3 +4671,31 @@ def create_subscription_for_trial_sites():
...
+		"Site",
+		filters=filters,
+		fields=["name"],
+	)
+	if not failed_sites:
+		return
+
+	for site in failed_sites:
+		try:
+			site = frappe.get_doc("Site", site.name)
```
balamurali27:
```suggestion
			site = Site("Site", site.name)
```

This is also doable. Allows for better typing

balamurali27:
fixes #774

press/press/doctype/site/site.json
```diff
@@ -735,6 +736,13 @@
    "fieldtype": "Data",
    "label": "Reason for Disabling Monitoring",
    "read_only": 1
+  },
+  {
+   "default": "0",
+   "fieldname": "creation_failed",
+   "fieldtype": "Check",
+   "label": "Creation Failed",
+   "read_only": 1
```
balamurali27, line 744: Instead of checkbox, we can do a timestamp field. That way, we can set the same, for Restore jobs (Restore Site, New Site from Backup) as well; Using the timestamp itself to check if we need to archive such sites in the cron job.

press/press/doctype/site/site.py
balamurali27: Need to set creation failure here for failure case

press/press/doctype/site/site.py
```diff
@@ -4647,3 +4671,31 @@ def create_subscription_for_trial_sites():
 		except Exception:
 			frappe.db.rollback()
 			log_error(title="Creating subscription for trial sites", site=trial_site)
+
+
+def archive_creation_failure_sites():
+	seven_days_ago = frappe.utils.add_days(frappe.utils.today(), -7)
```
balamurali27: Can make it longer. 14 days perhaps.

The number 7 should be a capitalized constant variable at top of file like CREATION_FAILURE_GRACE_DAYS

press/press/doctype/site/site.json
balamurali27: FF might put a default value in timestamp field, so please double check

balamurali27 (dismissed):

press/press/doctype/site/site.py
```diff
@@ -4738,3 +4767,28 @@ def create_subscription_for_trial_sites():
...
+		["status", "!=", "Archived"],
+	]
+
+	failed_sites = frappe.db.get_all("Site", filters=filters, fields=["name"], pluck="name")
+
+	for site in failed_sites:
+		try:
+			site = Site("Site", site)
+			site.archive(
+				reason="Site creation failed and was not restored within 7 days",
```
balamurali27: 7 needs to be replaced with variable

dashboard/src/components/SiteOverview.vue
```diff
@@ -10,6 +10,14 @@
 			:ctx_name="$site?.doc?.name"
 		/>
 
+		<AlertBanner
+			v-if="$site?.doc?.creation_failed"
+			class="col-span-1 lg:col-span-2"
+			type="error"
+			title="Site creation failed. You can restore the site from a backup or drop this site to create a new one."
```
balamurali27: Here also, we should mention that it'll be dropped in x days

press/press/doctype/site/site.py
```diff
@@ -4738,3 +4767,28 @@ def create_subscription_for_trial_sites():
...
+
+def archive_creation_failed_sites():
+	creation_failure_retention_date = frappe.utils.add_days(
+		frappe.utils.now(), -CREATION_FAILURE_RETENTION_DAYS
+	)
+
+	filters = [
+		["creation_failed", "!=", None],
+		["creation_failed", "<", creation_failure_retention_date],
+		["status", "!=", "Archived"],
```
balamurali27: This feels dangerous. Checking for Broken is better

Issue #4101, opened 2025-12-04
Using custom domain while generating invoices.


huzan-kazi:
Currently, if the customer site is a standby one, then the invoice is generated using the standby site's name itself. A lot of the time, the customers demand that the invoice should be generated using the site's primary custom domain.

This only applies to standby sites.

PR #4112, opened 2025-12-05
fix(invoice): Rename site with hostname if present to avoid confusion

Bowrna:
related: #4101 
fixes: #4101


barredterra:
@Bowrna @shadrak98 this makes integrations difficult. I import sites and invoices via API. Now I can no longer match them because `document_name` holds a value that is not the document name.

barredterra:
I guess an additional field `document_title` would work well  (for the hostname, in the case of Sites)

PR #4123, opened 2025-12-06
fix: Mark site update doc as cancelled when cancelling an update rather than deleting it

phot0n:
- [x] probably dont allow editing a scheduled update if it has been cancelled? (check siteupdatedialog.vue) -- there is a check for status to be scheduled so it should be fine for the most part i guess

### DIFF
--- a/press/press/doctype/site/site.py
+++ b/press/press/doctype/site/site.py
@@ -1276,14 +1276,14 @@ class Site(Document, TagHelpers):
 	def cancel_scheduled_update(self, site_update: str):
 		try:
 			if (
-				_status := frappe.db.get_value("Site Update", site_update, "status", for_update=True, wait=False)
+				_status := frappe.db.get_value(
+					"Site Update", site_update, "status", for_update=True, wait=False
+				)
 			) != "Scheduled":
 				frappe.throw(f"Cannot cancel a Site Update with status {_status}")
 
 		except (frappe.QueryTimeoutError, frappe.QueryDeadlockError):
-			frappe.throw(
-				"The update is probably underway. Please reload/refresh to get the latest status."
-			)
+			frappe.throw("The update is probably underway. Please reload/refresh to get the latest status.")
 
 		# used document api for applying doc permissions
 		doc = frappe.get_doc("Site Update", site_update)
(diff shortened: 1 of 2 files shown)


press/press/doctype/site_update/site_update.py
```diff
@@ -466,7 +464,7 @@ def has_pending_updates(self):
 			"Site Update",
 			{
 				"site": self.site,
-				"status": ("in", ("Pending", "Running", "Failure", "Scheduled", "Recovering")),
```
phot0n, line 469: not sure about this but by just reading the code this didn't make sense to me

maybe this means something else?

phot0n: prolly shouldn't touch unrelated stuff :P

phot0n:
a bit unsure about things like:

https://github.com/frappe/press/blob/develop/press/press/doctype/site_update/site_update.py#L453-L462

and 

https://github.com/frappe/press/blob/develop/press/press/doctype/site_update/site_update.py#L893-L901

these dont have a filter for status

press/press/doctype/site/site.py
```diff
@@ -1274,11 +1274,21 @@ def edit_scheduled_update(
...
 
-		# TODO: Set status to cancelled instead of deleting the doc
-		frappe.delete_doc("Site Update", site_update)
+		except (frappe.QueryTimeoutError, frappe.QueryDeadlockError):
+			frappe.throw(
+				"The update is probably underway. Please reload/refresh to get the latest status."
+			)
+
+		# used document api for applying doc permissions
+		doc = frappe.get_doc("Site Update", site_update)
```
phot0n, line 1289: no option of nowait in this plus we need to respect perms as well (user facing)

press/press/doctype/site_update/site_update.js
```diff
@@ -2,17 +2,6 @@
...
-		frm.set_query('destination_bench', function () {
-			return {
-				filters: {
-					status: 'Active',
-					server: frm.doc.server,
-				},
-			};
-		});
-	},
-
```
phot0n, line 15: site update is a "log" - not sure why this was needed

press/press/doctype/site/site.py
phot0n: will need to reduce this code duplication

PR #4137, opened 2025-12-08
fix(billing): Add order type validation for razorpay order

AyshaHakeem:
- Validate against  plan and invoice documents for applicable transactions
- Check minimum amount threshold is met for other Prepaid Credits


ssiyad (changes requested):
Do not use `type` as variable name because a [builtin](https://www.w3schools.com/python/ref_func_type.asp) with that name already exists.

PR #4195, opened 2025-12-11
fix(step-handler): Better VM sync & commit post step completion

Aradhya-Tripathi:


press/runner.py
```diff
@@ -324,8 +324,11 @@ def handle_vm_status_job(
 		step.attempt = 1 if not step.attempt else step.attempt + 1
 
 		# Try to sync status in every attempt
-		with suppress(Exception):
-			frappe.get_doc("Virtual Machine", virtual_machine).sync()
+		try:
+			virtual_machine_doc: "VirtualMachine" = frappe.get_doc("Virtual Machine", virtual_machine)
+			virtual_machine_doc.sync()
+		except Exception:
+			pass
```
phot0n, line 331: i mean ..lol

Aradhya-Tripathi: We generally need to sync asap, and sync sometimes errors out with a timestamp mismatch error.

phot0n: I meant it does the exact same thing as the code you removed... something was screaming in precommit so i did it that way lol

Aradhya-Tripathi: No mypy was upset as `vm_machine` changes from `str` to `Document` good catch actually.

phot0n: Correct ..just that that code was correct which you removed lol... i had taken care of it in my proxy failover pr .. you just reverted it again lol and appended "doc" to var name

Anyhoo forget it

PR #4204, opened 2025-12-12
feat(import-errors): Catch import errors in custom apps

Aradhya-Tripathi:
Import errors in custom apps top the chart in site migrate failures, catch them in pre-build validations

image


Aradhya-Tripathi:
Need to do this on the agent, as this process is very compute hungry.

PR #4206, opened 2025-12-12
feat: Virtual disk resize from dashboard

phot0n:
- [x] notification on faliure
- [ ] guess mountpoint magic
- [ ] check if reduction should be allowed (based on backup size)

### DIFF
--- a/press/infrastructure/doctype/virtual_disk_resize/virtual_disk_resize.py
+++ b/press/infrastructure/doctype/virtual_disk_resize/virtual_disk_resize.py
@@ -619,7 +619,7 @@ class VirtualDiskResize(Document):
 		self.save()
 
 		create_new_notification(
-			frappe.get_doc("Virtual Machine", self.virtual_machine).team,
+			frappe.db.get_value("Virtual Machine", self.virtual_machine, "team"),
 			"Disk Resize",
 			self.doctype,
 			self.name,


press/api/server.py
```diff
@@ -806,3 +806,37 @@ def create_record(action: str, scheduled_time: datetime) -> None:
...
+	frappe.get_doc(
+		{
+			"doctype": "Virtual Disk Resize",
+			"virtual_machine": server.virtual_machine,
+			"old_volume_id": volume_id,
+			"expected_volume_size": expected_volume_size,
+			"scheduled_datetime": scheduled_datetime,
+			"status": "Scheduled",
+		}
+	).insert()
```
tanmoysrt, line 842: You can move this method on `BaseServer` class in `doctype/server/server.py` and apply the `@dashboard_whitelist()` decorator for permission checks.

Or, you can override the query condition and has_permission hook of `Virtual Disk Resize` as well
We have couple of those of other doctypes here : https://github.com/phot0n/press/blob/afec0fb33e65d4c6e47e3a394b869307684934df/press/hooks.py#L105-L158

press/api/server.py
```diff
@@ -806,3 +806,37 @@ def create_record(action: str, scheduled_time: datetime) -> None:
 
 	create_record("Scale Up", formatted_scheduled_scale_up_time)
 	create_record("Scale Down", formatted_scheduled_scale_down_time)
+
+
+@frappe.whitelist()
+@protected(["Server", "Database Server"])
+def schedule_disk_resize(
```
tanmoysrt, line 814: For sanity, we should check, whether if any Site Update / Restore / Recovery job is running on the server. Else, that can break something else due to resize.

phot0n: so we'll have to check this in them as well - if any disk resize is running then don't allow to schedule them right?

phot0n:
might not be needed anymore

Issue #2662, opened 2025-04-27
Plan upgrade for MariaDB to 11.8 or 11.4


ankush:
- 11.8 is the next planned LTS release, 11.4 is the current LTS, but that has a longer planned lifetime.
- 10.6 will be EOL next year
- We should wait till 11.8 is released and stable, and then switch all of our servers.

We need to pick between 11.4 and 11.8 as a minimum supported version for Framework so it's reasonably future-proof. 

![Image](https://github.com/user-attachments/assets/3827c882-f585-45e9-9289-b962c6a073dc)

Plenty of relevant notes here: https://github.com/frappe/frappe/pull/32289

ankush:
Note: Use mariadb's PPA for installing client libraries. Ubuntu ones are older and have a bug that causes ~30ms of latency on each connection (!)


https://github.com/frappe/frappe/pull/32289 (see change in CI)

iron3oxide:
closed by #6078

PR #4345, opened 2025-12-23
feat: Ubuntu 24.04 (Noble) compatibility for server setup

mn-frappe:
## Summary
This PR adds comprehensive Ubuntu 24.04 (Noble Numbat) support for Frappe Press server provisioning, addressing multiple breaking changes in the new LTS release.

### Ansible Roles - APT Key Management
- **docker**: Use modern GPG keyring approach instead of deprecated apt_key
- **mariadb**: Skip MariaDB repo setup on Noble (use Ubuntu's repos with MariaDB 10.11), use modern GPG keyring
- **mariadb_memory_allocator**: Remove deprecated apt_key task, add directory creation for systemd drop-in
- **filebeat**: Use modern GPG keyring approach

### Ansible Roles - Package/Binary Installation
- **node_exporter**: Replace unarchive with remote_src (broken with Python 3.12) with curl+tar, add idempotency check
- **mysqld_exporter**: Same fix as node_exporter
- **nginx**: Same fix for brotli module download

### Ansible Roles - Service/Config Changes
- **gpg_config**: Handle both legacy /etc/apt/trusted.gpg and modern keyrings directory
- **sshd_hardening**: Use 'ssh' service name on Noble instead of 'sshd'

### Python Code Fixes
- **server.py**: Fix `create_dns_record()` to handle empty domain gracefully
- **runner.py**: 
  - Handle `ignore_errors` parameter in `v2_runner_on_failed` callback (use 'Skipped' status instead of invalid 'Ignored')
  - Convert list messages to strings for debug tasks (fixes "Value for Exception cannot be a list" error)

## Breaking Changes Addressed
| Issue | Solution |
|-------|----------|
| `apt_key` module deprecated | GPG keys now stored in `/usr/share/keyrings/` or `/etc/apt/keyrings/` |
| MariaDB 10.6 repo doesn't exist for Noble | Use Ubuntu's MariaDB 10.11 from default repos |
| `libmariadbclient18` renamed | Changed to `libmariadb3` |
| Python 3.12 HTTPSConnection changes break Ansible's `unarchive` with `remote_src` | Use `curl` + `tar` shell commands |
| `/etc/apt/trusted.gpg` no longer exists | Conditional check for legacy vs modern keyring locations |
| `sshd` service renamed to `ssh` on Noble | Use distribution-aware service name |
| systemd drop-in directories may not exist | Create directories before placing config files |

## Testing
- ✅ Tested on Ubuntu 24.04 with Python 3.12 and Ansible 7.7.0
- ✅ Successfully provisioned Proxy Server 
- ✅ Successfully provisioned Database Server
- ✅ All 197 Ansible tasks complete without stuck 'Running' status

## Related
This supersedes #4344 with additional fixes for:
- Database server provisioning (MariaDB, mysqld_exporter)
- Ansible callback issues causing tasks to remain in "Running" status

### DIFF
--- a/press/playbooks/roles/cadvisor/tasks/main.yml
+++ b/press/playbooks/roles/cadvisor/tasks/main.yml
@@ -5,21 +5,29 @@
     state: directory
     mode: 0755
 
+- name: Check if cAdvisor binary exists
+  stat:
+    path: /opt/cadvisor/cadvisor
+  register: cadvisor_binary
+
+# Use curl instead of get_url to avoid Python 3.12 HTTPSConnection issues
 - name: Download cAdvisor Binary (amd64)
-  get_url:
-    url: https://github.com/google/cadvisor/releases/download/v0.50.0/cadvisor-v0.50.0-linux-amd64
-    checksum: sha256:c63976400ca980b4b34bd074ef93c5321650050e45e6dc4d0cb1baf7dc772639
-    dest: /opt/cadvisor/cadvisor
-    mode: 0755
-  when: ansible_architecture == "x86_64"
+  shell: |
+    curl -sL -o /opt/cadvisor/cadvisor https://github.com/google/cadvisor/releases/download/v0.50.0/cadvisor-v0.50.0-linux-amd64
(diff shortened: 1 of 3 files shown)


iron3oxide:
Since switching to a newer MariaDB version is necessary for this PR, would it be possible to use the latest LTS version, 11.8? This would then also close #2662.

PR #4352, opened 2025-12-26
fix: December fixes 1

balamurali27:
- **chore(weekend-support): Add more engineers**
  Also remove unused code
  

- **fix(nginx): Allow installing on servers without ipv6**
  The default nginx config file tries to listen to `::` which breaks
  during dpkg configure
  

- **fix(server): Add prepare method for Vodacom**
  

- **fix(server): Add Vodacom as provider option**
  

- **fix(bench-group): Inform user about USD 25 or higher plan on creation**
  Users should be made aware of the restrictions up-front
  

- **fix(bench): Remove scheduler from processlist**
  - It's always exited (v14 and up)
  - Promotes misconception to user that scheduler is not running.
  

- **refactor: Remove nondeterministic checks**
  

- **fix(site-migration): Start on insert if no scheduled time is set**
  This is consistent with behaviour of scheduled site update.
  
  Also, after such insertion, it can run and conflict with site update at
  same time.
  

- **fix(site-update): Allow cancelling pending ones from desk**
  

- **fix(bench): Hide scheduler process when appropriate**
  Only when it's in Exited state (expected) and v14 and above
  

- **fix(bench): Don't create if private_ip not available**

### DIFF
--- a/.github/helper/install.sh
+++ b/.github/helper/install.sh
@@ -19,7 +19,7 @@ sed -i 's/redis_socketio:/# redis_socketio:/g' Procfile
 
 bench get-app press "${GITHUB_WORKSPACE}"
 
-bench setup requirements
+bench setup requirements --dev
 
 CI=Yes bench build --app frappe &
 bench new-site --no-mariadb-socket --db-root-password root --admin-password admin test_site


press/press/doctype/site_update/site_update.js
```diff
@@ -47,7 +47,9 @@ frappe.ui.form.on('Site Update', {
 			__('Change Status'),
 			() => {
 				let options = ['Success', 'Recovered', 'Failure', 'Fatal'];
-				frm.doc.status === 'Scheduled' ? options.push('Cancelled') : null;
+				['Scheduled', 'Pending'].includes(frm.doc.status)
+					? options.push('Cancelled')
+					: null;
```
phot0n, line 52: we should not do this @balamurali27 imo

if the status is pending that means one of these is already on their way: https://github.com/frappe/press/blob/467699eca90158269d786e1bc8aa783937282804/press/press/doctype/site_update/site_update.py#L353-L358

phot0n: "cancel" only makes sure if the thing is scheduled - just set it to cancelled so that it can't run

plus this change wont really have any effect:

https://github.com/frappe/press/blob/8ae45fb751b60ca1f23e26e9b7c79967033c113a/press/press/doctype/site_update/site_update.py#L728-L741

Issue #4386, opened 2026-01-05
remove public ips from servers (except proxy, standalone)


phot0n:
this is specifically for AWS (for now - dunno about other providers)

reason: https://aws.amazon.com/blogs/aws/new-aws-public-ipv4-address-charge-public-ip-insights/

phot0n:
- [ ] ansible 
      https://github.com/frappe/press/blob/a9f88a2314b0e98e3b0b73392b756b3664bfcb13/press/runner.py#L200-L214
- [ ] agent
- [ ] monitoring

PR #4402, opened 2026-01-06
feat: Provision for removing public ip for app and db servers

phot0n:
addresses: https://github.com/frappe/press/issues/4386

agent: https://github.com/frappe/agent/pull/412

need to implement the automatic nat failover - prolly do in a separate pr

### DIFF
--- a/press/press/doctype/ansible_console/ansible_console.py
+++ b/press/press/doctype/ansible_console/ansible_console.py
@@ -201,6 +201,9 @@ class AnsibleAdHoc:
 			server_type = (
 				SERIES_TO_SERVER_TYPE.get(matched_key) if matched_key else SERIES_TO_SERVER_TYPE.get(src[0])
 			)
+			if not server_type:
+				continue
+
 			type_to_sources.setdefault(server_type, []).append(src)
 
 		all_servers = []


press/agent.py
```diff
@@ -1193,7 +1216,8 @@ def get_jobs_status(self, ids):
 		return status
 
 	def get_jobs_id(self, agent_job_ids):
-		return self.get(f"agent-jobs/{agent_job_ids}")
+		ids = ",".join(agent_job_ids) if isinstance(agent_job_ids, (list, tuple)) else agent_job_ids
+		return self.get(f"agent-jobs/{ids}")
```
phot0n, line 1238: dunno why this started failing just now

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -518,7 +518,11 @@ def _provision_aws(self):  # noqa: C901
 			},
 			"NetworkInterfaces": [
 				{
-					"AssociatePublicIpAddress": True,
+					"AssociatePublicIpAddress": self.series
+					not in (
+						"f",
+						"m",
+					),
```
phot0n: make this opt out

press/press/doctype/server/server.py
```diff
@@ -536,7 +538,7 @@ def create_dns_record(self):
 								"Name": self.name,
 								"Type": "A",
 								"TTL": 3600 if self.doctype == "Proxy Server" else 300,
-								"ResourceRecords": [{"Value": self.ip}],
+								"ResourceRecords": [{"Value": self.ip or self.private_ip}],
```
phot0n, line 569: could be problematic?

press/agent.py
```diff
@@ -1006,6 +1011,25 @@ def raw_request(self, method, path, data=None, raises=True, timeout=None):
...
+			server_ip, server_private_ip = frappe.db.get_value(
+				self.server_type, self.server, ("ip", "private_ip")
+			)
+			if not server_ip and server_private_ip:
+				cluster = frappe.db.get_value(self.server_type, self.server, "cluster", cache=True)
+				proxy = frappe.db.get_value(
+					"Proxy Server",
+					{"status": "Active", "cluster": cluster, "use_as_proxy_for_agent_and_metrics": 1},
+				)
+
```
phot0n, line 1043: should we error out here if we dont get a proxy?

press/agent.py
```diff
@@ -1006,6 +1011,30 @@ def raw_request(self, method, path, data=None, raises=True, timeout=None):
 			response.raise_for_status()
 		return json_response
 
+	def _get_request_url(self, path):
+		if self.server_type in ("Server", "Database Server"):
+			proxy = None
+			server_ip, server_private_ip = frappe.db.get_value(
+				self.server_type, self.server, ("ip", "private_ip")
+			)
+			if not server_ip and server_private_ip and not frappe.flags.in_test:
```
phot0n, line 1034: i dont like that i have to resort to in_test flag for the tests to pass

but in our tests i think we by default dont set any public ip

phot0n: should i fix the tests? or should i rely on this flag?

phot0n: maybe i should have another flag?

press/playbooks/roles/nat/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+          {{ network_interface.stdout }}:
+            dhcp4: true
+            addresses:
+              - {{ primary_ip }}/16
+              - {{ secondary_ip }}/16
+  when: secondary_ip is defined and secondary_ip
+
+- name: Apply Netplan
+  command: netplan apply
+  when: secondary_ip is defined and secondary_ip
```
phot0n, line 38: https://easyscale.net/aws-ec2-ubuntu-18-04-secondary-ip-with-netplan/

phot0n: we can also do ` sudo ip addr add {{secondary_ip}}/16 dev eth0`

press/press/doctype/virtual_machine/virtual_machine.js
```diff
@@ -3,9 +3,53 @@
...
+		}
+
 		[
 			[__('Sync'), 'sync', false, frm.doc.status != 'Draft'],
-			[__('Provision'), 'provision', true, frm.doc.status == 'Draft'],
+			[
+				__('Provision'),
+				'provision',
+				true,
+				frm.doc.status == 'Draft' && frm.doc.cloud_provider != 'AWS EC2',
```
tanmoysrt, line 51: `Provision` is required locally, when we don't have any VMI and need to use `Setup Server` ansible scripts.
At that time, need to provision server manually.

press/agent.py
tanmoysrt: Modify the tests instead to support this new uri format as well : https://github.com/frappe/press/blob/develop/press/press/doctype/agent_job/test_agent_job.py#L94-L97

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -1364,6 +1410,73 @@ def update_name_tag(self, name):
...
+		ec2 = self.client()
+		instance = ec2.describe_instances(InstanceIds=[self.instance_id])
+		ec2.modify_network_interface_attribute(
+			NetworkInterfaceId=instance["Reservations"][0]["Instances"][0]["NetworkInterfaces"][0][
+				"NetworkInterfaceId"
+			],
+			AssociatePublicIpAddress=False,
+		)
+		frappe.flags.force_update_dns = True
+		self.sync()
```
tanmoysrt, line 1517: Take a lock on this vm doctype before doing this operation ?
Because `sync` can cause TimestampMismatchError.

There are some background workers which also do bulk vm syncing. That can cause issue with our `sync` if we dont take the lock.

Same for other other added functions also,

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -623,6 +646,12 @@ def _provision_aws(self):  # noqa: C901
 		self.status = self.get_aws_status_map()[response["Instances"][0]["State"]["Name"]]
 		self.save()
 
+		if self.series == "nat":
+			# Disable Source/Dest Check for NAT instances
+			response = self.client().modify_instance_attribute(
+				InstanceId=self.instance_id, SourceDestCheck={"Value": False}
+			)
+
```
tanmoysrt, line 653: keep a flag on vm doctype ?
And add two functions to enable / disable src/dst check of interfaces

press/press/doctype/server/server.py
```diff
@@ -1784,6 +1806,23 @@ def _parse_semantic_version(version_str: str) -> semantic_version.Version:
...
+				playbook="nat_iptables.yml",
+				server=self,
+				user=self._ssh_user(),
+				port=self._ssh_port(),
+				variables={"nat_gateway_ip": self.nat_gateway_ip},
+			)
+			ansible.run()
+		except Exception:
+			log_error("NAT Iptables Setup Exception", server=self.as_dict())
+
```
tanmoysrt, line 1894: We usually have to provide static ip for some servers on request.

Throw error, in case someone call this nat related  function on a particular server with `is_static_ip` set to `true`

press/press/doctype/server/server.py
```diff
@@ -558,7 +565,7 @@ def create_dns_record(self):
 								"Name": self.name,
 								"Type": "A",
 								"TTL": 3600 if self.doctype == "Proxy Server" else 300,
-								"ResourceRecords": [{"Value": self.ip}],
+								"ResourceRecords": [{"Value": self.ip or self.private_ip}],
```
tanmoysrt, line 569: Maybe better to have some flag on `server` and `vm` to indicate, whether the server is using NAT.

If the flag is set, then pass private ip in dns-ip mapping.

press/press/doctype/server/server.js
```diff
@@ -262,6 +262,12 @@ frappe.ui.form.on('Server', {
 				true,
 				frm.doc.is_server_setup,
 			],
+			[
+				__('Install NAT iptables'),
+				'install_nat_iptables',
+				true,
+				frm.doc.is_server_setup && !!!frm.doc.ip && frm.doc.private_ip,
+			],
```
tanmoysrt, line 277: Expose two buttons ? Enable / Disable NAT on the server.

press/press/doctype/nat_server/nat_server.py
```diff
@@ -0,0 +1,122 @@
...
+			if play.status != "Success":
+				raise
+		except Exception:
+			log_error(
+				"NAT Server Failover Exception",
+				secondary=secondary,
+				primary=self.name,
+				reference_doctype="NAT Server",
+				reference_name=secondary,
+			)
```
tanmoysrt, line 130: Can we use the `Step Runner`  or `Press Job` for it ?

Because
1. there are some blocking external requests and ansible runner. So, this nat_server and vm doctype going to have TimeStampMismatchError due to background sync
2. Have fallbacks for failure. What if you detach the static ip, but it fails during attaching to another server ?
  - In that case, keep trying to attach the ip back to previous NAT
  - If fallback also fails, alert somewhere (In telegram or Raven)

tanmoysrt (changes requested):
We are handling setting up the nat during the server setup, `setup_server` function on `Server` and `Database Server`.

In prod, we usually take the VMI of a completely setup machine and spawn new machine from it and re-configure it.  Also, VMI could have old ip of NAT, so need to make sure that the config stays correct.

There could be 2 way,
1. If you are planning to spawn server with no public ip, then you have to use [cloud_init](https://github.com/frappe/press/blob/846aac145ef49562f7ac8e833568a5c9a8b85116/press/press/doctype/virtual_machine/virtual_machine.py#L697-L701) [cloud_init.yml](https://github.com/frappe/press/blob/develop/press/press/doctype/virtual_machine/cloud-init.yml.jinja2)  to setup the nat thing because without that press can't reach it for setting up the server,

2. Or, we can provision the machine with a public ip, and then switch to NAT,
  - Still, in cloud init, reset the default network routing path, so that machine can be accesible over public internet
  - In Server and Database Server doctype's [`set_additional_config` ](https://github.com/frappe/press/blob/develop/press/press/doctype/server/server.py#L2234) function, add the code to trigger NAT Setup. This is part of [Create Server Press Job](https://github.com/frappe/press/blob/846aac145ef49562f7ac8e833568a5c9a8b85116/press/fixtures/press_job_type.json#L742)
  - Or, you can add a additional step in `Create Server` and `Create Server From Snaoshot` press job

In the second approach, we are going to do the domain mapping to public ip and then it will be switched to private ip. Check if TTL of 5 minute can hurt the server setup.

Because, for new servers, it's expected that the server is ready immediately to deploy benches sites.

press/press/doctype/virtual_machine/virtual_machine.py
phot0n: damn ..cool

press/press/doctype/virtual_machine/virtual_machine.js
phot0n: yes ...so there are 2 diff instance of this provision button ..one if it's an aws machine and one if it's not

i show a dialog box to assign a public ip or not when doing it for aws machines so i've separated both of them out

press/playbooks/roles/nat/tasks/main.yml
tanmoysrt: netplan is better ig, it will persist config across reboots.

press/press/doctype/logical_replication_backup/logical_replication_backup.py
```diff
@@ -329,14 +329,14 @@ def hot_standby_database_server_doc(self) -> "DatabaseServer | None":
...
-	def site_replication_config_dict(self) -> dict:
+	def site_replication_config_dict(self) -> dict:  # type: ignore[return-value]
 		try:
 			return json.loads(self.site_replication_config or "{}")
 		except json.JSONDecodeError:
 			frappe.throw("Invalid site replication config JSON format.")
 
 	@property
-	def bench_replication_config_dict(self) -> dict:
+	def bench_replication_config_dict(self) -> dict:  # type: ignore[return-value]
```
phot0n: remove?

phot0n:
need to keep the ansibleaddhoc api intact

PR #4458, opened 2026-01-09
fix(unified-server): Create corresponding server and database names

Aradhya-Tripathi:


press/press/doctype/virtual_machine/virtual_machine.py
tanmoysrt, line 1565: Throw error , if someone tries to create unified server without using `u` series ?

PR #4499, opened 2026-01-11
feat(server): Add feature flags and default configs for server creation

tanmoysrt:


tanmoysrt:
@mergify backport master

PR #4501, opened 2026-01-11
fix(dashboard): Auto filter available plan types based on filters

tanmoysrt:


tanmoysrt:
@mergify backport master

PR #4506, opened 2026-01-11
fix(bench-deploy-version): Add app release query in CommitChooser

regdocs:
Screenshot 2026-01-11 at 7 10 20 PM

Screenshot 2026-01-11 at 7 08 58 PM

### DIFF
--- a/dashboard/src/components/utils/CommitChooser.vue
+++ b/dashboard/src/components/utils/CommitChooser.vue
@@ -10,12 +10,39 @@
 		</template>
 		<template #body="{ isOpen, togglePopover }">
 			<div
-				v-show="isOpen"
+				v-if="isOpen"
+				@vue:mounted="focusSearch()"
 				class="relative mt-1 rounded-lg bg-surface-modal text-base shadow-2xl"
 			>
-				<div class="max-h-[15rem] overflow-y-auto px-1.5 pb-1.5 pt-1.5">
+				<div
+					class="max-h-[15rem] w-[clamp(0px,50vw,40rem)] overflow-y-auto px-1.5 pb-1.5 pt-1.5"
+				>
+					<FormControl
+						ref="searchInput"
+						class="mt-0.5 mb-1"
+						type="search"
+						placeholder="Search releases..."
+						:modelValue="searchQuery"
+						@update:model-value="
+							(val) => {
+								searchQuery = val;
+								onQueryChange();
+							}
+						"
+					/>
+
 					<div
-						v-for="option in options"
+						v-if="!searchQuery.length"
+						class="text-xs font-medium text-gray-500 py-2 pl-2"
+					>
+						Recent
+					</div>
+
+					<!-- displayedOptions ideally contains query results when searchQuery is non-empty. It is populated with options prop otherwise. -->
+					<!-- In both the above cases, displayedOptions may be empty if there are no releases created. -->
+
+					<div
+						v-for="option in displayedOptions"
 						:key="option.value"
 						@click="selectOption(option, togglePopover)"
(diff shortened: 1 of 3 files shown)


press/api/bench.py
```diff
@@ -1155,3 +1155,24 @@ def confirm_bench_transfer(key: str):
 		http_status_code=403,
 		indicator_color="red",
 	)
+
+
+@frappe.whitelist()
+def search_releases(
```
regdocs: @ssiyad Hey could you please review the security of this?

dashboard/src/components/utils/CommitChooser.vue
```diff
@@ -83,6 +123,56 @@ export default {
...
+	resources: {
+		releases() {
+			return {
+				url: 'press.api.bench.search_releases',
+				params: {
+					app: this.app,
+					source: this.source,
+					query: this.searchQuery?.trim(),
+					fields: ['name', 'message', 'timestamp', 'hash'],
+					limit: 20,
```
ssiyad, line 148: Limit is hardcoded. Pagination or load more is better?

press/api/bench.py
```diff
@@ -1155,3 +1155,24 @@ def confirm_bench_transfer(key: str):
...
+	app: str,
+	source: str,
+	fields: list,
+	query: str | None = None,
+	page_length: int = 20,
```
ssiyad, line 1166: There is `page length` but no `current page`?

dashboard/src/components/utils/CommitChooser.vue
ssiyad, line 148: Endpoint takes `page_length` instead of `limit`

dashboard/src/components/utils/CommitChooser.vue
```diff
@@ -83,6 +123,56 @@ export default {
...
+		},
+	},
+	watch: {
+		searchQuery(newQuery) {
+			if (newQuery.length === 0) {
+				this.queryResult = [];
+				return;
+			}
+		},
 	},
```
ssiyad, line 176: Do we really have to reset results whenever the query is changed? Even if we have to, use `beforeSubmit` in resource?

press/api/bench.py
```diff
@@ -1155,3 +1155,24 @@ def confirm_bench_transfer(key: str):
...
+):
+	DocType = frappe.qb.DocType("App Release")
```
ssiyad, line 1168: Use `AppRelease` as variable name instead of `DocType` since we know that doctype it is.

press/api/bench.py
```diff
@@ -1155,3 +1155,24 @@ def confirm_bench_transfer(key: str):
...
+	q = (
+		frappe.qb.from_(DocType)
+		.select(*fields)
```
ssiyad, line 1171: Maybe sanitize fields? SQLi risk is minimal but we can never be so sure. Its better to return a consistent set of data at all times. So either,

1. Sanitize list of fields or
2. Return a set of pre-defined fields

press/api/bench.py
```diff
@@ -1155,3 +1155,24 @@ def confirm_bench_transfer(key: str):
...
+		.where(DocType.hash.like(f"%{query}%") | DocType.message.like(f"%{query}%"))
+		.where((DocType.public == 1) & (DocType.status == "Approved"))
+		.where((DocType.app == app) & (DocType.source == source))
+		.orderby(DocType.timestamp, order=frappe.qb.desc)
+		.limit(page_length)
+	)
+	return q.run(as_dict=1)
```
ssiyad, line 1178: Return `(query).run(as_dict=True)`. No need to declare `q` since we are not doing anything with it.

PR #4519, opened 2026-01-12
fix(storage): Update subscription on disk size update

tanmoysrt:


tanmoysrt:
@mergify backport master

PR #4526, opened 2026-01-12
fix(site): Show version followed by install compliance app option

Bowrna:
Show version followed by install compliance app option based on app source for supporting v16


shadrak98:
@mergify backport master

PR #4573, opened 2026-01-13
feat(flag): Unified server behind team flag

Aradhya-Tripathi:


press/api/server.py
```diff
@@ -740,6 +740,7 @@ def secondary_server_plans(
 
 @frappe.whitelist()
 def plans(name, cluster=None, platform=None, resource_name=None, cpu_and_memory_only_resize=False):  # noqa C901
+	team = get_current_team(get_doc=True)
```
tanmoysrt: Use `frappe.local.team()`  instead of `get_current_team()` ?

press/api/server.py
```diff
@@ -792,6 +793,7 @@ def plans(name, cluster=None, platform=None, resource_name=None, cpu_and_memory_
 	for plan in plans:
 		if not plan.get("plan_type"):
 			plan["plan_type"] = default_server_plan_type
+		plan["allow_unified_server"] = team.allow_unified_servers and plan.get("allow_unified_server", False)
```
tanmoysrt: Allow unified plans for `System User` without any feature flag


https://github.com/frappe/press/blob/131504e4b1c85f79acb4cfb971762dc85fb6d0a0/press/api/server.py#L591-L601

press/api/server.py
Aradhya-Tripathi: Done

press/api/server.py
Aradhya-Tripathi: Done

press/api/server.py
```diff
@@ -793,6 +793,20 @@ def plans(name, cluster=None, platform=None, resource_name=None, cpu_and_memory_
...
 
+		if (
+			(frappe.session and frappe.session.data and frappe.session.data.user_type)
+			or (
+				frappe.session
+				and frappe.session.user
+				and frappe.get_cached_value("User", frappe.session.user, "user_type")
+			)
+		) == "System User":
+			plan["allow_unified_server"] = True
```
tanmoysrt: It should be whatever mentioned on the plan ? not true always

Aradhya-Tripathi: Yeah fixed, missing commit

PR #4618, opened 2026-01-18
fix(virtual-disk-resize): Don't count overlay stuff multiple times when checking the old filesystem usage

phot0n:


press/infrastructure/doctype/virtual_disk_resize/virtual_disk_resize.py
```diff
@@ -275,14 +275,13 @@ def verify_mount_point(self, device, filesystem):
...
 		"""Reaffirm file system usage using du"""
-		output = self.ansible_run(f"du -s {mountpoint}")["output"]
+		output = self.ansible_run(f"du -sx {mountpoint}")["output"]
 
 		if not output:
 			frappe.throw("Error occurred while fetching filesystem size")
 
 		size = float(output.split()[0])
-		size *= 512  # du measures size in units of 512-byte blocks
-		return size / 1024**3
```
phot0n, line 285: image

phot0n: i dont think we set POSIXLY_CORRECT anywhere (?)

made things explicit anyway using `--block-size`

press/playbooks/roles/copy/tasks/main.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: Copy Files from Source to Destination
-  command: "cp --archive --recursive {{ source }}/. {{ destination }}/"
+  command: "cp --archive --recursive {{ extra_options | default('') }} {{ source }}/. {{ destination }}/"
```
phot0n, line 3: this should preserve hard links and so it shouldn't really do multiple copies but not sure - need to test

press/infrastructure/doctype/virtual_disk_resize/virtual_disk_resize.py
```diff
@@ -453,6 +452,7 @@ def start_copy(self) -> StepStatus:
 		server.copy_files(
 			source=self.filesystem_mount_point,
 			destination=self.new_filesystem_temporary_mount_point,
+			# extra_options="-x",
```
phot0n, line 455: wont make a diff i think as /var/lib/docker is on /opt/volumes/benches ?

phot0n:
merging on staging for now

PR #4635, opened 2026-01-19
fix(do): Check against public key

Aradhya-Tripathi:


Aradhya-Tripathi:
@mergifyIO backport master

PR #4693, opened 2026-01-22
fix(github): Add ability to validate private branches

regdocs:


press/api/github.py
```diff
@@ -260,6 +261,36 @@ def branches(owner, name, installation=None):
 	frappe.throw("Error fetching branch list from GitHub: " + response.text)
 	return None
 
+@frappe.whitelist()
+@protected("Release Group")
+def validate_branch(bench, app, branch):
+	"""Validates whether a branch is available for the `app`"""
+	rg = frappe.get_doc("Release Group", bench)
```
ssiyad: Avoid short forms for variables if possible. `release_group` here is far better than `rg`, which conveys no meaning.

press/api/github.py
```diff
@@ -260,6 +261,36 @@ def branches(owner, name, installation=None):
...
+	app_source = rg.get_app_source(app)
+
+	installation_id = app_source.github_installation_id
+	repo_owner = app_source.repository_owner
+	repo_name = app_source.repository
```
ssiyad: Avoid re-declaring variables if possible. In this case, these variables are used only once can be referenced directly.

press/api/github.py
```diff
@@ -260,6 +261,36 @@ def branches(owner, name, installation=None):
...
+	response = requests.get(
+		f"https://api.github.com/repos/{repo_owner}/{repo_name}/branches/{branch}",
+		headers=headers,
+		timeout=10,
+	)
+
+	if response.ok:
+		return response.json()
+	frappe.throw("Error validating branch from GitHub: " + response.text)
+	return None
```
ssiyad: Do we need both `throw` and `return`? Only one should be enough.

regdocs: ah tried it but ruff was complaining

PR #4924, opened 2026-02-04
fix(dashboard): consider scheduled dashboard banners

phot0n:
* also remove queries in for loop (perf)


phot0n:
unrelated tests failing

PR #4938, opened 2026-02-05
feat(marketplace-version-select): Add custom branch support

regdocs:


regdocs:
merged via https://github.com/frappe/press/pull/4896

Issue #4394, opened 2026-01-06
bug: TLS Renewal Skipped for Standalone Server


tanmoysrt:
In case of standalone server, TLS renewal isn't happenning for wildcard hosts.

PR #5028, opened 2026-02-09
fix(tls-certificate): setup wildcard hosts in standalone servers after tls renewal

phot0n:
fixes: https://github.com/frappe/press/issues/4394


press/press/doctype/tls_certificate/tls_certificate.py
```diff
@@ -228,6 +229,30 @@ def trigger_self_hosted_server_callback(self):
...
+					"status": ("!=", "Archived"),
+					"domain": self.domain,
+					"server": ("in", standalone_servers),
+				},
+				distinct=True,
+				pluck="server",
+			)
+
+			for server in servers:
+				frappe.get_doc("Server", server).setup_wildcard_hosts()
```
phot0n, line 254: maybe this can be improved to only update the cert of the domain that has been updated - not all of them

Issue #1202, opened 2023-11-29
Memory limit additional containers on proxy


balamurali27:
Child table for process name with limits seem fine

PR #5046, opened 2026-02-11
feat(proxy-server): allow setting memory limits for certain processes

phot0n:
fixes: https://github.com/frappe/press/issues/1202


press/press/doctype/proxy_server/proxy_server.py
```diff
@@ -515,6 +518,51 @@ def pre_failover_tasks(self):
 
 		return "Added a dashboard banner and queued reduction of dns record ttl on sites"
 
+	@frappe.whitelist()
+	def get_memory_limits(self):
+		if self.mem_limits:
+			return json.loads(self.mem_limits)
+		return None
```
phot0n, line 525: can be removed

PR #5085, opened 2026-02-14
chore: Improved error messages in client.py

AryamanSharma14:


press/api/client.py
```diff
@@ -124,7 +124,10 @@ def get_list(
 
 	meta = frappe.get_meta(doctype)
 	if meta.istable and not (filters.get("parenttype") and filters.get("parent")):
-		frappe.throw("parenttype and parent are required to get child records")
+		frappe.throw(
+			"parenttype and parent are required to get child records. "
```
mngshm, line 128: You have not used capitalized but on line #259 you've made specific words capitalized. Stick to one ?

PR #5124, opened 2026-02-17
fix(signup): Validate the prefilled subdomain name to avoid issue

Bowrna:
This will validate the prefilled subdomain name. In order to avoid making multiple db calls, we are validating a list of few suggestions in one-go and picking up the best from it.

### DIFF
--- a/press/saas/doctype/product_trial/product_trial.py
+++ b/press/saas/doctype/product_trial/product_trial.py
@@ -434,12 +434,32 @@ class ProductTrial(Document):
 		if email and "@" in email:
 			domain = email.split("@")[1].lower()
 			if domain not in free_email_providers:
-				suggested_subdomain = domain.split(".")[0]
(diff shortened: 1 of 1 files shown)


press/saas/doctype/product_trial/product_trial.py
```diff
@@ -434,12 +434,32 @@ def get_prefilled_subdomain(self, account_request: str | None = None) -> str:
...
+					"subdomain": ("in", candidates),
+					"domain": self.domain,
+					"status": ("!=", "Archived"),
+				},
+				pluck="subdomain",
+			)
+		)
+		for candidate in candidates:
+			if candidate not in existing:
+				return candidate
```
tanmoysrt, line 462: Also, you might need to check the `Site Domain` doctype.
For example. saas site has two domains (one random and another user provided). So, the user provided domain will be there in `Site Domain` doctype.

PR #5130, opened 2026-02-17
fix(login): User accepting invite should not get permission error

Bowrna:
Validate the permission error that user gets when invited to another team

### DIFF
--- a/press/api/account.py
+++ b/press/api/account.py
@@ -268,16 +268,25 @@ def accept_team_invite(key: str):
 	if not account_request.invited_by:
 		frappe.throw("You are not invited by any team")
 
-	team = account_request.team
-	first_name = account_request.first_name
-	last_name = account_request.last_name
-	email = account_request.email
-	password = None
-	role = account_request.role
-	press_roles = account_request.press_roles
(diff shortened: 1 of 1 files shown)


press/api/account.py
```diff
@@ -270,16 +268,25 @@ def accept_team_invite(key: str):
...
-	role = account_request.role
-	press_roles = account_request.press_roles
+	if frappe.session.user != account_request.email:
+		frappe.throw("This invite is not for your account")
 
-	team_doc = frappe.get_doc("Team", team)
-	return team_doc.create_user_for_member(first_name, last_name, email, password, role, press_roles)
+	current_user = frappe.session.user
+	try:
+		frappe.set_user("Administrator")
```
BreadGenie, line 276: What scenario are we fixing here? 
This might be dangerous.

Bowrna: @BreadGenie Doing this because one of the user was facing permission error, when they want to join using invitation. The invitee user already is part of different team and they don't have permission to read the team doc and are not able to accept invite. So followed the same steps of elevating permission like how we do in setup account. Let me know if this is ok or i will see a different approach to handle this case.

Issue #4965, opened 2026-02-06
Handle server resize on unified correctly


balamurali27:
### Information about bug

Mariadb memory limits aren't updated during resize

### Steps to Reproduce

resize unified server

### Relevant log output / Stack trace / Full Error Message.

```shell

```

PR #5132, opened 2026-02-17
fix(unified): Run additional config after resize server

Aradhya-Tripathi:
fixes: #4965

image


Aradhya-Tripathi:
> > @tanmoysrt do we need to do anything in case of disk upgrades as well in the fixture?
> 
> For AWS, check if we are trigerring `Extend EC2 ....` Press Job.

Will make another pr

PR #5163, opened 2026-02-18
feat(billing): Create mandates for UPI to autopay

Bowrna:


dashboard/src/components/billing/UPIAutopayForm.vue
```diff
@@ -0,0 +1,184 @@
...
+		loadRazorpayScript() {
+			if (window.Razorpay) return;
+			const script = document.createElement('script');
+			script.src = 'https://checkout.razorpay.com/v1/checkout.js';
+			script.async = true;
+			document.head.appendChild(script);
+		},
+
+		setupAutopay() {
+			if (!this.maxAmount || this.maxAmount < 500) {
```
shadrak98, line 120: Can we also validate for `this.maxAmount > 100000` ?

dashboard/src/pages/BillingUPIAutopay.vue
```diff
@@ -0,0 +1,268 @@
...
+						format(value) {
+							return value || '-';
+						},
+					},
+					{
+						label: 'Status',
+						fieldname: 'status',
+						type: 'Component',
+						component({ row }) {
+							const statusColors = {
```
shadrak98: badge colours are fetched from `badge.vue` file. Can we add these there?

dashboard/src/pages/BillingUPIAutopay.vue
```diff
@@ -0,0 +1,268 @@
...
+						label: 'Max Amount',
+						fieldname: 'max_amount',
+						format(value) {
+							return `₹${value?.toLocaleString('en-IN') || 0}`;
+						},
+					},
+					{
+						label: 'Expires On',
+						fieldname: 'expires_on',
+						type: 'Date',
```
shadrak98, line 141: lets also format date values, like we have in invoice and balance list views
Those are more human readable

dashboard/src/pages/BillingUPIAutopay.vue
```diff
@@ -0,0 +1,268 @@
...
+						},
+					];
+				},
+				orderBy: 'creation desc',
+				primaryAction: ({ listResource }) => {
+					// Hide button if active mandate exists
+					const hasActiveMandate = listResource.data?.some(
+						(mandate) => mandate.status === 'Active',
+					);
+					if (hasActiveMandate) {
```
shadrak98: Instead of checking for mandate and returning null, we can just check `(!hasActiveMandate)` and then show the primary action.

press/api/billing.py
```diff
@@ -674,6 +674,199 @@ def is_paypal_enabled() -> bool:
...
+				_(
+					"A mandate authorization is already in progress. Please complete or cancel it before creating a new one."
+				)
+			)
+
+	if auth_type not in ("upi", "netbanking"):
+		frappe.throw(_("Invalid auth type. Must be 'upi' or 'netbanking'"))
+
+	if max_amount < 500:
+		frappe.throw(_("Maximum amount must be at least ₹500"))
```
shadrak98: the error message should be "Minimum amount for UPI must be atleast 500"

Also can we add validation for **Maximum** amount i.e 1 lac

press/api/billing.py
```diff
@@ -674,6 +674,199 @@ def is_paypal_enabled() -> bool:
...
+	# Verify signature
+	try:
+		client.utility.verify_payment_signature(
+			{
+				"razorpay_payment_id": razorpay_payment_id,
+				"razorpay_order_id": razorpay_order_id,
+				"razorpay_signature": razorpay_signature,
+			}
+		)
+	except Exception:
```
shadrak98, line 815: Is throwing in the exception necessary? the except block becomes redundant then. Maybe we can log it in error log?

press/press/doctype/invoice/invoice.py
```diff
@@ -480,6 +490,136 @@ def get_mandate_id(self, customer_id):
 			return ""
 		return mandate_id
 
+	def create_razorpay_payment(self):
+		"""Create a recurring payment via Razorpay mandate"""
+		if self.razorpay_payment_id:
+			self._check_existing_razorpay_payment()
+			return
+
+		if self.amount_due_with_tax <= 0:
```
shadrak98: This check is not needed since `create_razorpay_payment` will only be called when `amount_due` > 0 ref Line 320

press/press/doctype/invoice/invoice.py
```diff
@@ -1114,7 +1252,7 @@ def finalize_draft_invoices():
 	for name in invoices:
 		invoice = frappe.get_doc("Invoice", name)
 		# don't finalize if invoice ends today and time is before 6 PM
-		if invoice.period_end == today and current_time.hour < 18:
+		if invoice.period_end == today and current_time.hour < 10:
```
shadrak98: lets not change the invoice finalisation time to 10 am in the morning. 
The reason it is at 6 pm because by that time all the usage records for the day has been created.

dashboard/src/components/billing/UPIAutopayForm.vue
Bowrna: @shadrak98 Added validation. Thanks for pointing out.

dashboard/src/components/global/Badge.vue
```diff
@@ -31,18 +31,19 @@ export default {
...
+				Cancelled: 'red',
 				'Update Available': 'blue',
 				Enabled: 'blue',
 				'Awaiting Approval': 'orange',
 				'Awaiting Deploy': 'orange',
 				Success: 'green',
 				Completed: 'green',
 				Deployed: 'green',
-				Expired: 'red',
+				Expired: 'gray',
```
shadrak98, line 46: Let's keep expired as red only. This is used at other places as well.

press/api/billing.py
Bowrna: Yes mandate creation itself is failed and we have to return this to user.

dashboard/src/components/global/Badge.vue
Bowrna: I verified and it was not used in other places @shadrak98 . Thats why i changed it here to gray as it makes more sense. But if you think it was for common use case, i will extend this part and use gray in place for upi autopay part. Let me know what do you think.

PR #5229, opened 2026-02-20
 fix(site): In site archival, release the site name at the end

tanmoysrt:
Example : https://cloud.frappe.io/app/error-log/9k67mb361t


tanmoysrt:
@mergify backport master

PR #5239, opened 2026-02-22
fix(resize): Check for unified server

Aradhya-Tripathi:


Aradhya-Tripathi:
@mergifyIO backport master

PR #5360, opened 2026-02-27
perf(bench): Faster bench creation

Bowrna:
This was one of the slowest query analysed and the index was missing. Fixing it can greatly improve the new bench creation process.


phot0n:
`perf(bench): faster bench creation` - as title?

also maybe some before/after in description? (if time allows)

PR #5426, opened 2026-03-04
feat(investigator): Add pattern detection and internal warnings & recommendations

Aradhya-Tripathi:


Aradhya-Tripathi:
Server 3 has failing test, being addressed in https://github.com/frappe/press/pull/5439

PR #5460, opened 2026-03-05
fix(region-migration): In case a deploy is being created add server to group

Aradhya-Tripathi:


press/press/doctype/release_group/release_group.py
```diff
@@ -1519,41 +1519,42 @@ def add_server(self, server: str, deploy=False, force_new_build: bool = False) -
 		create a deploy check if the image has not been pruned from the registry in case of
 		missing image create new build.
 		"""
-		if deploy:
-			server_platform = frappe.get_value("Server", server, "platform")
-			last_successful_deploy_candidate_build = self.get_last_successful_candidate_build(
-				platform=server_platform
-			)
+		if not deploy:
+			return None
```
tanmoysrt, line 1523: @Aradhya-Tripathi   why this check though ?

We might just want to add the server in list, but dont want to trigger the deploy immediately.

Aradhya-Tripathi: We generally do that via the child table directly.

tanmoysrt: You may keep the code something like this.

image


Else, there is no point of `deploy` argument in the function.

Aradhya-Tripathi: Some how master already had this, due to failed backport(?)
`if not deploy: return` 
Don't want to touch anything else for now, we can always internally `add_server` via desk, using this,

https://github.com/frappe/press/blob/397a2c9c5a9cb4be371d7cb182313ece67d3a517/press/press/doctype/release_group/release_group.js#L85-L89

and when we don't want to create a deploy we anyways just append in the child table, have to move the argument anyways

PR #5599, opened 2026-03-10
refactor(bench): Improving bench queue execution & Metrics

Aradhya-Tripathi:
- On server archival ensure that
  - Release Groups with only the archived server in them are disabled
  - Release Groups with multiple servers including the archived server remove the archived server from their list

- Bench execution queue favours private benches
- Introduce cache based locking for bench queue execution

### DIFF
--- a/press/press/doctype/app_source/test_app_source.py
+++ b/press/press/doctype/app_source/test_app_source.py
@@ -55,13 +55,3 @@ class TestAppSource(FrappeTestCase):
 
 		for req_app in source.required_apps:
 			self.assertEqual("https://github.com/frappe/erpnext", req_app.repository_url)
-
-		app: App = self.create_app("lms", "LMS")
-		source = app.add_source(
-			frappe_version="Nightly",
-			repository_url="https://github.com/frappe/lms",
-			branch="develop",
-			team=team_name,
-		)
-
-		self.assertEqual([], source.required_apps)


press/press/doctype/server/server.py
```diff
@@ -1165,6 +1166,51 @@ def rename_server(self):
...
+		"""Remove application server from all enabled release groups.
+		This is done to allow successful deploys on the release group post server archival
+		"""
+		if self.doctype != "Server":
+			return
+
+		ReleaseGroup = frappe.qb.DocType("Release Group")
+		ReleaseGroupServer = frappe.qb.DocType("Release Group Server")
+
+		release_group_filter = (
```
tanmoysrt, line 1191: Add a validation here anyway to ensure there is no active bench on the server.
In case, someone later plan to use it for other purpose.

press/press/doctype/bench/bench.py
```diff
@@ -1681,22 +1682,58 @@ def group_supervisor_processes(processes: list[SupervisorProcess]):
...
+
+	# Initially we fill up the slots with the private benches.
+	private_bench_tasks = query.where(ReleaseGroup.public == 0).limit(slots_available).run(as_dict=True)
+	remaining_slots = slots_available - len(private_bench_tasks)
+	# In case there are still slots available after queuing all private benches, we can queue public benches as well.
+	public_benches_tasks = (
+		query.where(ReleaseGroup.public == 1).limit(remaining_slots).run(as_dict=True)
+		if remaining_slots > 0
+		else []
+	)
```
phot0n, line 1708: won't this lead to starvation for public benches?

Aradhya-Tripathi: Maybe 🤡 
We rarely deploy them, and it's fine if they just wait in queue for sometime anyways, also the queue size is large enough to account for this, we generally have 15-20 new bench jobs running concurrently for private benches.

press/press/doctype/server/server.py
Aradhya-Tripathi: Done: 94c6358cd7495ecb096b59155bbe0bd8bc90f820

PR #5674, opened 2026-03-13
feat(status-page): Add status page

Aradhya-Tripathi:
## Automatically create banner in case of incident. Ideally should be team level.
image

## Redirects to status page
image

image

image


Aradhya-Tripathi:
Unrelated tests seem to be failing.

PR #5713, opened 2026-03-15
feat(cluster): Support for enable/disable public ip provisioning for new servers

phot0n:


phot0n:
love the faliures :heart:

press/api/server.py
```diff
@@ -247,6 +248,7 @@ def new(server):
 		frappe.throw("You cannot create a new server because your account is disabled")
 
 	cluster: Cluster = frappe.get_doc("Cluster", server["cluster"])
+	_add_nat_server_if_required(cluster)
```
tanmoysrt: Move this to the method of cluster ?

https://github.com/phot0n/press/blob/26b62b3ddb09a28daba3565e58563fd3c3ecc728/press/press/doctype/cluster/cluster.py#L1294-L1302

press/api/server.py
```diff
@@ -226,6 +226,7 @@ def new_unified(server: UnifiedServerDetails):
 		limit=1,
 	)[0]
 
+	_add_nat_server_if_required(cluster)
 	cluster.proxy_server = proxy_server.get("name")
```
tanmoysrt, line 229: Move this call to the method in cluster doctype ?

https://github.com/phot0n/press/blob/26b62b3ddb09a28daba3565e58563fd3c3ecc728/press/press/doctype/cluster/cluster.py#L1238-L1242

press/fixtures/press_job_type.json
```diff
@@ -738,229 +534,150 @@
...
-  "modified": "2026-02-27 11:03:50.212370",
+  "modified": "2026-03-15 21:28:00.551361",
   "name": "Create Server",
   "steps": [
    {
-    "parent": "Create Server",
-    "parentfield": "steps",
-    "parenttype": "Press Job Type",
-    "script": "machine = frappe.get_doc(\"Virtual Machine\", doc.virtual_machine)\nmachine.provision()\n",
+    "script": "machine = frappe.get_doc(\"Virtual Machine\", doc.virtual_machine)\nassign_public_ip = (\n    frappe.db.get_value(\"Cluster\", machine.cluster, \"disable_public_ips_for_servers\") \n    if machine.series in (\"f\", \"m\") \n    else True\n)\nmachine.provision(assign_public_ip)\n",
```
tanmoysrt: Maybe move this logic to VM's provision method directly.
We can have some flag on either server or vm, which can be set at the time of server creation.

press/fixtures/press_job_type.json
```diff
@@ -738,229 +534,150 @@
...
     "wait_until_true": 1
    },
    {
-    "parent": "Create Server",
-    "parentfield": "steps",
-    "parenttype": "Press Job Type",
+    "script": "if doc.server_type in (\"Server\", \"Database Server\"):\n    server = frappe.db.get_value(doc.servert_type, doc.server, (\"public_ip_address\", \"nat_server\"), as_dict=True)\n    if not server.public_ip_address and server.nat_server:\n        doc.install_nat_iptables()\n",
+    "step_name": "Install NAT Iptables",
+    "wait_until_true": 0
+   },
```
tanmoysrt: In next step, wait for installation to complete first.

Example : https://github.com/phot0n/press/blob/26b62b3ddb09a28daba3565e58563fd3c3ecc728/press/fixtures/press_job_type.json#L160-L164

tanmoysrt: Secondly, maybe better to put this in cloud init (if possible)

https://github.com/phot0n/press/blob/26b62b3ddb09a28daba3565e58563fd3c3ecc728/press/press/doctype/virtual_machine/virtual_machine.py#L729-L752

https://github.com/phot0n/press/blob/26b62b3ddb09a28daba3565e58563fd3c3ecc728/press/press/doctype/virtual_machine/cloud-init.yml.jinja2

press/api/server.py
phot0n: added it here coz we figure out the proxy server here as well ..but sure can be moved

press/fixtures/press_job_type.json
phot0n: will keep this separate for now ...want to be able to monitor this separately

and make this intentional as well

press/api/server.py
phot0n: added it here coz we figure out the proxy server here as well ..but sure can be moved

press/fixtures/press_job_type.json
phot0n: there is a dialog box that pops up when we provision the machine (in desk) 

guess i'll have to change that

press/fixtures/press_job_type.json
phot0n: i guess we'll have to put this in cloud init due to 

https://github.com/frappe/press/blob/develop/press/press/doctype/virtual_machine/cloud-init.yml.jinja2#L18

PR #5897, opened 2026-03-25
Feat dark theme

siduck:
image

PR is still  draft cuz there are many colors we use which frappe-ui doesnt have  cuz Timeless didnt  update it


siduck:
closing this as conflicts are too much

PR #5912, opened 2026-03-25
feat(bench-cleanup): Respond and react to zombie bench complains

Aradhya-Tripathi:
Goes with https://github.com/frappe/agent/pull/479


press/api/bench.py
```diff
@@ -1252,3 +1257,68 @@ def search_releases(
...
+		.select(Bench.name)
+	).run(pluck="name")
+
+	if zombie_benches_with_active_sites:
+		frappe.db.set_value(
+			"Bench",
+			{"name": ("in", zombie_benches_with_active_sites)},
+			"status",
+			"Active",
+		)
```
tanmoysrt: Then, the release group also need to be marked as active and remove the ".archived" prefix.

But, ideally we should check, why it happens at first place. Why archival of bench with active sites become even successful.

press/api/bench.py
```diff
@@ -1252,3 +1257,68 @@ def search_releases(
...
+	if not sent_password or not server or not frappe.db.exists("Server", server):
+		frappe.throw(insufficient_permissions_message, frappe.PermissionError)
+
+	agent_password = frappe.get_doc("Server", server).get_password(
+		"agent_password",
+	)
+	pbkdf2.verify(agent_password, sent_password) or frappe.throw(
+		insufficient_permissions_message, frappe.PermissionError
+	)
+
```
tanmoysrt: We shouldn't do this kind of validations though. Accepting password hash and doing validation can open up way to brute force and finding agent password.

Maybe we should issue random tokens instead.

press/api/bench.py
Aradhya-Tripathi: Actually it never happens checked via system-console, just wanted to be safe regardless can remove this for now.

press/api/bench.py
Aradhya-Tripathi: abandoning daemon process for now, however we need a better way for back comm from agent.

PR #5944, opened 2026-03-27
feat(deploy): Revamp deploy flow

Aradhya-Tripathi:
- [x] Borrowed and integrated workflow engine from https://github.com/frappe/press-agent-manager
- [x] Consolidated validation build and bench jobs from all over the place.
- [x] Currently residing in a Workflow doctype
- [x] Figure out the retry logic.
- [x] Figure out auto build retry logic
- [x] Figure out auto bench retry logic
- [x] Add support for validation failure notifications
- [x] Remove update available button till the pipeline execution is not in it's terminal state
- [ ] Figure out design and UI for new deploy display

fixes: https://github.com/orgs/frappe/projects/200/views/4?pane=issue&itemId=169406145&issue=frappe%7Cpress%7C5939

### All the deployment flow is now marked in a single doctype, Managed by the Workflow Engine.

image

### Internal steps are shown here as well
image

### While the `Press Workflow` is the execution brain, the state manager for deploy is the `Release Pipeline` doctype

image

Should effectively get rid of long running transactions in the request cycle when a new deploy is created, allowing for instant deploy creations and avoiding random lock-wait timeout issues due to lock contention in long running transactions. Should also show the retry logic with a single source of truth `Release Pipeline` doctype which will **not** reach terminal state until all the retries (Build & Bench) have reached their individual terminal states.

### DIFF
--- a/press/press/doctype/release_pipeline/test_release_pipeline.py
+++ b/press/press/doctype/release_pipeline/test_release_pipeline.py
@@ -1,7 +1,6 @@
 # Copyright (c) 2026, Frappe and Contributors
 # See license.txt
 import shutil
-import typing
 from unittest.mock import Mock, patch
 
 import frappe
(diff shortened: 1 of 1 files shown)


Aradhya-Tripathi:
Remove a test,

`test_deploy_and_update_fn_fails_without_release_argument`

Quite ambiguous in nature, we only throw validation error when it's a in-place update, otherwise we let sites, be empty

https://github.com/frappe/press/blob/e9947a7979cbca937d427f78d69d4bb42a508aac/press/press/doctype/bench_update/bench_update.py#L48-L53

Secondary this test was only passing since we also throw, the same validation error in;

https://github.com/frappe/press/blob/e9947a7979cbca937d427f78d69d4bb42a508aac/press/api/bench.py#L770

Issue #5776, opened 2026-03-18
Show correct country code in RazorPay modal


adityahase:
### Information about bug

For teams outside India, we show +91 country code while adding prepaid credits via PayPal.

Customers can select the correct code from the dropdown.

We should show the correct code since we know the country.

<img width="1552" height="987" alt="Image" src="https://github.com/user-attachments/assets/0f89f172-122c-4cc9-b3c5-13d31f7b82d6" />

### Steps to Reproduce

Non-INR team -> Add Prepaid Credits > PayPal

### Relevant log output / Stack trace / Full Error Message.

```shell

```

PR #5984, opened 2026-03-30
fix(billing): Correct country code prefilled in razorpay checkout

Bowrna:
partly fixes: #5776 

Edit: As phone number is optional, we will prefill contact only for users who has given contact number during signup and not otherwise


dashboard/src/components/billing/BuyCreditsRazorpay.vue
```diff
@@ -117,13 +117,20 @@ const handlePaymentFailed = createResource({
...
 function processOrder(data) {
 	const options = {
 		key: data.key_id,
 		order_id: data.order_id,
 		name: 'Frappe Cloud',
 		image: 'https://frappe.io/files/cloud.png',
-		prefill: { email: team.doc?.user },
+		prefill: {
+			email: team.doc?.user,
+			...(getRazorpayContact() && { contact: getRazorpayContact() }),
```
ssiyad, line 132: `contact: team.doc?.phone_number` will suffice, right?

Bowrna: in case of undefined, the contact param should not be passed.

ssiyad: If `team.doc?.phone_number` is `undefined`, this works as expected. If it is any other falsy value, `contact: team.doc?.phone_number || undefined` should work?

Bowrna: It can be undefined or empty, that's why I added the spread operator to include contact param if it has defined value and not pass otherwise. let me know if it can be improved. i will include the changes and push it

Bowrna: @ssiyad let me know if this comment can be resolved or do you have more feedback on this part?

PR #5986, opened 2026-03-30
refactor(billing): Modify the billing flow on welcome

Bowrna:
Billing setup in welcome page should redirect to billing tab.


dashboard/src/components/billing/PaymentDetails.vue
```diff
@@ -38,6 +38,31 @@
 				v-if="team.doc.payment_mode == 'Card'"
 				class="my-3 h-px bg-gray-100"
 			/>
```
Bowrna, line 40: Reordered the Billing address to come first in the billing page, before setting up the payment mode

PR #6056, opened 2026-04-09
feat(marketplace-auditor): Improve marketplace auditor

prathameshkurunkar7:
Improves the Marketplace App Auditor by adding stronger compatibility and correctness checks, expanding Semgrep security/correctness rules, and making metadata link validation more reliable with request headers/timeouts. This helps catch more real-world app issues earlier and reduces false negatives during audit.

The security and correctness checks are all taken from `semgrep-rules` repo and added a few more. The relevant ones are added which identify several issues such as Monkey patching, SQL Injection etc. More checks can be added later as well as per requirement, the architecture is fairly extensible. 

The Auditor has the logic in place where it parses the semgrep findings into readable text for the developers as this can later directly be used for communication.

Apart from security checks, also introduced the compatibility check which was discussed with @Aradhya-Tripathi  - 
Apps installed on v15 and v16 benches later change the dependency and make it only v16 compatible which leads to broken v15 benches due to incompatible apps. The Auditor manages to catch such cases. 

Currently the Auditor is not taking any strict action, will introduce that in a separate PR with few more modifications.

### Screenshot:

Automatically creates such check results:

CleanShot 2026-04-09 at 12 34 11@2x


Aradhya-Tripathi:
Does this approve/reject app sources yet?

prathameshkurunkar7:
> Does this approve/reject app sources yet?
> 
> 

Not yet, will do that in a separate PR.

PR #6085, opened 2026-04-11
fix(api): Tighten site backup link validation

barredterra:
- align `get_backup_link` with the existing site-level backup download behavior
- tighten validation around backup link resolution and accepted file types
- add regression coverage for mismatched backup/site lookups and invalid file selections


press/api/site.py
```diff
@@ -664,8 +664,15 @@ def backups(name):
 @frappe.whitelist()
 @protected("Site")
 def get_backup_link(name, backup, file):
+	if file not in ["database", "public", "private", "config"]:
+		frappe.throw("Invalid file type")  # nosemgrep
```
ssiyad, line 668: Please wrap in `_` for user facing messages.

ssiyad (changes requested):

press/api/site.py
barredterra: This is not user-facing, usually. Also not translating messages is an established pattern in this file that would need separate comprehensive fixing, out of scope for this PR.

ssiyad (changes requested):
Tests are failing. Please fix.

PR #6096, opened 2026-04-13
feat(compute): Increase disk size

the-bokya:


press/press/doctype/server/server.py
tanmoysrt, line 1141: Add `Frappe Compute` related handling here to extend disk automatically.

the-bokya: Ohkay :+1:

press/press/doctype/server/server.py
```diff
@@ -1052,6 +1052,42 @@ def enqueue_extend_ec2_volume(self, device, log):
...
+			ansible = Ansible(
+				playbook="extend_frappe_compute_volume.yml",
+				server=self,
+				user=self._ssh_user(),
+				port=self._ssh_port(),
+				variables={"restart_mariadb": restart_mariadb, "device": device},
+			)
+			play = ansible.run()
+			if log:
+				frappe.db.set_value("Add On Storage Log", log, "extend_frappe_compute_play", play.name)
```
tanmoysrt: This is a field on "Add on Storage Log" doctype. You have to create the field.

press/press/doctype/server/server.py
```diff
@@ -1096,6 +1132,9 @@ def increase_disk_size(self, increment=50, mountpoint=None, log: str | None = No
 			# Non-boot volumes might not need resize
 			self.break_glass()
 			self.reboot()
+		elif self.provider == "Frappe Compute":
+			device = "/dev/vda"
```
tanmoysrt: Better not to hardcode this.

If `guess_data_disk_mountpoint` function return correct volume row, you can use the device if from that `volume` variable.

press/press/doctype/server/server.py
the-bokya: Will handle later in a separate PR.

PR #6108, opened 2026-04-14
fix(site): Dont add subscription if site creation is broken

### LINKED ISSUE #6110: Don't create Subscription if Site creation fails
### Information about bug

Currently even if the site creation fails, the subscription record is created and user is being billed for unusable / broken site

### Steps to Reproduce

-

### Relevant log output / Stack trace / Full Error Message.

```shell

```

Bowrna:
fixes: #6110
If the site is broken during creation we should not enable subscription for that case and charge customers.

Reference ticket id: 65320


Bowrna:
@shadrak98 let me know if you think this will work out for the case where site creation is broken.

PR #6114, opened 2026-04-15
fix(version-upgrade): Correct date time comparison in Version Upgrade dialog

AyshaHakeem:
- In case where bench creaiton is required for upgrading site, scheduled time for upgrade has to be at least 30 minites from now
- However, string input emitted by DateTimePicker was interpreted as UTC instead of user's local timezone, hence the above comparison fails

### Before:
Before

### After:
after

### DIFF
--- a/dashboard/src/components/site/SiteVersionUpgradeDialog.vue
+++ b/dashboard/src/components/site/SiteVersionUpgradeDialog.vue
@@ -312,6 +312,11 @@ import { getCachedDocumentResource, LoadingIndicator } from 'frappe-ui';
 import { toast } from 'vue-sonner';
 import AlertBanner from '../AlertBanner.vue';
 import DateTimePicker from 'frappe-ui/src/components/DatePicker/DateTimePicker.vue';
+import dayjs from '../../utils/dayjs';
+import customParseFormat from 'dayjs/plugin/customParseFormat';
+
+const PICKER_DATETIME_FORMAT = 'YYYY-MM-DD h:mm a';
+const IST_TIMEZONE = 'Asia/Calcutta';
 
 export default {
 	name: 'SiteVersionUpgradeDialog',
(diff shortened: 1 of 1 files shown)


dashboard/src/components/site/SiteVersionUpgradeDialog.vue
```diff
@@ -312,6 +312,11 @@ import { getCachedDocumentResource, LoadingIndicator } from 'frappe-ui';
 import { toast } from 'vue-sonner';
 import AlertBanner from '../AlertBanner.vue';
 import DateTimePicker from 'frappe-ui/src/components/DatePicker/DateTimePicker.vue';
+import dayjs from '../../utils/dayjs';
+import customParseFormat from 'dayjs/plugin/customParseFormat';
+
+const PICKER_DATETIME_FORMAT = 'YYYY-MM-DD h:mm a';
+const IST_TIMEZONE = 'Asia/Calcutta';
```
BreadGenie, line 318: could this cause issues for a user that's not in IST timezone?

AyshaHakeem: This is only for converting time to IST before setting the `scheduled_time` field of Version Upgrade record
- https://github.com/AyshaHakeem/press/blob/f58e278ce3afe7557785e991bb37e04e4a08bc30/dashboard/src/components/site/SiteVersionUpgradeDialog.vue#L382C3-L382C16
- https://github.com/AyshaHakeem/press/blob/f58e278ce3afe7557785e991bb37e04e4a08bc30/dashboard/src/components/site/SiteVersionUpgradeDialog.vue#L442

BreadGenie: you can probably mention why in the code

BreadGenie (approved):

PR #6120, opened 2026-04-15
fix(marketplace-add-version): Handle more than one app source

regdocs:
When fetching branches from app sources for a marketplace app version, we're picking the top 1 result which may not always be the intended branch. Fixed this by getting all branches from all app sources created by the team in the past for that app.

ref: https://support.frappe.io/helpdesk/tickets/65418


press/api/marketplace.py
```diff
@@ -1242,19 +1242,23 @@ def options_for_version(name):
...
 	versions = list(set(frappe_version).difference(set(added_versions)))
-	branches_list = [branch["name"] for branch in branches_list]
 
 	return [{"version": version, "branch": branches_list} for version in versions]
 
 
 @protected("Marketplace App")
 @frappe.whitelist()
-def add_version(name, branch, version):
+def add_version(name, repo_owner, repo_name, branch, version):
```
prathameshkurunkar7, line 1259: @regdocs  can we add type hints to all whitelisted methods? Add it to this one and `add_version()` in `marketplace_app`

prathameshkurunkar7 (changes requested):

prathameshkurunkar7 (approved):

PR #6121, opened 2026-04-15
fix(notifications): unreadcount & open support access modal instead of navigation

siduck:
- open support access modal if the notification tile.type == 'Support Access' instead of going to that link 
- Unread dot indicator for notif icon in sidebar

image

<hr/>

- Open notification panel's request section on search item click

https://github.com/user-attachments/assets/9a928009-22aa-4725-b542-09a437934d06

### DIFF
--- a/dashboard/src/App.vue
+++ b/dashboard/src/App.vue
@@ -31,7 +31,10 @@
 				</div>
 			</div>
 		</div>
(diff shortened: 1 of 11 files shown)


siduck:
@mergifyio backport master

PR #6122, opened 2026-04-15
fix(site): Check unified server with private ip

balamurali27:
Since we don't set public ip anymore


phot0n:
might not be needed https://github.com/frappe/press/pull/6126

PR #6139, opened 2026-04-16
fix: Archival of suspended sites

AyshaHakeem:
- Add `suspended_at` field to Site doctype to track site suspension time
- Add Patch to populate `suspended_at` field in Site doctype based on 'Suspend Site' activity record's creation
- Set and unset `suspended_at` field on site suspension and activation respectively
- Update archive_suspended_sites job
    - Filter based on the newly added suspended_at field value
    - Move out logic for notifying the user before archival into a separate job
- Add test


AyshaHakeem:
@mergify backport master

PR #6143, opened 2026-04-16
fix(notifications): Decrement unread count only on unread row clicks only

siduck:


siduck:
@mergifyio backport master

PR #6161, opened 2026-04-17
feat(marketplace-auditor): Wire up the auditor to take actions

prathameshkurunkar7:
### What this changes

**Removes Code Screening entirely.** The old static-analysis screener is gone — dashboard components (`CodeReview.vue`, `NewComment.vue`), API endpoints, the `App Release Approval Code Comments` child DocType, screening fields on `App Release Approval Request`, and the `review_stage` field on `Marketplace App`. The Auditor now owns that responsibility.

**Makes the Auditor an active decision-maker.** On audit completion with a `Fail` result (and when `enable_audit_actions` is on in Marketplace Settings - feature flag), it now:
- Yanks the release via the existing Yank mechanism — which blocks deployment and marks the release invalid, without introducing a new status.
- Auto-rejects any open approval requests for that same release with the audit summary as the rejection reason.

**Prevents approval requests for already-yanked releases.** New guard in `AppReleaseApprovalRequest.before_insert` avoids resurrecting bad releases into the review queue.

**Adds a "Send Report to Publisher" action.** Visible on the audit Desk form when result is `Fail` / `Warn` / `Needs Improvement`. Sends a structured Jinja email listing each failing check with severity, category, semgrep occurrences (file, line, rule-id, message), and remediation guidance. A CC option lets reviewers keep a copy(behind feature flag).

**Simplifies status ownership.** Status now lives in one place per concept — `Marketplace App.status` for listing lifecycle, `App Release.status` for release state, `Marketplace App Audit.audit_result` for audit outcome. No more duplicated/overlapping fields.

### Data migration

Added `cleanup_code_screening_data` patch that nulls legacy screening fields and clears orphaned code-comment rows before `bench trim-tables` drops the unused columns/tables.


press/marketplace/doctype/marketplace_app_audit/marketplace_app_audit.py
```diff
@@ -266,3 +268,101 @@ def _build_summary(self, tally: dict) -> str:
...
 		return "\n".join(lines)
+
+	def _handle_audit_outcome(self):
+		"""
+		This method will ensure that some action is taken on the audit outcome.
+		"""
+		# first check if enable audit action is enabled in Marketplace settings
+		if not frappe.db.get_single_value("Marketplace Settings", "enable_audit_actions"):
+			return
+
```
Aradhya-Tripathi, line 279: Maybe mark the app as "Attention Required" as well since that would ensure we have the last releases deploying correctly?

press/marketplace/doctype/marketplace_app_audit/marketplace_app_audit.py
```diff
@@ -266,3 +268,101 @@ def _build_summary(self, tally: dict) -> str:
...
+			)
+			request.save()
+
+	@frappe.whitelist()
+	def send_report_to_publisher(self):
+		"""Email the app publisher a structured audit report with all failing checks."""
+		marketplace_app_team, marketplace_app_title = frappe.db.get_value(
+			"Marketplace App", self.marketplace_app, ["team", "title"]
+		)
+		publisher_email = frappe.db.get_value("Team", marketplace_app_team, "user")
```
Aradhya-Tripathi, line 329: Maybe auto send report if the app is marked as "Attention Required"

prathameshkurunkar7: Will add this later, need to first monitor whether the template being sent is correct or not. Will keep track for a week and then add this.

PR #6178, opened 2026-04-19
fix(ui):  padding issues and improve banner

siduck:
https://drive.frappe.io/writer/w/1g854gspja


siduck:
@mergifyio backport master

PR #6194, opened 2026-04-20
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `High` | **File**: `dashboard/src/components/CustomToast.vue:L22`

The component renders the `html` prop using `v-html` without sanitization. If this prop can contain user-controlled or server-returned content, attackers can inject arbitrary HTML/JavaScript, leading to account takeover or data exfiltration in the dashboard context.

## Solution

Avoid `v-html` for untrusted data. Render as plain text (`{{ html }}`) or sanitize strictly with a vetted sanitizer (e.g., DOMPurify) before rendering.

## Changes

- `dashboard/src/components/CustomToast.vue` (modified)
- `dashboard/src/components/AlertBanner.vue` (modified)

### DIFF
--- a/dashboard/src/components/AlertBanner.vue
+++ b/dashboard/src/components/AlertBanner.vue
@@ -11,7 +11,9 @@
 				v-if="showIcon && type === 'info'"
 				:class="`ml-1 size-4 text-${color}-600 shrink-0`"
 			/>
-			<div class="prose-sm font-medium text-gray-800" v-html="title" />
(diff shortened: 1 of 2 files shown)


ssiyad (changes requested):
This may break existing HTML based content rendering. Can you please update this to use DOMPurify?

PR #6213, opened 2026-04-22
fix(release-pipeline): Get latest retried build with the same platform

Aradhya-Tripathi:
- Get the latest required build of the same platform.
(Might require composite indexing)
- Explicitly use the `build` queue in production.
- More permissions to `Press Workflow`

### DIFF
--- a/press/press/doctype/release_pipeline/release_pipeline.py
+++ b/press/press/doctype/release_pipeline/release_pipeline.py
@@ -168,7 +168,12 @@ class ReleasePipeline(WorkflowBuilder):
 			"Press Workflow", {"linked_doctype": "Release Pipeline", "linked_docname": self.name}, "name"
 		)
 
-	@task
+	@staticmethod
+	def _get_task_execution_queue() -> str:
+		"""Determine the appropriate queue for task execution based on the release group."""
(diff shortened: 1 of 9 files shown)


press/press/doctype/release_pipeline/release_pipeline.py
```diff
@@ -164,17 +163,11 @@ def release_group_doc(self) -> "ReleaseGroup":
 		return frappe.get_doc("Release Group", self.release_group)
 
 	@cached_property
-	def workflow_name(self) -> str:
+	def current_workflow(self) -> str:
 		return frappe.db.get_value(
 			"Press Workflow", {"linked_doctype": "Release Pipeline", "linked_docname": self.name}, "name"
 		)
```
tanmoysrt, line 169: If you are inside the workflow, it will have `self.workflow_name` populated by `Workflow Builder`

Aradhya-Tripathi: I tried that last time got a None for `workflow_name` also typing says it might be `None`

https://github.com/frappe/press/blob/4d40d810b2a3bde9b8c26bf04bdc14bdd9578ede/press/workflow_engine/doctype/press_workflow/workflow_builder.py#L47

PR #6242, opened 2026-04-24
refactor(digital-ocean): Use tags to manage firewall and droplets

Aradhya-Tripathi:
Ensure we use tags to group and add multiple droplets to a firewall since, digital ocean caps number of droplets attached to a firewall to 10.

### DIFF
--- a/press/press/doctype/virtual_machine/virtual_machine.py
+++ b/press/press/doctype/virtual_machine/virtual_machine.py
@@ -489,7 +489,7 @@ class VirtualMachine(Document):
 
 		cluster: Cluster = frappe.get_doc("Cluster", self.cluster)
 
-		droplet_tag = f"Frappe-Cloud-Production-{cluster.name.replace(' ', '-')}"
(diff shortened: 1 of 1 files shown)


press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -515,7 +517,7 @@ def _provision_digital_ocean(self):
 
 		try:
 			for group in self.get_security_groups():
-				self.client().firewalls.assign_droplets(group, {"droplet_ids": [self.instance_id]})
+				self.client().firewalls.add_tags(group, {"tags": [droplet_tag]})
```
tanmoysrt, line 520: Maybe we should add the tag during cluster setup, when we actually create the firewall.

Aradhya-Tripathi: Will add that this is just for retrospective fix in case some other region (Toronto) does not have this already

tanmoysrt: maybe we need to detach the firewall from existing servers, add tag on existing servers and  on firewall also -- this part mostly need to be do manually from sys console.

PR #6257, opened 2026-04-26
feat(bench): Worker allocation based on plan

Aradhya-Tripathi:
Provide minimum floor guarantees based on site plans for worker allocation.
This weighted fair queue (ish) implementation now guarantees minimum worker _share_ allocation to each site plan, and accounts for all the sites in a bench (site-plan level worker allocation)

To consider: 
- Manage genuine resource contains on the server.
- Estimates on worker slots are currently based purely on RAM (resulting in 250 + Gunicorn workers for most public serves), however since we are spawning processes makes sense to check the number of vCPUs as well.
- Disproportionately large benches (public benches) off balance the Gunicorn ideal case numbers `(2 * $num_cpu) + 1` worker logic.


Aradhya-Tripathi:
Dropped after production tests; and v2 plan

PR #6260, opened 2026-04-27
fix(press-job): Save doc after changing status to running

tanmoysrt:


tanmoysrt:
@mergify backport release/2026-W18

PR #6262, opened 2026-04-27
fix(dashboard): Remove unwanted `br` tag in roles' relaxed mode switch

ssiyad:


siduck:
@mergify backport master

PR #6269, opened 2026-04-27
feat: changelog pages

siduck:
Just a static site built using astro, content is via md


siduck:
@mergify backport master

PR #6273, opened 2026-04-27
feat(compute): Public IP removal

the-bokya:
Implements necessary methods for public IP removal for Frappe Compute

### DIFF
--- a/press/press/doctype/ip_removal_log/ip_removal_log.py
+++ b/press/press/doctype/ip_removal_log/ip_removal_log.py
@@ -115,7 +115,6 @@ class IPRemovalLog(Document, StepHandler):
 		doc.save()
 
 		try:
-			cluster = frappe.get_doc("Cluster", self.cluster)
 			ansible = Ansible(
 				playbook="nat_iptables.yml",
 				server=doc,
@@ -123,7 +122,7 @@ class IPRemovalLog(Document, StepHandler):
 				port=doc._ssh_port(),
 				variables={
 					"nat_gateway_ip": frappe.cache.get_value(f"{self.name}:{self.nat_server}"),
-					"is_frappe_compute": cluster.cloud_provider == "Frappe Compute",
+					"is_frappe_compute": doc.provider == "Frappe Compute",
 				},
 			)
 			self.handle_ansible_play(step, ansible)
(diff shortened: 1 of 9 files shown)


press/press/doctype/ip_removal_log/ip_removal_log.py
```diff
@@ -115,13 +115,15 @@ def remove_ip_and_add_nat_conf(self, step):
...
 		try:
+			cluster = frappe.get_doc("Cluster", self.cluster)
 			ansible = Ansible(
 				playbook="nat_iptables.yml",
 				server=doc,
 				user=doc._ssh_user(),
 				port=doc._ssh_port(),
 				variables={
 					"nat_gateway_ip": frappe.cache.get_value(f"{self.name}:{self.nat_server}"),
+					"is_frappe_compute": cluster.cloud_provider == "Frappe Compute",
```
phot0n: you can get this from the server doc itself

phot0n (commented):
other places to look at

https://github.com/frappe/press/blob/6610cd4db9c5760d24d614e1e0672a4b5e95173b/press/press/doctype/cluster/cluster.py#L1823

https://github.com/frappe/press/blob/6610cd4db9c5760d24d614e1e0672a4b5e95173b/press/press/doctype/virtual_machine/virtual_machine.js#L101-L107

https://github.com/frappe/press/blob/6610cd4db9c5760d24d614e1e0672a4b5e95173b/press/press/doctype/virtual_machine/virtual_machine.py#L2235

https://github.com/frappe/press/blob/6610cd4db9c5760d24d614e1e0672a4b5e95173b/press/press/doctype/virtual_machine/virtual_machine.py#L211-L212

---

some notes (if they help):
- nat servers need to be able to accept all inbound traffic from the private network
- for failover since you dont have secondary private ip, you'll need to extend the nat failover to support that - plus static ip attach/detach should also be there (for public ip to remain the same)

the-bokya:
> * for failover since you dont have secondary private ip, you'll need to extend the nat failover to support that - plus static ip attach/detach should also be there (for public ip to remain the same)

Or I can implement secondary private IP too. Static IP attach/detach will require a few extra steps. A portion of the functionality is on my list.

the-bokya:
@mergify backport master

PR #6292, opened 2026-04-29
feat: Billing for static ip

phot0n:
Cases:
- attach-detach (on the same vm)
- vm terminated
- say the static ip was moved to another vm (attach-detach cycle of original vm has to be tracked and the attach-detach of the new vm as well)

### DIFF
--- a/press/press/doctype/database_server/database_server.py
+++ b/press/press/doctype/database_server/database_server.py
@@ -24,6 +24,7 @@ from press.press.doctype.database_server_mariadb_variable.database_server_mariad
 	DatabaseServerMariaDBVariable,
 )
 from press.press.doctype.server.server import PUBLIC_SERVER_AUTO_ADD_STORAGE_MIN, Agent, BaseServer, Server
+from press.press.doctype.static_ip_log.static_ip_log import create_static_ip_log
 from press.runner import Ansible
 from press.utils import get_press_base_url, log_error
 from press.utils.database import find_db_disk_info, parse_du_output_of_mysql_directory
@@ -231,7 +232,7 @@ class DatabaseServer(BaseServer):
 	def on_update(self):
 		self.publish_linked_server_realtime_update()
 
-		if self.flags.in_insert or self.is_new():
+		if self.flags.in_insert:
 			return
 
 		if self.is_replication_setup and self.auto_purge_binlog_based_on_size:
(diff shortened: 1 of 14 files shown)


press/press/doctype/static_ip_log/static_ip_log.py
```diff
@@ -0,0 +1,91 @@
...
+		if not plan:
+			frappe.throw(f"No active Static IP Plan found for provider {self.provider}")
+
+		if frappe.db.exists(
+			"Subscription",
+			{
+				"enabled": 1,
+				"document_type": self.server_type,
+				"document_name": self.server,
+				"plan": "Static IP Plan",
```
shadrak98: this should be `plan_type` instead of `plan`. 

```suggestion
				"plan_type": "Static IP Plan",
				"plan": plan
```

press/press/doctype/subscription/subscription.py
```diff
@@ -240,12 +241,24 @@ def is_usage_record_created(self, date=None):
...
+		elif self.interval == "Hourly":
+			last_usage_record = frappe.db.get_value(
+				"Usage Record", filters, ["date", "time"], order_by="creation desc", as_dict=True
+			)
+			if not last_usage_record:
+				return False
+
+			last_datetime = datetime.datetime.combine(
+				last_usage_record.date, frappe.utils.get_time(last_usage_record.time)
+			)
```
phot0n, line 259: we could use `creation` as well instead of doing this

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -1479,23 +1479,26 @@ def update_servers(self):  # noqa: C901
...
 				if doctype in ["Server", "Database Server"]:
-					frappe.db.set_value(doctype, server, "ram", self.ram)
+					server.ram = self.ram
 				if doctype in ("NAT Server",):
-					frappe.db.set_value(doctype, server, "secondary_private_ip", self.secondary_private_ip)
+					server.secondary_private_ip = self.secondary_private_ip
+
+				server.status = status_map[self.status]
+				server = server.save()
+
```
phot0n, line 1509: i wonder if doing this would lead to a lot of

"doc has been modified" errors

phot0n:
TODO:
- [x] smarter way to add usage records (shifted to daily)
- [x] only for aws

phot0n:
audit log test failure happening in other places as well: https://github.com/frappe/press/actions/runs/25908962071/job/76149597021?pr=6441

press/press/doctype/static_ip_log/static_ip_log.py
```diff
@@ -0,0 +1,142 @@
...
+				{
+					"enabled": 1,
+					"document_type": self.server_type,
+					"document_name": self.server,
+					"plan_type": "Static IP Plan",
+				},
+				"name",
+			)
+		):
+			frappe.throw(f"No active Static IP subscription found for {self.server_type} {self.server}.")
```
phot0n, line 128: ah this should be moved to `_check_if_can_disable_subscription`

press/press/doctype/database_server/database_server.py
```diff
@@ -244,9 +245,11 @@ def on_update(self):
 		):
 			self.update_memory_limits()
 
-		if not self.is_new() and self.has_value_changed("team"):
+		if self.has_value_changed("team"):
 			self.update_subscription()
```
phot0n, line 249: need to take care of updation of static ip subscription if team got changed as well

phot0n:
also check unified server, since the doc is supposed to be treated as one in that case, will it cause errors?

PR #6295, opened 2026-04-29
fix(incident): Investigator wait time should be in seconds unit

tanmoysrt:
(cherry picked from commit 16d1b6c)


Aradhya-Tripathi (approved):
wow

PR #6310, opened 2026-05-01
fix(notifications): Reset pagination params on tab change

siduck:
https://github.com/user-attachments/assets/2a199976-8a2b-474c-97c7-574bcc7bdb93


siduck:
@mergify backport master

PR #6312, opened 2026-05-01
fix(site-group-deploy): use frappe.qb.DocType instead of frappe.qb.doctype

phot0n:


phot0n:
@mergifyio backport master

PR #6315, opened 2026-05-02
fix(site): correctly call is_setup_wizard_complete method

phot0n:


phot0n:
@mergifyio backport master

PR #6346, opened 2026-05-06
fix(marketplace): Fix options for version payload error

regdocs:
ref: https://support.frappe.io/helpdesk/tickets/67492


regdocs:
@mergifyio backport master

PR #6357, opened 2026-05-06
feat(compute): Virtual Disk Snapshots

the-bokya:


the-bokya:
@mergify backport master

PR #6360, opened 2026-05-07
feat(marketplace): Add is_desk_user function to check system user status

prathameshkurunkar7:


prathameshkurunkar7:
@mergify update

PR #6364, opened 2026-05-07
fix(compute): Add snapshots background method

the-bokya:


the-bokya:
@mergify backport master

PR #6366, opened 2026-05-07
fix(compute): Pending status for pending snapshot

the-bokya:


the-bokya:
@mergify backport master

PR #6367, opened 2026-05-07
fix(onboarding): page crash due to accessing null field in trialSite

siduck:
image

### Reason

- user has 0 trialSites as per api response
-  but the site_created field is true
- Because the user created site year ago but it got archived
- so the trialSite value here is always undefined and we are trying to access its value because site_created is true? its true because the user created site year ago!

This is a rare edgecase

Maybe the backend api must return archived sites in trial sites. For now ive added my check on the frontend which fixes the issue

### DIFF
--- a/dashboard/src/App.vue
+++ b/dashboard/src/App.vue
@@ -108,7 +108,7 @@ provide('team', team);
 provide('session', session);
 
 watch(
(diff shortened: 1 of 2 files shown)


siduck:
@mergify backport master

PR #6369, opened 2026-05-07
feat(registry): Add multi cluster harbor registries

Aradhya-Tripathi:
Utility to configure and provision cluster level registries.

Moving away from the out-of-the-box docker registries and using [Harbor](https://github.com/goharbor/harbor)

Biggest reason being the [seamless GC](https://goharbor.io/docs/edge/administration/garbage-collection/) along with configurable layer level support.


Aradhya-Tripathi:
Dropped in favour of V2

PR #6370, opened 2026-05-07
refactor(newServer): improve wording of auto-addon section

siduck:
- Removed useless `<br/>` tags 
- Also bullet point glyphs 😂 , prose class should take care of it, idk why we even had those.  

### Before : 

image

### After : 

image


siduck:
@mergify backport master

PR #6348, opened 2026-05-06
feat(mariadb): Improvements For MariaDB Memory Usage / Spike Reduction

tanmoysrt:
- [x] Add plays to upgrade to cgroup v2
- [x] Add tcmalloc_release_rate configuration support
- [x] In mariadb-monitor, add external db healthcheck for better db health validation
- [x] Auto-trim memory of Database Server
  - [x] Monitor PSI, hit rate / failcnt of `MemoryHigh`, Swap usage of database. Based on that trigger memory reclaim.
  - [x] Add TCMalloc UDF in mariadb to be able to call `ReleaseMemory`
  - [x] Ask TCMalloc to release it's free memory back
  - [x] Ask db to trim it's innodb buffer size to give back some memory. Check `Innodb_buffer_pool_resize_status` for status
  - [x] Force `swapoff & swapon` to tell mariadb to stop using cache.
  - [x] Need to make sure it happens in controlled way and doesn't make the situation worse.
 

Perf schema will be disabled on shared server and we will take care of it later in seperate pr : https://github.com/frappe/press/issues/6380

PR #6388, opened 2026-05-08
feat(database): Install tcmalloc udf hook in mariadb monitor install

tanmoysrt:
Related https://github.com/frappe/press/pull/6348


tanmoysrt:
@mergify backport master

PR #6395, opened 2026-05-09
fix(theme): comment out system theme for now

siduck:
- This was setting the theme to system and as a result dark theme was set for users automatically which isnt good since dark theme is still in beta.


siduck:
@mergify backport master

PR #6407, opened 2026-05-11
fix(search): add condition for hiding team link

siduck:
https://github.com/frappe/security/issues/214


siduck:
@mergify backport master

PR #6409, opened 2026-05-11
refactor(serverOverview): Use computed property for servers()

siduck:
## Description
- removes ugly logic in the markup
- Previous logic was creating a divider like ui cuz 
  example : [ "server", "database server" ]
  
  now we loop over and render 2 cards as per the array 
  
  but if we have unified server then we were not rendering the database server* item. But the outer wrapper had border on it. hence causing divider. 🫠 
  
 ### Before: 

image

image

### After:  

image


siduck:
@mergify backport master

PR #6412, opened 2026-05-11
refactor(sidebar): Reuse Item component & pure css for collapsable animation

siduck:
### Before
- Since search, notifcation btns arent link elements i made separate components for them duplicating the style logic of the Item.vue in sidebar

### After
- Make single Item.vue to allow `:is` prop so we can make it render a link tag/button or whatever 

## Remove JS for ItemGroup

- ItemGroup is the collapsable section you see in the sidebar

### Before 

- Simple vue state to toggle the itemgroup
- CollapseTransition wrapper component for animating it 

### Now
- native `<details>` element to toggle content. no need of js
- use grid-template-rows for pure css animations

### DIFF
--- a/dashboard/src/components/navigation/sidebar/Item.vue
+++ b/dashboard/src/components/navigation/sidebar/Item.vue
@@ -1,11 +1,13 @@
 <script setup lang="ts">
 import type { NavItemProps } from "./types";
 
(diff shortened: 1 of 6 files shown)


siduck:
@mergify backport master

PR #6435, opened 2026-05-14
fix: Add missing biome.json & remove eslint/prettier

siduck:


siduck:
@greptile-apps rereview

PR #6439, opened 2026-05-15
feat(plans): Move Site and Server plans to fixtures

regdocs:
Fixtures should treat the database as the source of truth. Any changes made directly in the database are expected to be synced back to the corresponding JSON fixture files.

Document deletions are not handled during fixture import. The framework performs an optimistic merge, meaning it updates existing documents and inserts missing ones, but it does not remove records that are absent from the fixture files.

If a document with the same name already exists, its contents will be overwritten with the data from the JSON fixture.

To export:
```sh
bench --site frappe.cloud export-fixtures
```

To import:
```sh
bench --site frappe.cloud migrate
```


regdocs:
@mergifyio backport master

PR #6450, opened 2026-05-18
fix(sidebar): @ getting clipped due to truncate class

siduck:
### Before: 

image

### After: 

image

leading-none was causing it


siduck:
@mergify backport master

PR #6461, opened 2026-05-19
refactor(server): Clean up server archival exception handling using an inner function

phot0n:
* also make the job a daily runner rather than hourly


phot0n:
@mergifyio backport master

PR #6469, opened 2026-05-20
fix(backend): Account Request permission check picks wrong team name

ssiyad:
This is extremely flaky and does not happen to all records. The root cause is a mismatch in team names present in account request doc. The identifier is sometimes the user's email and sometimes the random hash. This patch handles both cases.


ssiyad:
@greptileai review

PR #6557, opened 2026-05-29
fix: Pick only enabled Root Domains for Product Trials

the-bokya:


the-bokya:
@mergify backport master

Bowrna:
@the-bokya  I think we have to fix in doctype product_trial.py for create_standby_site too like mentioned in this review. Otherwise we will have sites trying to create in standby site background job

PR #6589, opened 2026-06-02
fix: Notify user if all backup attempts fail

AyshaHakeem:
- If all attempts to backup a site fails in a day, send an email to user


AyshaHakeem:
@mergify backport master

PR #6600, opened 2026-06-03
chore(changelog):  Update astro version

siduck:


siduck:
@mergify backport master

PR #6606, opened 2026-06-03
fix(deploy-ui): agentjobs not saving in state

siduck:
- Also attach socket stuff only if deploy is running


siduck:
@mergify backport master

PR #6608, opened 2026-06-03
fix(deploy-ui): Reload doc resource as fallback after an interval 

siduck:
- Pipeline status isnt updated if one of the build steps fails, i have to reload the page to see the failiure status. So for now reload the resource after 1 minute after actual failure detcted


siduck:
@mergify backport master

PR #6622, opened 2026-06-05
fix(deploy-candidate): Show app name in fetch error notification

regdocs:
When the fetch error fires from send_build_instructions_and_add_build_steps (outside the "apps" stage), the else branch showed a generic message with no app name. The GithubFetchError already carries the app name in args[1], so extract it and include it in the notification message.


siduck:
@mergify backport master

PR #6627, opened 2026-06-05
fix(manage-product-warranty): Fix site warranty count query

regdocs:
- Added to filter out sites with status Archived or Suspended


regdocs:
@mergifyio backport master

PR #6628, opened 2026-06-05
fix(signup): Never fall back to apex domain for product trial signups

balamurali27:
## Problem

New signups landed on the apex `frappe.cloud` instead of a per-cluster subdomain (`m.frappe.cloud`, `v.frappe.cloud`). AWS Route 53 has a record limit for each hosted zone, so the apex is not usable for new sites. New sites must spread across the cluster subdomains.

The signup domain resolver fell back to the apex without a message in two cases:

- cluster resolution returned `None`, because the geo or IP lookup was down or rate-limited, or because the Account Request had no country
- the nearest cluster had no `*.frappe.cloud` subdomain (UAE, for example)

`get_request` and `create_standby_site` each had a copy of the resolution logic, and both copies had the hole. `create_site` trusted the domain that the frontend posted. The apex passes validation, because it is an enabled Root Domain.

## Fix

Resolution now lives on `ProductTrial`, with one rule: never the apex.

- **`get_signup_domain(cluster)`** resolves to the enabled `*.{apex}` subdomain of the cluster. If there is none, it falls back to the nearest enabled subdomain (UAE goes to Bahrain), then to the subdomain of the default cluster of the apex. It returns the apex only when no subdomain exists.
- **`get_cluster_for_request(account_request)`** moves here from the api module, so `get_request` and `create_site` share it.
- **`create_site`** resolves the domain again on the server. An apex from a stale frontend or a failed geo lookup can no longer pass.
- **`create_standby_site`** uses the same resolver for pooled standby sites.
- **utils**: `CLUSTER_COORDINATES` and `_haversine_distance` move to module level. A new `get_nearest_cluster_among(target, candidates)` serves the nearest-subdomain fallback.

`press/saas/README.md` has the description.

## Tests

New tests in `test_product_trial.py` cover the exact cluster subdomain, the nearest fallback (UAE to Bahrain), and the apex only when no subdomain exists. All pass. `ruff` and pre-commit are clean.

## Note

This PR corrects new signups and new standby sites only. Standby sites and signed-up sites from before can still have apex `frappe.cloud` host names. Those need a separate cleanup.


balamurali27:
@copilot what happens when no geolocation info is there in signup in this pr

regdocs:
@copilot are you therE?

regdocs:
@copilot hi

balamurali27:
@regdocs please reopen when you're working on it

regdocs:
@balamurali27 will add to tasks this month

PR #6630, opened 2026-06-05
fix(roles): Change deletion condition

ssiyad:
- Admins should be able to delete roles
- Conditionally render delete button
- Show toast on error


ssiyad:
The comment from @greptileai seems to be invalid. We don't bother with pre-defined roles in existing page.

PR #6645, opened 2026-06-08
fix(account): Clear request key after accepting invitation

ssiyad:


ssiyad:
@copilot 

Fix the following 1 code review issue. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 1
press/api/account.py:262-292
**Parallel invite path still doesn't clear the key**

`accept_team_invite` (line 262) also fetches the Account Request by `request_key` via `get_account_request_from_key`, creates the team member, and returns — but never calls `db_set("request_key", None)`. An existing user accepting a team invite via this path can replay the same invite link indefinitely, getting re-added to the team each time `create_user_for_member` is called.

ssiyad:
@mergify backport master

PR #6646, opened 2026-06-09
fix(support): Read web.error.log from bench, not site

balamurali27:
## Summary

- `get_web_error_log` was calling `Site.get_server_log("web.error.log")`, which hits the agent at `benches/{bench}/sites/{site}/logs/web.error.log` — a per-site path that is empty or does not exist
- Gunicorn's stderr is a bench-level file at `benches/{bench}/logs/web.error.log`, accessible via `Bench.get_server_log`
- Switched the collector to read from the bench; updated the test mock accordingly

## Why it matters

Startup failures — syntax errors in `app.py`, import errors that crash gunicorn workers — are logged only to the gunicorn error log. The old code was silently returning `error_count: 0` even when the site was down because of such a crash (verified on staging: probe reported HTTP 500, but `web_error_log` showed no errors).

## Test plan

- [ ] Existing `TestSupportAgentInvestigation` tests pass (mock now attached to `bench_doc.get_server_log`)
- [ ] On staging: run an investigation against a site with a known gunicorn error; confirm `web_error_log.error_count > 0`

### DIFF
--- a/press/incident_management/support_agent/collectors.py
+++ b/press/incident_management/support_agent/collectors.py
@@ -55,7 +55,7 @@ def collect_site_context(site_name: str) -> dict[str, Any]:
 		"bench_processes": get_bench_process_status(site.get("bench")),
 		"site_uptime": get_site_uptime(site_name),
 		"site_performance": get_site_performance_summary(site_name, site.get("bench")),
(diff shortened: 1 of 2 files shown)


balamurali27:
@mergifyio backport release/2026-W24

PR #6658, opened 2026-06-10
fix(site-domain): Allow adding custom domains if the domain ip matches the proxy ip

phot0n:


phot0n:
@mergifyio backport release/2026-W24

PR #6661, opened 2026-06-10
fix(deploy-ui): Show AppVersions dialog for pipeline view too

siduck:


siduck:
@mergify backport master

PR #6666, opened 2026-06-10
chore(CODEOWNERS): Remove `@ssiyad` as catch-all

ssiyad:


ssiyad:
@mergify backport master

PR #6670, opened 2026-06-10
fix(server): Don't allow server creation for 2 or more unpaid invoices

shadrak98:
image


phot0n:
add a server side check as well?

shadrak98:
@phot0n missed your suggestion. Will add it later

PR #6675, opened 2026-06-10
fix(server): remove and add nat config back if applicable during resize server job

phot0n:
* added as a temp fix untill we find the permanent fix for this

### DIFF
--- a/press/press/doctype/press_job/jobs/resize_server.py
+++ b/press/press/doctype/press_job/jobs/resize_server.py
@@ -19,6 +19,7 @@ if TYPE_CHECKING:
 class ResizeServerJob(PressJob):
 	@flow
 	def execute(self):
+		self.remove_nat_config_if_applicable()
 		self.stop_virtual_machine()
 		self.wait_for_virtual_machine_to_stop()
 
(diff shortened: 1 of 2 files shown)


press/press/doctype/press_job/jobs/resize_server.py
```diff
@@ -147,3 +171,7 @@ def on_press_job_failure(self, workflow: PressWorkflow):
 
 		from_plan: ServerPlan = frappe.get_doc("Server Plan", plan_change.from_plan)
 		self.server_doc._change_plan(from_plan)
+
+		self.wait_for_virtual_machine_to_start()
+		self.wait_for_server_to_be_accessible()
+		self.add_nat_config_if_applicable()
```
phot0n: are these fine? like will the failure callback keep on looping as well?

press/press/doctype/press_job/jobs/resize_server.py
```diff
@@ -125,6 +149,8 @@ def increase_disk_size(self):
 
 	def on_press_job_failure(self, workflow: PressWorkflow):
 		self.start_virtual_machine()
+		# TODO: fix this; this won't really do much as the vm might've just been brought up but it's better than nothing
+		self.add_nat_config_if_applicable()
```
phot0n, line 153: better to move this to a diff job

phot0n:
merging this for now - will raise separate pr for adding a press job which adds nat config back on failure properly

PR #6678, opened 2026-06-11
feat(server): Add skip-standby-site-creation flag

regdocs:
Adds a `Skip Standby Site Creation` Check field to the Server doctype. When enabled, the scheduler will not pick that server when replenishing standby site pools.

The flag is respected in all three pool systems:
- `product_trial.get_server_from_cluster` (product trial pools)
- `saas_site.get_saas_bench` (legacy SaaS app pools)
- `erpnext_site.get_erpnext_bench` (ERPNext site pool)

### DIFF
--- a/press/press/doctype/server/server.json
+++ b/press/press/doctype/server/server.json
@@ -251,6 +251,12 @@
 			"label": "Use For New Sites"
 		},
 		{
+			"default": "0",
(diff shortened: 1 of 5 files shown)


phot0n:
@mergifyio backport master

PR #6686, opened 2026-06-11
fix(product-warranty-switch): Gate warranty enabling only

regdocs:
- Fixes "failed to change product warranty" error


regdocs:
@mergifyio backport master

PR #6690, opened 2026-06-11
fix: Standby site updates and start-benches timeout

balamurali27:
## What

A few small fixes:

- **Always update standby sites** — `is_site_in_deploy_hours` now returns early for standby sites so they update regardless of deploy hours.
- **Fetch `is_standby`** — the standby guard above relied on `site.is_standby`, but `sites_with_available_update` did not select that field. On a `frappe._dict` the missing field returned `None` (falsy), so the guard was a silent no-op. The field is now fetched.
- **Increase timeout for start benches job** — `start_active_benches` is now enqueued on the `long` queue with a 3600s timeout, since the ansible `start_benches` playbook can exceed the 300s default.

## Why

The standby fix landed but never actually fired because the required field wasn't queried; this PR makes it work.


AyshaHakeem:
@Mergifyio backport master

PR #6691, opened 2026-06-11
fix(server): Persist auto increase storage disable on DB servers

regdocs:
`configure_auto_add_storage` is always dispatched on the app server from the dashboard ($appServer), with the real target passed as the `server` argument. The enable branch routed to the right doc via `server`, but the disable branch hardcoded `self`, so disabling on a DB/replication server cleared the flag on the app server instead. The target server kept `auto_increase_storage = True` and the checkbox reverted on reload.

Resolve the target doc once and use it in both branches.


regdocs:
@greptileai rereview

ssiyad (approved):

regdocs:
@mergifyio backport master

PR #6693, opened 2026-06-11
fix(deploy-ui): fallback jobstep.duration with empty string

siduck:
Prevents Nan from being shown in Jobsteps duration


siduck:
@mergify backport master

PR #6697, opened 2026-06-11
fix(site): pick only enabled root domains when fetching options for new site

phot0n:


phot0n:
@mergifyio backport master

PR #6701, opened 2026-06-12
fix(deploy-ui): Make page responsive

siduck:


siduck:
@mergify backport master

PR #6705, opened 2026-06-13
fix(team): Better user experience

ssiyad:


ssiyad:
@mergify backport master

PR #6716, opened 2026-06-15
fix(deploy-ui): Show team instead of owner in deploys list

siduck:
The deploys list was displaying the document owner instead of the team that initiated the deploy. 

Joined team.user to get the team's user email and exposed the team field via the whitelist in release_pipeline.py.

### Before 
image


siduck:
@mergify backport master

Issue #6208, opened 2026-04-21
Prevent certain bench commands from running in production


regdocs:
### Information about bug

Like those for running tests, or get-app or install-app which can otherwise disrupt the synchronisation between press and the local bench environment.

eg. ref: 
- someone ran tests in prod: https://support.frappe.io/helpdesk/tickets/65699
- someone ran get-app via ssh: https://support.frappe.io/helpdesk/tickets/64661

### Steps to Reproduce

N/A

### Relevant log output / Stack trace / Full Error Message.

```shell

```

PR #6718, opened 2026-06-15
feat(bench): Trigger install of SSH command guard on benches

regdocs:
Pairs with the [agent-side bench wrapper](https://github.com/frappe/agent/pull/543) that blocks destructive commands (drop-site, migrate, restore, etc.) run directly over SSH on production benches, which would put a site's state out of sync with Frappe Cloud.

Press owns the blocklist (BLOCKED_BENCH_COMMANDS) and sends it to the agent, so it can change without rebuilding the bench image. The guard lives on the container filesystem rather than the image, so it is (re)installed every time a bench turns Active - covering both new benches and fresh containers from a redeploy. backfill_command_guard_on_active_benches covers benches already running; it is left out of patches.txt so it does not fire jobs before the agent endpoint is deployed.

The enforcement can't live in the bench CLI (pinned/immutable in prod), hence routing it through press > agent.

### DIFF
--- a/press/agent.py
+++ b/press/agent.py
@@ -125,6 +125,16 @@ class Agent:
 			"Update Bench Configuration", f"benches/{bench.name}/config", data, bench=bench.name
 		)
 
+	def setup_bench_command_guard(self, bench: str):
+		from press.press.doctype.bench.bench import BLOCKED_BENCH_COMMANDS
+
+		return self.create_agent_job(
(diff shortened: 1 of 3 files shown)


balamurali27:
fixes #6208

regdocs:
Better implementation is to include this as part of frappe/bench, hence closing.

PR #6731, opened 2026-06-17
fix(github): Refresh access token after app (re)installation

regdocs:
Token refresh in the GitHub callback was gated entirely on an OAuth `code` being present:

    if code and state:
        obtain_access_token(code, team)

When a user (re)installs the Frappe Cloud GitHub App, GitHub redirects back to /github/authorize with installation_id + setup_action + the echoed state but no `code` (a code only appears when user-authorization runs). So the branch was skipped, the stale token survived in the Team doc, and the next "Add from GitHub" used a dead token.

Now when the callback arrives with a valid state but no code (and no error), start GitHub user-authorization so the user bounces back with a code that refreshes the token. If the app is already authorized this is a silent round-trip; a denied authorization falls through to a plain dashboard redirect, so there is no loop. Code-bearing callbacks are unchanged.

The guest-login redirect now preserves state without a code too, so a guest landing on the install callback completes the same refresh after logging in.

### DIFF
--- a/press/api/github.py
+++ b/press/api/github.py
@@ -27,6 +27,7 @@ if TYPE_CHECKING:
 
 
 DEFAULT_GITHUB_REDIRECT_PATH = "/dashboard"
+GITHUB_OAUTH_AUTHORIZE_URL = "https://github.com/login/oauth/authorize"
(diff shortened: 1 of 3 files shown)


regdocs:
@mergifyio backport master

PR #6734, opened 2026-06-17
fix(release-pipeline): Emit status update event only after commit

siduck:
### Description 

- Many times when a some step in build fails, we dont get the emitted event for release pipeline. As a result the status  badge just shows Running but after page reload it shows the correct one i.e `Failure`

- This PR fixes it


siduck:
@mergify backport master

PR #6757, opened 2026-06-20
fix(deploy-candidate): Fix Schedule Build and Deploy popup crash

regdocs:
The shared desk button handler builds a link with `data.message.message`, but the two methods it serves return different shapes:

- `build` returns `{error, message: <build_name>}`
- `schedule_build_and_deploy` returns `{error, name: <build_name>}`

So for Schedule Build and Deploy, `data.message` is undefined and `data.message.message` throws `TypeError: Cannot read properties of undefined (reading 'message')`. The exception aborts the handler before the success msgprint and `frm.refresh()`, so the button appears to do nothing (the Scheduled build is actually created server-side and runs via run_scheduled_builds). For `build` the same access silently produced a `/app/deploy-candidate-build/undefined` link.

Read `data.message ?? data.name` so the build name resolves for both methods. The Python return contract is left unchanged on purpose: release_pipeline.py, api/bench.py and release_group.py read the `name` key off schedule_build_and_deploy's result.

Regression introduced in 373f237d0 (data.message -> data.message.message).

The rest of the diff is biome stripping semicolons: the file predates the
repo's `semicolons: asNeeded` rule and pre-commit reformats it on touch.

### DIFF
--- a/press/press/doctype/deploy_candidate/deploy_candidate.js
+++ b/press/press/doctype/deploy_candidate/deploy_candidate.js
@@ -7,62 +7,62 @@ frappe.ui.form.on('Deploy Candidate', {
 			return {
 				query: 'press.press.doctype.deploy_candidate.deploy_candidate.desk_app',
 				filters: { release_group: doc.group },
-			};
-		};
+			}
+		}
 		set_handler(
 			frm,
 			'Complete',
 			'build',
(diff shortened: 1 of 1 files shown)


regdocs:
@mergifyio backport master

PR #6764, opened 2026-06-22
fix(site): Allow enabling product warranty irrespective of cooldown

regdocs:
### Problem

Enabling product warranty for a dedicated-server site was gated by **both**
the server quota and the per-site change cooldown.

The cooldown exists to deter partners from rotating a limited number of
support slots across many sites — disable a slotted site, enable an
unslotted one — and so getting support for more sites than they pay for.

But it also blocked a legitimate case:

> A customer has 5 supported sites (quota: 5). They disable support for one
> of them. Later they realise they still need it and try to re-enable — but
> the cooldown blocks them, even though re-enabling does **not** exceed quota.

They aren't misusing the quota, so they should be allowed.

### Insight

The cooldown is enforced **per-site**. The only enable the cooldown ever
blocked is a site **reclaiming the slot it itself just gave up** — exactly
the legitimate case above. Enabling a *different* site was never blocked by
the cooldown; it's gated by **quota** (a slot must be freed first).

What actually deters rotation is the cooldown on **disabling**: once a
partner frees a slot, that site is locked, so they can't rapidly
free-and-reassign support across many sites.

### Change

Split the two gates by direction:

- **Enabling** → gated by **quota only**. A site can reclaim a free slot at
  any time, irrespective of cooldown.
- **Disabling** → gated by the **cooldown only** (freeing a slot never needs
  quota).

Re-enabling only succeeds when a slot is genuinely free — if the freed slot
was already taken by another site, quota is full and the reclaim is correctly
rejected. So the quota ceiling and the anti-rotation deterrent are both
preserved.

### Backend — `press/api/site.py`

`_check_warranty_restrictions` now branches on direction: enabling checks
quota and returns; disabling checks the cooldown.

### Frontend — `dashboard/src/components/ManageProductWarrantyDialog.vue`

The toggle and Save button mirror the backend (quota gates enabling, cooldown
gates disabling). The "Next change available after" row now appears only while
warranty is **on** and the cooldown is still active — the single state where
the cooldown blocks the action. It's hidden when warranty is off, since
enabling ignores the cooldown there.

### Tests — `press/api/tests/test_site.py`

`TestCheckWarrantyRestrictions` updated to the new policy, adding:
- enabling is allowed during cooldown when a slot is free (the fix)
- enabling is still blocked when quota is exhausted, even during cooldown

### Note

The disable cooldown is now the **sole** deterrent against slot rotation, so
the `site_warranty_change_cooldown` value on Server matters more than before.

### DIFF
--- a/dashboard/src/components/ManageProductWarrantyDialog.vue
+++ b/dashboard/src/components/ManageProductWarrantyDialog.vue
@@ -66,10 +66,23 @@ const nextChangeAvailableOn = computed(() => {
 	}
 })
 
+const hasWarrantyQuota = computed(
+	() => site.doc?.dedicated_server_warranty_limit?.available > 0,
+)
+
+const cooldownPassed = computed(
+	() => nextChangeAvailableOn.value === 'Available Now',
+)
+
+// Enabling warranty is gated only by available quota (the cooldown does not
+// apply, so a site can reclaim a free slot any time); disabling is gated by the
+// cooldown.
+const isToggleDisabled = computed(() =>
+	isSupportEnabled.value ? !cooldownPassed.value : !hasWarrantyQuota.value,
+)
+
 const disablePrimaryAction = computed(
-	() =>
-		switchValue.value === isSupportEnabled.value ||
-		nextChangeAvailableOn.value !== 'Available Now',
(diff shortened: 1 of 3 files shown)


regdocs:
@mergifyio backport master

PR #6765, opened 2026-06-22
fix(site): Skip file-tree walk when refreshing only database usage

balamurali27:
## Problem

`refresh_database_usage` (polled by the dashboard's Site Overview / Database Analyzer pages every few seconds) goes through the generic `sync_info`, which calls the agent's `get_usage`. That recursively walks the site's **entire file tree** (public, private, backups) just to sum file sizes — even though the caller only wants the database size.

On sites with large file trees the walk exceeds the gunicorn timeout and kills agent web workers (`WORKER TIMEOUT`). Because the dashboard re-polls every ~3s, a single bloated site keeps the walk running back-to-back.

## Fix

Thread a `database_only` flag through `get_site_info` → `fetch_info` → `sync_info`:

- In that mode the agent is asked (via `?database_only=1`) for just the cheap database size.
- `_sync_database_usage` records a `Site Usage` row that **carries the last-known file totals forward** instead of recomputing them, so disk-usage reporting is unaffected.
- Both DB-refresh callers (the `refresh_database_usage` parser-off branch and the `process_refresh_database_usage_job_update` callback) pass `database_only=True`.

## Deploy note

**Paired PR:** frappe/agent#548

Requires the matching agent change to honour `?database_only` (**frappe/agent#548**). Without it the agent ignores the param and falls back to the full walk (current behaviour), so the two can be deployed in either order safely.

## Test

Added `test_sync_info_database_only_refreshes_db_size_and_carries_files_forward` — asserts a database-only sync refreshes the DB size while carrying `public/private/backups` forward, and that `fetch_info` is called with `database_only=True`.

### DIFF
--- a/press/agent.py
+++ b/press/agent.py
@@ -1198,8 +1198,11 @@ Response: {reason or getattr(result, "text", "Unknown")}
 			result = self.get(f"benches/{site.bench}/sites/{site.name}/sid")
 		return result and result.get("sid")
 
-	def get_site_info(self, site):
-		result = self.get(f"benches/{site.bench}/sites/{site.name}/info")
+	def get_site_info(self, site, database_only=False):
+		path = f"benches/{site.bench}/sites/{site.name}/info"
+		if database_only:
+			path += "?database_only=1"
+		result = self.get(path)
 		if result:
 			return result["data"]
 		return None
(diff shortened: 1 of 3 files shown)


balamurali27:
recent f1-ksa stuck jobs incident

PR #6778, opened 2026-06-23
fix(product-trial): store tracked agent jobs as json

netchampfaris:
## Summary
- replace Product Trial Request child-table job tracking with a JSON field
- remove the Product Trial Request Agent Job child DocType
- keep required job lookup based on tracked JSON with agent_job fallback

## Verification
- jq empty press/saas/doctype/product_trial_request/product_trial_request.json
- python -c "import ast, pathlib; ast.parse(pathlib.Path('press/saas/doctype/product_trial_request/product_trial_request.py').read_text())"
- git diff --check

Frappe tests not run; site name required.


netchampfaris:
Updated for Greptile P2: malformed/non-list tracked_agent_jobs now logs an Error Log before falling back.\n\nNot adding a backfill patch for the old child table per follow-up decision; new writes no longer use the child table and agent_job remains the fallback for records without tracked_agent_jobs.

PR #6785, opened 2026-06-24
feat: Remove server plan fixtures

tanmoysrt:
It's linked with multiple dynamic fields which only exists on production.

Making it harder to roll-out new plans.


tanmoysrt:
@mergify backport master

PR #6789, opened 2026-06-24
feat(nat): Extend NAT server into DigitalOcean

20vikash:
- Previously, NAT servers were only implemented in AWS, Hetzner and Frappe Compute
- This PR extends that into DigitalOcean.


phot0n (commented):
mostly looks good! 

But i can't seem to find a method to remove ips from existing servers (?) - [this](https://github.com/frappe/press/blob/develop/press/press/doctype/ip_removal_log/ip_removal_log.py) tool can be used to remove the ips from servers (maybe we can extend this?)

also, some places to look at:
- https://github.com/frappe/press/blob/develop/press/press/doctype/cluster/cluster.py#L2073
- https://github.com/frappe/press/blob/bed0110c6a23a062a0b09599f1837df18ae9218e/press/press/doctype/virtual_machine/virtual_machine.js#L101-L108

---

are we also extending nat failover support for this?
we'd need to add support for secondary private ip (if present) - so that upon failover we can attach/detach the secondary private ip from the primary nat server to secondary one - so that we dont have to update the netplan configs in all the servers
(you can check this for that: https://github.com/frappe/press/blob/bed0110c6a23a062a0b09599f1837df18ae9218e/press/press/doctype/nat_failover/nat_failover.py#L44-L53)

phot0n:
@mergifyio backport master

PR #6792, opened 2026-06-24
feat(signup): Add new fields on signup

shadrak98:
image


shadrak98:
@siduck Can you review and fix the styling on the screen? Apparently the font-weight doesn't work consistently with checkbox as it does with text fields. 

cc: @nehasan16

siduck:
cleaned up the form a bit https://github.com/frappe/press/pull/6792/changes/910fca5e30bc94a1ebd5d51f89e28f81414209fa 

image

PR #6793, opened 2026-06-24
fix(site): Skip host_name update when site domain is not yet active

tanmoysrt:
- Agent jobs with persistent callback failures (≥ 5000) are now marked as Failure and a comment with the full exception traceback is added to the job, preventing endless retry loops
- Skip setting host_name in process_add_domain_job_update when the site domain is not yet Active, avoiding a race condition where the "Add Domain" job completes before "Add Domain to Upstream" sets the domain active
- https://github.com/frappe/press/pull/6794


tanmoysrt:
@greptile review

tanmoysrt:
@greptile review

PR #6797, opened 2026-06-24
feat(nat): IP removal for hetzner

20vikash:
Extend Hetzner into IP removal

### DIFF
--- a/press/playbooks/roles/nat_iptables/tasks/main.yml
+++ b/press/playbooks/roles/nat_iptables/tasks/main.yml
@@ -43,8 +43,8 @@
   when:
     - cloud_provider == "Hetzner"
     - network_gateway is defined
-    - should_setup_nat_gateway is defined
-    - should_setup_nat_gateway
+    - nat_gateway_ip is defined
+    - nat_gateway_ip
     - network_gateway
 
 - name: Configure private interface (Frappe Compute specific)
--- a/press/playbooks/roles/nat_iptables/tasks/main.yml
+++ b/press/playbooks/roles/nat_iptables/tasks/main.yml
@@ -37,6 +37,8 @@
   when:
     - cloud_provider == "Hetzner"
     - network_gateway is defined
+    - should_setup_nat_gateway is defined
+    - should_setup_nat_gateway
     - network_gateway
 
 - name: Configure private interface (Frappe Compute specific)
--- a/press/playbooks/roles/nat_iptables/tasks/main.yml
+++ b/press/playbooks/roles/nat_iptables/tasks/main.yml
@@ -1,7 +1,13 @@
 ---
 - name: Determine Hetzner interface
   set_fact:
-    primary_interface: "{{ ansible_interfaces | difference(['lo']) | first }}"
+    primary_interface: >-
+      {{
+        ansible_interfaces
+        | reject('equalto', 'lo')
+        | reject('equalto', 'docker0')
+        | first
+      }}
   when: cloud_provider == "Hetzner"
 
 - name: Add NAT gateway
--- a/press/press/doctype/ip_removal_log/ip_removal_log.py
+++ b/press/press/doctype/ip_removal_log/ip_removal_log.py
@@ -19,6 +19,7 @@ from press.runner import Ansible, StepHandler
 
 if TYPE_CHECKING:
 	from press.press.doctype.cluster.cluster import Cluster
+	from press.press.doctype.virtual_machine.virtual_machine import VirtualMachine
 
 
 class IPRemovalLog(Document, StepHandler):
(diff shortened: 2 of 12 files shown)


press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -1632,11 +1644,42 @@ def enable_source_dest_check(self):
...
 		)
 
+	def disassociate_hetzner_public_ip(self):
+		client = self.client()
+		server_instance = self.get_hetzner_server_instance(fetch_data=True)
+
+		should_power_on = server_instance.status == "running"
+
+		if should_power_on:
+			client.servers.power_off(server_instance).wait_until_finished(HETZNER_ACTION_RETRIES)
```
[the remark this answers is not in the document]
20vikash: Each server DOCTYPE will have its own VM. No two app servers will have the same underlying VM.

press/playbooks/roles/nat_iptables/tasks/main.yml
```diff
@@ -37,6 +43,8 @@
   when:
     - cloud_provider == "Hetzner"
     - network_gateway is defined
+    - should_setup_nat_gateway is defined
```
the-bokya: Why is this a separate condition? You could reuse nat_gateway_ip used for AWS and Frappe Compute.
```
    - nat_gateway_ip is defined
    - nat_gateway_ip
    - cloud_provider in ["AWS EC2", "Frappe Compute"]
```

press/press/doctype/ip_removal_log/ip_removal_log.py
```diff
@@ -129,13 +133,14 @@ def remove_ip_and_add_nat_conf(self, step):
 				port=doc._ssh_port(),
 				variables={
 					"nat_gateway_ip": frappe.cache.get_value(f"{self.name}:{self.nat_server}"),
-					"is_frappe_compute": doc.provider == "Frappe Compute",
+					"is_frappe_compute": True if doc.provider == "Frappe Compute" else None,
```
the-bokya, line 136: Inconsequential? I mean the condition is already handled if it is False.

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -1632,11 +1644,42 @@ def enable_source_dest_check(self):
...
+
+			if server_instance.public_net.primary_ipv6:
+				client.primary_ips.unassign(server_instance.public_net.primary_ipv6).wait_until_finished(
+					HETZNER_ACTION_RETRIES
+				)
+
+		if should_power_on:
+			client.servers.power_on(server_instance).wait_until_finished(HETZNER_ACTION_RETRIES)
+
+		time.sleep(25)  # Wait for sshd to come back before using ansible
```
the-bokya: Is this usually the case? Can take way more or less than 25 secs.
Can't you use Ansible ping in a loop maybe? 
ref: https://github.com/frappe/press/blob/develop/press/playbooks/ping.yml

the-bokya (changes requested):
Mostly LGTM. A few nitpicks.

press/press/doctype/ip_removal_log/ip_removal_log.py
20vikash: Ill check it out but for some reason, tasks for frappe compute started to execute until I changed it from False to None. Ill check it again

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -1632,11 +1644,100 @@ def enable_source_dest_check(self):
...
+				return
+
+			time.sleep(interval)
+
+		frappe.throw(
+			f"Timed out waiting for SSH on {self.name}. "
+			"Please check the server status and network connectivity."
+		)
+
+	def disassociate_hetzner_public_ip(self):
```
phot0n, line 1700: is this supposed to be a http request as well (as we do for aws and fcompute)?

then this wait loop should prolly be moved out (?)

ideally this wait loop should only be executed in ip removal log step - maybe have another method in ip removal controller for wait after executing the meat of this?

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -1632,11 +1644,100 @@ def enable_source_dest_check(self):
...
+				"-o",
+				"ConnectTimeout=5",
+				"-J",
+				f"{server.bastion_host.ssh_user}@{server.bastion_host.ip}:{server.bastion_host.ssh_port}",
+				f"root@{server.private_ip}",
+				"true",
+			],
+			stdout=subprocess.DEVNULL,
+			stderr=subprocess.DEVNULL,
+		)
```
phot0n, line 1666: we already have a ping.yml i guess (@the-bokya also mentioned the same i guess)

we use it in multiple places - if you dont wanna create ansible plays then its another thing

ex: https://github.com/frappe/press/blob/develop/press/press/doctype/press_job/jobs/create_server.py#L90

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -1632,11 +1644,100 @@ def enable_source_dest_check(self):
...
+			if server_instance.public_net:
+				if server_instance.public_net.primary_ipv4:
+					client.primary_ips.unassign(server_instance.public_net.primary_ipv4).wait_until_finished(
+						HETZNER_ACTION_RETRIES
+					)
+
+				if server_instance.public_net.primary_ipv6:
+					client.primary_ips.unassign(server_instance.public_net.primary_ipv6).wait_until_finished(
+						HETZNER_ACTION_RETRIES
+					)
```
phot0n, line 1723: we dont use ipv6 :P

press/press/doctype/virtual_machine/virtual_machine.py
phot0n: another question: do we need to wait for the machine to power off as well? before we can detach the ip?

phot0n:
@mergifyio backport master

PR #6801, opened 2026-06-25
fix(ui): Various minor ui issues

siduck:
- Removed blue colored links in sitemigration page 

### Before
image

### After
image


siduck:
@mergify backport master

PR #6804, opened 2026-06-25
feat(ui): Include changes column in site update dialog

siduck:
## Before

There was no changes column  

image

## After

image


siduck:
@mergify backport master

PR #6806, opened 2026-06-25
feat(server): Warn users before downgrading server plan

### LINKED ISSUE #6118: [Enhancement]: Warn users when they try to downgrade server plan
### Affected Area

None

### Problem Statement

Ref https://support.frappe.io/helpdesk/tickets/65663?view=VIEW-HD+Ticket-1232 

When users downgrade their server, it can lead to failed or slower operations due to reduced resources.
We should display a warning whenever they attempt to downgrade, regardless of the reason.

### Proposed Solution

_No response_

Shariqk3:
### Problem

Closes #6118

Users can downgrade a server's plan with a single click. Moving to a plan with fewer resources (vCPU, memory, or disk) can cause sites and background jobs on that server to run slower or fail under load, but there is currently no warning before the change is applied.

### Solution

In the server **Change Plan** dialog (`ServerPlansDialog.vue`), detect when the selected plan has fewer resources than the current plan and show a confirmation dialog before submitting:

- Added an `isDowngrade` computed that compares the selected plan's `vcpu` / `memory` / `disk` against the server's `current_plan`.
- `changePlan()` now routes downgrades through a red-themed `confirmDialog` warning; the plan change is only submitted after explicit confirmation.
- Upgrades and same-tier changes are unaffected — they submit directly with no extra prompt.

Reuses the existing `confirmDialog` helper and styling for consistency with other destructive confirmations (e.g. site deactivation).

### How to test

1. Open a server's **Change Plan** dialog.
2. Select a plan smaller than the current one → the "Downgrade Server Plan" warning appears and only proceeds on confirmation.
3. Select a larger plan → submits with no warning.

### DIFF
--- a/dashboard/src/components/server/ServerPlansDialog.vue
+++ b/dashboard/src/components/server/ServerPlansDialog.vue
@@ -243,8 +243,12 @@ export default {
 					message: `
 						Are you sure you want to downgrade this server's plan from
 						<b>${fromPlan}</b> to <b>${toPlan}</b>?<br><br>
-						The new plan has fewer resources than the current one. Sites and
+						The new plan has fewer resources than the current one, so sites and
 						background jobs on this server may run slower or fail under load.
+						If you're not sure the smaller plan can handle this server's
+						workload, reach out at
+						<a href="https://support.frappe.io" target="_blank" class="underline">support.frappe.io</a>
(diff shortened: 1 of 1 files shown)


Shariqk3:
image

PR #6810, opened 2026-06-25
fix(site): Disable subscription when site creation fails

### LINKED ISSUE #6110: Don't create Subscription if Site creation fails

Shariqk3:
### Problem

Closes #6110

When site creation fails, the site is marked **Broken** but the **Subscription** created during `Site.after_insert` stays enabled — so the user keeps getting billed for a site that never came up.

### Root cause

`process_new_site_job_update` sets the site status with `frappe.db.set_value(...)`, which bypasses the document's `on_update` hook. That hook (`update_subscription`) is what normally disables the subscription for `Broken`/`Archived`/`Suspended` sites and re-enables it for `Active` ones. Because it never runs in this path, the subscription state is never updated on create/fail/recover.

### Fix

In `process_new_site_job_update`, manage the subscription explicitly (reusing the existing `Site` methods), mirroring what `on_update` would have done:

- **Failure branch** → `disable_subscription()` so the user isn't billed for a broken site.
- **Success branch** → `enable_subscription()` so a site that **recovers** from a failed creation (a `New Site` retry or a restore — the only forward paths, since `check_allowed_actions` blocks `activate` on a creation-failed site) starts billing again. Idempotent for sites whose subscription was never disabled.

Other flows (manual activate/suspend, etc.) already go through `on_update` and are unaffected.

### Tests

- `test_subscription_is_disabled_when_site_creation_fails` — failed `New Site` job → site `Broken` **and** subscription disabled.
- `test_subscription_is_reenabled_when_site_recovers_to_active` — a disabled/Broken site whose creation jobs then succeed → site `Active` **and** subscription re-enabled.

Full `test_site` module passes (36 tests); `ruff check` / `ruff format` clean.

### DIFF
--- a/press/press/doctype/site/site.py
+++ b/press/press/doctype/site/site.py
@@ -4502,6 +4502,11 @@ def process_new_site_job_update(job):  # noqa: C901
 
 		site.sync_apps()  # Sync apps for this site as well to reflect dependant apps
 		marketplace_app_hook(site=site, op="install")
+		# Status is set via db.set_value below, bypassing on_update ->
+		# update_subscription. Re-enable the subscription explicitly so a site
+		# that recovers from a failed creation (retry / restore) starts billing
+		# again — the counterpart to disabling it on failure (#6110).
+		site.enable_subscription()
 	elif "Failure" in (first, second) or "Delivery Failure" in (first, second):
 		updated_status = "Broken"
 		frappe.db.set_value("Site", job.site, "creation_failed", frappe.utils.now())
--- a/press/press/doctype/site/site.py
+++ b/press/press/doctype/site/site.py
@@ -4505,6 +4505,10 @@ def process_new_site_job_update(job):  # noqa: C901
 	elif "Failure" in (first, second) or "Delivery Failure" in (first, second):
 		updated_status = "Broken"
 		frappe.db.set_value("Site", job.site, "creation_failed", frappe.utils.now())
+		# Status is set via db.set_value below, which bypasses on_update ->
+		# update_subscription. Disable the subscription explicitly so the user
+		# isn't billed for a site that never came up (#6110).
+		Site("Site", job.site).disable_subscription()
 	elif "Running" in (first, second):
 		updated_status = "Installing"
 	else:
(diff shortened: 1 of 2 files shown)


regdocs:
@shadrak98 please review

shadrak98:
@regdocs @Shariqk3 this looks fine but make sure the subscription is enabled again once the site is active from broken state. Can't find the flow in this PR. Assuming this is handled already if not please add.

Shariqk3:
> @regdocs @Shariqk3 this looks fine but make sure the subscription is enabled again once the site is active from broken state. Can't find the flow in this PR. Assuming this is handled already if not please add.

Can you check once again now ?

shadrak98:
Looks fine. One cosmetic change can be you can directly use `site.update_subscription` in both places as it is updates according to the site status. https://github.com/frappe/press/blob/develop/press/press/doctype/site/site.py#L2552

@Shariqk3 @regdocs

press/press/doctype/site/site.py
```diff
@@ -4503,9 +4503,18 @@ def process_new_site_job_update(job):  # noqa: C901
 
 		site.sync_apps()  # Sync apps for this site as well to reflect dependant apps
 		marketplace_app_hook(site=site, op="install")
+		# Status is set via db.set_value below, bypassing on_update ->
+		# update_subscription. Re-enable the subscription explicitly so a site
+		# that recovers from a failed creation (retry / restore) starts billing
+		# again — the counterpart to disabling it on failure (#6110).
+		site.enable_subscription()
```
shadrak98, line 4510: we can use `site.update_subscription` as well

press/press/doctype/site/site.py
```diff
@@ -4503,9 +4503,18 @@ def process_new_site_job_update(job):  # noqa: C901
...
 	elif "Failure" in (first, second) or "Delivery Failure" in (first, second):
 		updated_status = "Broken"
 		frappe.db.set_value("Site", job.site, "creation_failed", frappe.utils.now())
+		# Status is set via db.set_value below, which bypasses on_update ->
+		# update_subscription. Disable the subscription explicitly so the user
+		# isn't billed for a site that never came up (#6110).
+		Site("Site", job.site).disable_subscription()
```
shadrak98, line 4517: Here as well we can use `site.update_subscription`

shadrak98 (approved):

Shariqk3:
@shadrak98 If it's safe to merge, can you please merge it?

shadrak98:
LGTM!

PR #6816, opened 2026-06-26
fix(nat): enable dhcp and ignore its routes in nat config

phot0n:
for trying in frankfurt cluster

according to gemini:

In AWS, if two instances are in the same subnet, they receive the same default gateway (the AWS .1 address) via DHCP

`use-routes: false`: By explicitly defining this, you instruct systemd-networkd (or NetworkManager) to bind the private IP allocated by AWS DHCP but drop the standard AWS .1 gateway route before it hit the kernel's routing table


phot0n:
tried in a recent incident - worked

phot0n:
@mergifyio backport master

PR #6818, opened 2026-06-26
fix(virtual-machine): Give hetzner more default wait time for ssh

20vikash:
- Sometimes on Hetzner VMs, `systemd-networkd-wait-online.service` waits for its full 2-minute timeout before the network becomes available, delaying SSH readiness.


20vikash:
@mergifyio backport master

PR #6829, opened 2026-06-29
feat(site): Add migration shortcuts to the Actions tab

Shariqk3:
### What

Adds a **Migrations** card at the top of a site's **Actions** tab with a shortcut for each migration type:

- **In-Place Migrate Site** → *Migrate Site*
- **Move to a Different Server / Bench** → *Move Site*
- **Move to a Different Region** → *Move Site*

Clicking a shortcut routes the user to the **Migrations** tab and opens the migration dialog with the matching migration type **preselected** in the "Select Migration Type" dropdown.

### Why

The bench/server/region migration options were moved to the Migrations tab and are easy to miss from Actions (the Actions tab only shows a "moved to Migrations Tab" banner). These shortcuts make the common migrations discoverable and one click away, landing the user directly on the right choice.

### How

- New `migrationShortcuts` list in `SiteActions.vue`; each entry's `action` matches a key returned by `Site.get_migration_options`.
- `openMigration(action)` pushes to `Site Detail Migrations` and renders the existing `SiteMigration.vue` dialog with `defaultAction`.
- No backend change — reuses the dialog's existing `autoSelectMigrationOption()`, which already preselects from `defaultAction` / `?action=`.
- The card is hidden for `Archived` sites (matching the Migrations tab's own condition).

### Testing

Verified locally: the card renders in the Actions tab, and clicking a shortcut navigates to the Migrations tab and opens the dialog with the correct type preselected (e.g. "In-Place Migrate Site").

### DIFF
--- a/dashboard/src/components/SiteActions.vue
+++ b/dashboard/src/components/SiteActions.vue
@@ -22,12 +22,10 @@
 			</AlertBanner>
 			<!-- Migration shortcuts: deep-link to the Migrations tab with the dialog
 			     opened and the right migration type preselected. -->
-			<template
-				v-if="group.group == 'General Actions' && $site?.doc?.status !== 'Archived'"
-			>
+			<template v-if="$site?.doc?.status !== 'Archived'">
 				<div
 					class="py-3 first:pt-0 last:pb-0"
-					v-for="shortcut in migrationShortcuts"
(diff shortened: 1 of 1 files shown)


Shariqk3:
https://github.com/user-attachments/assets/575e9a57-b8c3-4df4-82e5-ca15d1c2d18f

regdocs (approved):

PR #6830, opened 2026-06-29
feat(nat): Extend NAT into OCI

20vikash:
- Previously, NAT servers, and removal of public IPs were only implemented in AWS, Hetzner, DO, and Frappe Compute
- This PR extends that into OCI.

**Note**

- This PR upgrades the OCI SDK version from `2.116.0` to `2.180.0` to make use of the new **`attach route table to VNIC`** feature.
- This version is compatible with `python 3.10`


press/press/doctype/server/server.py
```diff
@@ -3300,6 +3300,21 @@ def _setup_server(self):
...
+			nat_server = frappe.db.get_value(
+				"NAT Server",
+				{
+					"cluster": self.cluster,
+					"status": "Active",
+				},
+				"name",
+			)
+			if nat_server:
+				cluster.attach_route_table_to_instance_vnic_oci(vm, cluster.oci_nat_route_table_id)
```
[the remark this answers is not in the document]
20vikash: While creating NAT server, we are making sure that we add NAT route table in OCI. So, the field cannot be empty.
For IP removal, we are making sure that there is an active NAT server in the cluster before removing public IPs

press/press/doctype/database_server/database_server.py
```diff
@@ -821,7 +823,23 @@ def validate_server_id(self):
...
+			nat_server = frappe.db.get_value(
+				"NAT Server",
+				{
+					"cluster": self.cluster,
+					"status": "Active",
+				},
+				"name",
+			)
+			if nat_server:
+				cluster.attach_route_table_to_instance_vnic_oci(vm, cluster.oci_nat_route_table_id)
```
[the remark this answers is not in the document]
20vikash: While creating NAT server, we are making sure that we add NAT route table in OCI. So, the field cannot be empty.
For IP removal, we are making sure that there is an active NAT server in the cluster before removing public IPs

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -1762,22 +1807,85 @@ def disassociate_hetzner_public_ip(self):
...
+				"cluster": self.cluster,
+				"status": "Active",
+			},
+			"name",
+		)
+
+		if not nat_server:
+			return False
+
+		cluster.attach_route_table_to_instance_vnic_oci(self, cluster.oci_nat_route_table_id)
```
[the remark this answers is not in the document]
20vikash: While creating NAT server, we are making sure that we add NAT route table in OCI. So, the field cannot be empty.
For IP removal, we are making sure that there is an active NAT server in the cluster before removing public IPs

press/press/doctype/cluster/cluster.py
```diff
@@ -967,6 +970,124 @@ def create_nat_security_group_hetzner(self):
...
+				f"Please verify the instance configuration."
+			)
+
+		try:
+			network_client.update_vnic(
+				vnic_attachments[0].vnic_id,
+				UpdateVnicDetails(
+					route_table_id=nat_route_table_id,
+				),
+			)
```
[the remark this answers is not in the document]
20vikash: While creating NAT server, we are making sure that we add NAT route table in OCI. So, the field cannot be empty.
For IP removal, we are making sure that there is an active NAT server in the cluster before removing public IPs

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -2459,6 +2571,9 @@ def get_security_groups(self):
 			if self.cloud_provider == "Hetzner":
 				cluster: Cluster = frappe.get_doc("Cluster", self.cluster)
 				groups.append(cluster.create_nat_security_group_hetzner().id)
+			elif self.cloud_provider == "OCI":
+				cluster: Cluster = frappe.get_doc("Cluster", self.cluster)
+				groups.append(cluster.create_nat_security_group_oci().id)
```
phot0n: we dont need to create security groups here ... they should ideally be created when the cluster is created

press/press/doctype/virtual_machine/virtual_machine.py
phot0n: 👀

press/press/doctype/cluster/cluster.py
```diff
@@ -1128,6 +1249,11 @@ def provision_on_oci(self):
 		self.proxy_security_group_id = proxy_security_group.id
 
 		time.sleep(1)
+
+		self.create_nat_security_group_oci()
+
+		time.sleep(1)
```
phot0n, line 1263: ?

press/press/doctype/nat_server/nat_server.py
```diff
@@ -68,6 +74,25 @@ def setup_server(self):
 		frappe.enqueue_doc(self.doctype, self.name, "_setup_server", queue="long", timeout=2400)
 
 	def _setup_server(self):
+		if self.provider == "OCI":
+			vm: VirtualMachine = frappe.get_doc("Virtual Machine", self.virtual_machine)
+			cluster: Cluster = frappe.get_doc("Cluster", self.cluster)
+
+			vm.disable_source_dest_check()
```
phot0n: this can be done when we provision the vm

https://github.com/frappe/press/blob/aaf56fea920dd7420451787023bfcceea9f8a740/press/press/doctype/virtual_machine/virtual_machine.py#L739-L740

press/press/doctype/server/server.py
```diff
@@ -3300,6 +3300,21 @@ def _setup_server(self):
...
+			else:
+				frappe.throw("Failed to create server. Please create a NAT server in the cluster first")
```
phot0n: this can also be moved to when the vm is provisioned?

pyproject.toml
```diff
@@ -30,7 +30,7 @@ dependencies = [
     "dnspython==1.16.0",
     "docker==6.1.2",
     "dockerfile==3.2.0",
-    "oci==2.116.0",
+    "oci==2.180.0",
```
phot0n, line 33: doing this comes with risks - are we sure we're not introducing any breaking changes?

press/press/doctype/virtual_machine/virtual_machine.py
phot0n: we just need to append them here to the instance

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -1762,22 +1807,85 @@ def disassociate_hetzner_public_ip(self):
...
+		)
+
+		if not nat_server:
+			return False
+
+		cluster.attach_route_table_to_instance_vnic_oci(self, cluster.oci_nat_route_table_id)
+		network_client.delete_public_ip(public_ip.id)
+
+		frappe.flags.force_update_dns = True
+		self.sync()
```
phot0n: we do these 2 things in the method below i think

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -1762,22 +1807,85 @@ def disassociate_hetzner_public_ip(self):
...
+		try:
+			public_ip = network_client.get_public_ip_by_private_ip_id(
+				GetPublicIpByPrivateIpIdDetails(
+					private_ip_id=private_ips[0].id,
+				)
+			).data
+		except ServiceError as e:
+			if e.status == 404:
+				# No public IP attached
+				return None
```
phot0n: let the error be raised here as well?

press/press/doctype/nat_server/nat_server.py
20vikash: The reason I didnt do anything while provisioninig the VM was because, the VNIC object needs to be created in OCI for me to attach route tables, disable src check, etc, etc.. So, I moved that into nat_server code. The provisioning itself is async, and it failed for me in local(as it should)

press/press/doctype/server/server.py
20vikash: The reason I didnt do anything while provisioninig the VM was because, the VNIC object needs to be created in OCI for me to attach route tables, disable src check, etc, etc.. So, I moved that into nat_server code. The provisioning itself is async, and it failed for me in local(as it should)

pyproject.toml
20vikash: The version upgrade didnt bring any issues. its compatible with the older versions and is supported by python 3.10

press/press/doctype/virtual_machine/virtual_machine.py
20vikash: The reason I created nat secutiy group(not really, because that function will return if already exists, or will create if it dosent exist) because while migrating the old clusters, we need to create the NAT security group while creating the NAT server. It's also created in the cluster creation time(for future new clusters)

press/press/doctype/virtual_machine/virtual_machine.py
20vikash: So, VNIC is a first class citizen/object in OCI. So, I had to make sure there is a running NAT server in the cluster before attaching the route table to the VNIC object of the VM.

press/press/doctype/cluster/cluster.py
20vikash: This pattern is followed all over provision_oci(), i think to wait for the resource being useful inside OCI before using.

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -1762,22 +1807,77 @@ def disassociate_hetzner_public_ip(self):
...
+		)
+
+		if not nat_server:
+			frappe.throw(
+				"No active NAT servers found in the cluster. "
+				"Please create a NAT server before removing public IP"
+			)
+
+		cluster.attach_route_table_to_instance_vnic_oci(self, cluster.oci_nat_route_table_id)
+		network_client.delete_public_ip(public_ip.id)
```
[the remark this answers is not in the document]
20vikash: If a public VM got NAT route table attached in its VNIC, and failed in mid state without removing public IP, the VM itself would not lose network/internet connectivity(As every packets are routed to NAT VM, then to the internet).
And also, if I do a try/except block here, and revert the VNIC to have default route table, what would happen if that fails? infinite try/except's?
Also this function is idempotent. If something goes wrong in the mid, running this again would not break stuff and would end up in a consistent state.

press/press/doctype/virtual_machine/virtual_machine.py
```diff
@@ -279,6 +281,17 @@ def on_trash(self):
...
 
+		if (
+			self.has_value_changed("status")
+			and self.status == "Running"
+			and self.cloud_provider == "OCI"
+			and self.series == "nat"
+		):
+			cluster: Cluster = frappe.get_doc("Cluster", self.cluster)
+
+			self.disable_source_dest_check()
```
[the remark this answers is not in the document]
20vikash: We provide `self.private_ip` inside the VM provisioning code. We don't `list_private_ips` inside the route table creation function.
Even while disassociating the public IP for OCI, we are sure NAT VM exists, and route table exists in `running` state.

20vikash:
@mergifyio backport master

PR #6838, opened 2026-06-29
feat(site): Gate dedicated plan warranty on app + database server cost

regdocs:
The product warranty plan gate (`_is_plan_allowed_on_server`) decided whether a free, dedicated-server site plan could be used by comparing the site plan's `minimum_server_price_usd` against the *app server's* plan price alone. A dedicated server is really an app server plus a paired database server, so the database server's cost was silently ignored — a team running a small app server with a large database server could be blocked from a plan whose minimum they actually clear.

Compare against the combined cost instead. A new `get_dedicated_server_price` helper sums the app server plan price and the paired database server plan price; when a server has no database server or that server has no plan, the database portion is 0, so app-only servers behave exactly as before.

`_get_team_dedicated_server_info` now returns this combined price as `price_usd` (it previously joined `plan.price_usd`, app-only). That value feeds `serverPriceUsd` in NewSite.vue and the client-side `serverPlanPrice >= minimum_server_price_usd` check in SitePlansCards.vue, so the dashboard and the backend gate stay in agreement.

Note: existing `minimum_server_price_usd` values were likely calibrated against the app-server-only price; since the comparison base is now larger, those thresholds may need raising to preserve the same effective cutoff.

### DIFF
--- a/press/api/site.py
+++ b/press/api/site.py
@@ -331,10 +331,21 @@ def _is_plan_allowed_on_server(server: str, new_site_plan: dict) -> bool:
 		return False
 	if not new_site_plan.get("restrict_based_on_dedicated_server_plan", 0):
 		return True
-	app_server_plan = frappe.db.get_value("Server", server, "plan")
 	min_price = new_site_plan.get("minimum_server_price_usd", 0)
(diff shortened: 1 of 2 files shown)


regdocs:
@mergifyio backport master

PR #6839, opened 2026-06-30
feat(ip-removal-log): extend log to update nat config as well

phot0n:


phot0n:
@mergifyio backport master

PR #6858, opened 2026-06-30
fix(team): Do not pass `Admin` role to the backend

ssiyad:


ssiyad:
@mergify backport master

PR #6868, opened 2026-07-01
fix(site-backup): Derive timeout from previous backup duration

huzan-kazi:
## Problem

The `backup_timeout` field on Site (static default: 18000 s / 5 h) caused backup jobs to time out on larger sites whose backups took longer than that fixed ceiling. The minimum system-level agent job timeout is 4 hours, so any site that grew past ~2.5 h of backup time was at risk of hitting the wall.

## Solution

- **Removes** the `backup_timeout` field from the Site doctype entirely.
- **Adds** `get_dynamic_backup_timeout(site, with_files)` in `site_backup.py` that:
  - Queries the most recent *successful* logical backup for the site of the same type (`with_files` matched, so DB-only and full backups are estimated separately).
  - Computes `previous_duration × 1.5` — a 50% buffer to absorb normal data growth between runs.
  - Floors the result at 4 hours so new sites (no prior backup) always get a safe minimum.
- **Updates** `Agent.backup_site()` in `agent.py` to call `get_dynamic_backup_timeout` instead of reading the now-removed static field.

## Test plan

- [ ] Trigger a backup on a site that has a successful previous backup — confirm `agent_job_timeout` in the agent request is `previous_duration × 1.5` (minimum 14400 s).
- [ ] Trigger a backup on a brand-new site with no prior backup — confirm `agent_job_timeout` is 14400 s (4 h).
- [ ] Trigger a `with_files=True` backup — confirm it uses the last `with_files` backup's duration, not a DB-only one.
- [ ] Confirm no references to `backup_timeout` remain in Site records or API responses.

### DIFF
--- a/press/agent.py
+++ b/press/agent.py
@@ -568,9 +568,11 @@ class Agent:
 		)
 
 	def backup_site(self, site, site_backup: SiteBackup):
+		from press.press.doctype.site_backup.site_backup import get_dynamic_backup_timeout
+
 		data = {
 			"with_files": site_backup.with_files,
-			"agent_job_timeout": site.backup_timeout,
+			"agent_job_timeout": get_dynamic_backup_timeout(site, site_backup.with_files),
 		}
 		if site_backup.offsite:
 			backups_path = os.path.join(site.name, str(date.today()))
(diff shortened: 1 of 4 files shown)


huzan-kazi:
closing because of unclean history

PR #6881, opened 2026-07-02
fix(ui): Use named routes for bench list rows

siduck:


siduck:
@mergify backport master

PR #6883, opened 2026-07-02
fix(ui): autologout in benches list caused by using frappe.client.*

siduck:


siduck:
@mergify backport master

PR #6887, opened 2026-07-02
fix(team): Select correct role and other ui changes

ssiyad:


ssiyad:
@mergify backport master

PR #6888, opened 2026-07-02
fix(ui): Add backoff to release pipeline poll restart

siduck:


siduck:
@mergify backport master

PR #6905, opened 2026-07-03
fix(product-trial): Only hand out standby sites on the latest bench

regdocs:
## Problem

Users claiming trial (standby) sites were getting sites running **out-of-date app builds**.

Standby sites are created on whatever bench is latest at creation time and then rely on the generic `schedule_updates` cron to migrate them forward as new deploys happen. That migration is:

- **asynchronous** — runs every 15 min and the actual bench move (backup → restore) takes minutes;
- **throttled** — capped per server by `auto_update_queue_size`, so the pool competes with real customer sites for update slots.

Nothing verified a standby site had caught up before it was claimed, and `get_preferred_site` deliberately picks the **oldest** site (`order_by=... creation asc`) — precisely the one most likely to predate the latest deploy.

## Fix

Add a **claim-time guard**: `get_standby_site` now only considers sites sitting on the newest `Active` bench for their release group.

- Computed **per server** (`get_latest_benches`), since standby sites are spread across servers/clusters and each server has its own current bench — a single global "latest" would wrongly reject sites on other servers' current benches.
- Reuses the codebase's existing "latest bench" definition: `{group, status: "Active"}` ordered by `creation desc` (same as `site_update.py` destination selection and `site.py` bench placement).
- Sites still waiting on the async updater are simply skipped. When no fresh standby site is available, `get_standby_site` returns `None` and the caller already falls back to creating a fresh on-demand site — so the worst case is a normal site creation instead of a stale hand-off.

## Tests

`test_product_trial.py` — using real Bench/Site records (old + new bench on the same server):

- `get_latest_benches` returns only the newest bench per server
- a standby site on the older bench is not handed out; the fresh one is
- when only stale sites exist, `get_standby_site` returns `None`

## Out of scope

This does **not** backfill apps *added* to a product after a standby site was created — a site update migrates benches but never installs new apps. If that's part of the report, it needs a separate fix (e.g. also filtering the pool on app-set match).

### DIFF
--- a/press/saas/doctype/product_trial/product_trial.py
+++ b/press/saas/doctype/product_trial/product_trial.py
@@ -297,11 +297,35 @@ class ProductTrial(Document):
 		sites_without_incident = [site["name"] for site in sites_without_incident]
 		return sites_without_incident[0] if sites_without_incident else sites[0]
 
+	def get_latest_benches(self) -> list[str]:
+		"""Newest Active bench per server for this product's release group.
+
+		A standby site is only handed out if it sits on one of these benches,
+		so a user never receives a site that predates the latest deploy. Sites
+		still waiting on the async site-update job to migrate them forward are
+		excluded until they land on the current bench.
+		"""
+		benches = frappe.get_all(
+			"Bench",
+			filters={"group": self.release_group, "status": "Active"},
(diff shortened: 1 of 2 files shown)


regdocs:
@mergifyio backport master

PR #6906, opened 2026-07-03
fix(ui): Add server plan tooltip 

siduck:


siduck:
@mergify backport master

PR #6917, opened 2026-07-07
fix: Remove unnecessary validation in Release Group doctype

AyshaHakeem:


AyshaHakeem:
@mergify backport master

PR #6919, opened 2026-07-07
feat(app-source): Sync app source versions from repo on branch change

siduck:


siduck:
@mergify backport master

PR #6924, opened 2026-07-09
fix(site): use host name if present when trying to notify users of incoming site archival

phot0n:


phot0n:
@mergifyio backport master

PR #6925, opened 2026-07-09
fix(ui): properly match githuburl in add apps dialog

siduck:
## Description 

Make the regex matcher have github.com *  urls properly without prepending https://

image


siduck:
@mergify backport master

PR #6929, opened 2026-07-10
fix(ui): Remove duplicate primary btn in CommunicationInfoDialog

siduck:
## Before

image

## After

image


siduck:
@mergify backport master

PR #6937, opened 2026-07-13
fix: Sync apps list after a successful site update

AyshaHakeem:


AyshaHakeem:
@mergify backport master

PR #6939, opened 2026-07-13
fix(ui): Show proper versions in bench list filters

siduck:
use linkcontrol, dont hardcode options


siduck:
@mergify backport master

PR #6958, opened 2026-07-14
feat(site): Add support for freemium plan

tanmoysrt:
We want to provide users lifetime free subscriptions sometime.
So, adding support for that. In product trial, if we set trial days to -1, it will avoid setting any trial date for the site.

The plan will be still trial plan, we have handled logic for `is_trial_plan` at many places, so not introducing a new flag.


tanmoysrt:
@greptile review

PR #6959, opened 2026-07-14
Revert "fix(server): remove and add nat config back if applicable during resize server job"

phot0n:
Reverts frappe/press#6675


phot0n:
@mergifyio backport master

PR #6960, opened 2026-07-14
feat(waf): Implement WAF doctypes to configure WAF in agent

20vikash:


tanmoysrt:
@20vikash @ssiyad Can we do this in V2 instead of pushing new things on V1 (We should push only important patch fixes on V1, as none is available much for handling regressions due to new things) ?

ssiyad:
> Can we do this in V2 instead of pushing new things on V1 (We should push only important patch fixes on V1, as none is available much for handling regressions due to new things) ?

Agree. I was expecting this to be a small plugin/change. But it looks like there is a lot going on here. @tanmoysrt 

Let's keep this as a POC and try to land a better WAF in v2. @20vikash

PR #6963, opened 2026-07-14
perf(analytics): Split advanced site analytics into per-chart requests

regdocs:
## Problem

The site insights **Advanced Analytics** section (`sites/<name>/insights/analytics`) loads slowly. `press.api.analytics.get_advanced_analytics` runs ~11 Elasticsearch queries **sequentially** inside a single request handler, so the section blocks on the **sum** of every query's latency — the whole panel only renders once the slowest-accumulated chain finishes.

## Change

Split `get_advanced_analytics` into **eight whitelisted endpoints**, one per chart:

- `get_request_count_by_path`
- `get_request_duration_by_path` (bundles its slow-path breakdowns)
- `get_average_request_duration_by_path`
- `get_request_count_by_ip`
- `get_background_job_count_by_method`
- `get_background_job_duration_by_method` (bundles its report breakdowns)
- `get_average_background_job_duration_by_method`
- `get_background_job_usage` (job count + CPU time)

`SiteAnalytics.vue` now fires these as **independent parallel resources**, each landing on its own worker. Wall-clock time drops from *sum of all queries* to roughly the *slowest single chart*, and each card shows/clears its own loading spinner as its data arrives.

## Notes

- The underlying group-by helpers are already `@redis_cache`-wrapped, so the split adds **no duplicate computation**.
- The slow-path / slow-job breakdowns genuinely depend on the top paths of their parent duration chart, so they stay **bundled inside** the request / background-job duration endpoints rather than becoming a dependent second wave of requests.
- Only consumer of the old endpoint was `SiteAnalytics.vue`; no other callers.

### DIFF
--- a/dashboard/src/components/site/SiteAnalytics.vue
+++ b/dashboard/src/components/site/SiteAnalytics.vue
@@ -66,9 +66,7 @@
 
 	<div class="space-y-4 p-5">
 		<ErrorMessage
-			:message="
-				$resources.analytics.error || $resources.advancedAnalytics.error
-			"
+			:message="$resources.analytics.error || advancedAnalyticsError"
 		/>
 
 		<div class="grid grid-cols-1 gap-5 sm:grid-cols-2">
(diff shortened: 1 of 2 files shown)


regdocs:
@mergifyio backport master

PR #6972, opened 2026-07-15
fix(wazuh): Tolerate a missing unit

ssiyad:


ssiyad:
@mergifyio backport master

PR #6975, opened 2026-07-15
fix(sync): Check the public ip address field instead of check box

20vikash:


phot0n:
@mergifyio backport master

PR #6981, opened 2026-07-16
fix: Fix conflict, `plans` section fields got added twice

AyshaHakeem:
Conflict in press_settings.json where in fields under the`plans` section got added twice.
Likely happened while resolving conflicts when taking release


AyshaHakeem:
@mergify backport master

PR #6983, opened 2026-07-16
fix(bench): Stop fatal updates pinning benches they cannot help

balamurali27:
## Problem

Two empty benches, the source and destination of one recent Fatal Site Update, cannot be dropped:

> There was a recent **site update which has failed**. Due to the same reason, bench cannot be archived.

The `fatal_site_updates` guard in `Bench.check_ongoing_site_updates()` holds a bench for `EMPTY_BENCH_COURTESY_DAYS` (3) whenever a recent Fatal update names it as source or destination, so the source survives as a rollback target for a site left broken on the destination. Two holes let it pin benches that cannot serve that purpose:

1. **It keys on the site's status, not on where the site lives.** #6927 skips `Archived` sites, covering the site being gone but not the site being *elsewhere*. A site fixed by moving forward to a later deploy instead of rolling back is neither `Archived` nor on either bench — nothing left to roll back to, benches pinned anyway.
2. **`cause_of_failure_is_resolved` was never read.** It is set from the desk button and auto-set in `handle_failure()` for agent-update and incident failures. Press could conclude a failure was its own fault, record it, and still pin the benches.

In the reported case either hole alone was enough to keep both benches.

## Fix

Join `Site` on the update's benches as well as its name, so only a site still parked on the source or destination holds up the archive, and skip updates whose cause is already resolved.

- Keeping `site_updates.site == sites.name` in the join matters: without it an unrelated site on the destination blocks the source.
- The flag is compared to `0` rather than negated with `not`: a pypika `Field` is always truthy, so `not field` collapses to a constant `False` at build time and disables the whole guard. That sank a9c9d1c24, reverted in 7295a081d.
- The `!= "Archived"` condition from #6927 stays — archiving a site does not clear `Site.bench`.
- `ongoing_site_updates` is untouched.

## Testing

Two new tests, one per hole, each failing on `develop`:

- `test_fatal_site_update_doesnt_block_archival_when_site_left_both_benches`
- `test_fatal_site_update_doesnt_block_archival_once_cause_of_failure_is_resolved` — site left *on* the destination, so only the flag can unblock it.

The existing `test_if_fatal_site_update_with_source_bench_blocks_archival` still passes: site on the destination with the cause unresolved, source stays blocked. Full `test_bench` (29) and `test_site_update` (11) green.

### DIFF
--- a/press/press/doctype/bench/bench.py
+++ b/press/press/doctype/bench/bench.py
@@ -1126,13 +1126,20 @@ class Bench(Document):
 		fatal_site_updates = (
 			frappe.qb.from_(site_updates)
 			.join(sites)
-			.on(site_updates.site == sites.name)
+			.on(
+				(site_updates.site == sites.name)
+				& (
+					(sites.bench == site_updates.source_bench)
+					| (sites.bench == site_updates.destination_bench)
+				)
+			)
 			.select(site_updates.name)
 			.where((site_updates.source_bench == self.name) | (site_updates.destination_bench == self.name))
 			.where(
 				(site_updates.status == "Fatal")
 				& (site_updates.creation > frappe.utils.add_to_date(None, days=-EMPTY_BENCH_COURTESY_DAYS))
 				& (sites.status != "Archived")
+				& (site_updates.cause_of_failure_is_resolved == 0)
 			)
 			.limit(1)
 		).run()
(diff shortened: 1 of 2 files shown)


balamurali27:
@Mergifyio backport master

PR #6990, opened 2026-07-17
feat: Report public server pool health

AyshaHakeem:
- Add `press.press.doctype.server.server_monitoring.monitor_server_and_refresh_new_bench_and_site_server_pool`

- The hourly job does four things:
  - Finds active public primary servers that are eligible for site/bench creation
  - Fetches RAM, CPU, and swap metrics from Prometheus in bulk.
  - Chooses the best server per cluster for new benches and new sites.  
  - Reports health problems through Raven alerts and fallback incidents.

- It intentionally replaces the older `refresh_new_bench_and_site_server_pool` scheduler path so pool selection and health reporting use the same$

- A server is considered healthy when both are true:
  - Available memory ratio is at least 20%.
  - CPU idle ratio is at least 50%.

- The health issue list records:
  - `RAM utilization: X% > 80%`
  - `CPU utilization: X% > 50%`
  - `Swap usage: X GiB > threshold`

- Flag Updates
  - `use_for_new_benches` is enabled on selected bench servers.
  - `use_for_new_sites` is enabled on selected site servers.

- Fallback server
  - If a cluster has candidates but none are healthy, an incident is created but the job still selects one fallback server so scheduling does not$

### DIFF
--- a/press/hooks.py
+++ b/press/hooks.py
@@ -272,7 +272,7 @@ scheduler_events = {
 	],
 	"hourly_long": [
 		"press.press.doctype.release_group.release_group.prune_servers_without_sites",
-		"press.press.doctype.server.server.refresh_new_bench_and_site_server_pool",
+		"press.press.doctype.server.server_monitoring.monitor_server_and_refresh_new_bench_and_site_server_pool",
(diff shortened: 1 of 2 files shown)


AyshaHakeem:
@mergify backport master

PR #6994, opened 2026-07-17
fix(site-update): Block updates on sites with database over 100 GB

balamurali27:
## Problem

A logical backup of a very large database takes hours and often fails or times out midway through a site update, leaving the site in a bad state that needs manual recovery. Sites over ~100 GB are the ones that hurt.

## Fix

Block these updates in `SiteUpdate.validate` unless the update avoids the full dump, and point the customer at support so we can do the update by hand instead of failing halfway through.

Details worth noting:

- The size comes from the newest **Site Usage** record — the same source the dashboard shows on the site overview page (`Site.current_usage`). Site Usage sizes are in **MB, not bytes**: `SiteOverview.vue` renders them with `$format.bytes(v, 2, 2)`, where the third argument shifts the unit label up by two. Hence the threshold is `100 * 1024`. (Note `Site.current_database_usage` is *not* this number — it's a percentage of the plan limit.)
- **Physical and Logical Replication are both allowed through.** Neither takes a full dump — Logical Replication goes down the `create_logical_replication_backup_record` path, built on snapshots and replication. Only plain `Logical` does the slow dump this guards against.
- The check runs after the backup mode setters so it sees the final `backup_type`. `set_logical_replication_backup_mode_if_eligible` runs last and overwrites `backup_type` unconditionally, so checking any earlier would read a value that's about to change.
- It skips when `skipped_backups` is set — that's a deliberate operator override, not something to block.
- `Pull` updates are blocked too, not just `Migrate`. They also take a logical backup, which is the slow and failure-prone part.

## Tests

Three new tests using real Site Usage records (no mocks): 101 GB is blocked, 99 GB goes through, and a 101 GB site with a Logical Replication backup is allowed. Full `test_site_update` module passes (14 tests).

## Open question

`LARGE_DATABASE_SIZE_MB` is a hardcoded constant. If support needs to tune it per-site or turn it off without a deploy, it should be a Press Settings field instead — happy to change it.

### DIFF
--- a/press/press/doctype/site_update/site_update.py
+++ b/press/press/doctype/site_update/site_update.py
@@ -38,6 +38,9 @@ if TYPE_CHECKING:
 	)
 	from press.press.doctype.site.site import Site
 
+# Site Usage stores sizes in MB
+LARGE_DATABASE_SIZE_MB = 100 * 1024
+
 
 class SiteUpdate(Document):
 	# begin: auto-generated types
@@ -151,6 +154,7 @@ class SiteUpdate(Document):
 		self.validate_past_failed_updates()
 		self.set_physical_backup_mode_if_eligible()
 		self.set_logical_replication_backup_mode_if_eligible()
+		self.validate_backup_type_for_large_database()
 
 	def validate_destination_bench(self, differences):
 		if not self.destination_bench:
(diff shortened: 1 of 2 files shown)


balamurali27:
@Mergifyio backport master

PR #7000, opened 2026-07-20
fix(deploy): Show deploy creator in deploy list

BreadGenie:
instead of team as team is always the same


siduck:
@mergify backport master

PR #7002, opened 2026-07-20
fix(dns): Ignore failed-over proxies in wildcard skip check

balamurali27:
Follow-up to the [TLS incident (2026-07-13)](https://docs.frappe.io/cloud/recent-issues#tls-incident-across-multiple-regions-2026-07-13).

**Fix — `press/utils/dns.py`**
- Wildcard-covered site check counted failed-over proxies (`exclude_from_auto_selection`) as active.
- A live primary + failed-over old primary looked like "multiple proxies", forcing per-site records at the wrong proxy.
- Now counts only auto-selectable proxies.

**Audit — `press/scripts/audit_wildcard_dns.py`**
- Checks `*.<domain>` A record vs the default cluster's primary proxy for each enabled Root Domain.
- Prints only failed domains; `fix('x.frappe.cloud')` UPSERTs one at a time.

### DIFF
--- a/press/scripts/audit_wildcard_dns.py
+++ b/press/scripts/audit_wildcard_dns.py
@@ -59,7 +59,7 @@ failed = []
 for domain in frappe.get_all("Root Domain", {"enabled": 1}, pluck="name"):
 	try:
 		_, ip = expected_proxy_ip(domain)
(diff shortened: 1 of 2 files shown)


balamurali27:
@Mergifyio backport master

PR #7003, opened 2026-07-21
feat(proxy): Add playbook to update ProxySQL on proxies

balamurali27:
## Why

Older proxies run the ProxySQL **2.3.2** image, which fails to serve newly issued Let's Encrypt certificates. This adds an update path to move all proxies to **3.0.9**.

## What

- **`update_proxysql.yml` playbook + role** — mirrors the manual procedure already done by hand on a couple of proxies:
  1. stop the container
  2. archive rotated logs into `proxysql-old-logs/`
  3. back up the data dir to `proxysql-old/`
  4. `docker rm` the old container
  5. recreate from the new image, reusing the same volume and `proxysql.cnf`

  The backup (`creates:`) and log-archive steps are guarded so the play is re-runnable.

- **`ProxyServer.update_proxysql()`** — whitelisted method that enqueues the play on the `long` queue.

- **`PROXYSQL_VERSION` constant** in `proxy_server.py` is the single source of truth for the image tag. The playbook requires the variable to be passed and fails loud if it isn't.

## Rollout

```python
frappe.get_doc("Proxy Server", name).update_proxysql()
```

### DIFF
--- a/press/playbooks/roles/update_proxysql/tasks/main.yml
+++ b/press/playbooks/roles/update_proxysql/tasks/main.yml
@@ -1,0 +1,37 @@
+---
+# Updates the ProxySQL container to a new image version. Mirrors the manual
+# procedure: stop container, archive rotated logs, back up the data dir, then
+# recreate the container from the new image reusing the same volume and config.
+
+- name: Stop ProxySQL
+  become: yes
+  become_user: frappe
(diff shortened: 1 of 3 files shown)


balamurali27:
@Mergifyio backport master

PR #7005, opened 2026-07-21
feat(site): Allow more actions on inactive sites

balamurali27:
The `site_action` decorator gated several actions to `Active` only, but inactive and broken sites work fine for them. `get_actions` already exposed some (restore, reset) for inactive sites, so the button appeared then errored on the backend `site_action` check.

## Changes

- **reinstall, restore (×3), migrate** → add `Inactive` (already allowed `Broken`)
- **update_without_backup** → match its sibling `schedule_update` (`Active`, `Inactive`, `Suspended`, `Broken`)
- **delete_config** → add `Broken` to match `update_config`

## Notes

- `migrate` preserves the inactive state (`activate=False` + `status_before_update`). Restore/reinstall have no activate flag, so an inactive site comes back **active** after them — acceptable for now.
- **Version upgrade** and **site update** were already correct: both route through Site Update, which only blocks transitory/archived states and preserves `status_before_update`, and their UI conditions already include `Inactive`.

### DIFF
--- a/press/press/doctype/site/site.py
+++ b/press/press/doctype/site/site.py
@@ -1177,7 +1177,7 @@ class Site(Document, TagHelpers):
 		agent.new_upstream_file(server=self.server, site=self.name)
 
 	@dashboard_whitelist()
-	@site_action(["Active", "Broken"])
+	@site_action(["Active", "Broken", "Inactive"])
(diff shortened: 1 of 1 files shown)


balamurali27:
@Mergifyio backport master

PR #7008, opened 2026-07-21
fix(site): Improve app-mismatch error when moving a site

balamurali27:
## Why

When moving a site to another bench group, if the site has apps that aren't in the destination group, the error was hard to parse — app names were inline plain text and the "install / remove" steps were undirected.

## What

- **Bold** the offending app names (`frappe.bold`).
- Add a colon after "following apps:".
- Link **install** → [installing an app to a bench group](https://docs.frappe.io/cloud/installing-an-app#bench-group).
- Link **remove** → [uninstalling an app from a site](https://docs.frappe.io/cloud/how-to-uninstall-an-app-from-the-site).

Also documents in `AGENTS.md` how to find `docs.frappe.io/cloud` URLs via the wiki search API (guessed slugs 404), so future doc links point to real pages.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -39,3 +39,18 @@ npx playwright test tests-e2e/tests/dashboard/site-update-banner.test.ts --heade
 ```
 
 Requires a running bench (`bench start`) and `dashboard/tests-e2e/.env` with credentials.
+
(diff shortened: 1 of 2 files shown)


balamurali27:
@Mergifyio backport master

PR #7018, opened 2026-07-21
fix(telemetry): Fix signup identity and attribution in Pulse

nextchamp-saqib:
Pulse's cookieless anon id hashes the site into it, so signup events captured on the dashboard got a different id than the same visitor's browsing on the product website — the funnel between landing on the form and the account existing was detached from the session that led there. The forwarded `?aid=` is now kept in `localStorage` for an hour and sent as the guest's distinct id. Google signups never had an id to alias at all, since they skip `signup()`; it now rides the cached OAuth flow state that already carries `product`.

Also:
- `product` no longer falls back to `"fc"`, which read as a real cohort when it only meant "a plain Frappe Cloud signup".
- `signup_method` replaces `signup_via_email` / `signup_via_google`, which only fired on product signups and never on the OAuth redirect. It's read off the `oauth_signup` flag already on Account Request.
- Adds `signup_verified`. Captured from three pages because the main flow and the two product-trial flows land on different ones; OAuth counts as verified on arrival.

The alias itself is unchanged — still server-side at team creation, off the id on Account Request.


nextchamp-saqib:
CI has unrelated failure, there's an open PR for it: https://github.com/frappe/press/pull/7030

PR #7020, opened 2026-07-21
fix(login): Send already-logged-in product signup through quickstart

siduck:


siduck:
@mergify backport master

PR #7024, opened 2026-07-22
fix: Replace swap metrics with OOM for server health check

AyshaHakeem:


AyshaHakeem:
@mergify backport master

PR #7028, opened 2026-07-23
fix: Fix Raven channel name for alerting server health

AyshaHakeem:
Change channel name from "Frappe Cloud-server-alerts" to "frappe-cloud-server-alerts"


AyshaHakeem:
@mergify backport master

PR #7034, opened 2026-07-24
fix(ui): Reload bench apps page on deploy update

### LINKED ISSUE #7017: frontend: status state doesn't update until refresh
### Information about bug

The deploy has completed & its status is update to “Active”.

But the app still says “Not Deployed” which is stale & hasn’t been updated, upon refresh it updates.

Image

siduck:
closes #7017


siduck:
@mergify backport master

PR #7045, opened 2026-07-24
fix(app): Explain what makes a frappe version range invalid

balamurali27:
Follow-up to #7042 / #7044, where a range that reads fine got rejected with:

> Invalid version range: There is an inconsistency between the upper and lower bound major versions. Please ensure the version range specifies valid major versions.

That message names neither the range nor what to change, and the cause isn't visible in what the user wrote. `NpmSpec` pairs a prerelease bound with a companion clause on the *opposite* side — `<=17.0.0-dev` also yields `>=17.0.0`, `>=16.0.0-dev` also yields `<16.0.1` — so `get_lower_bound_major` sees an upper major (16) below a lower major (17) and throws.

**Now:**

> Invalid version range '>=16.0.0-dev <=17.0.0-dev': the upper bound's major version is below the lower bound's. Prerelease bounds cause this too — '<=17.0.0-dev' also implies '>=17.0.0'. Use stable versions for the bounds, e.g. '>=16.0.0-dev <17.0.0'. See [declaring app versions](https://docs.frappe.io/cloud/custom-apps/app-versioning/versioning) for how to set this up in pyproject.toml.

All three version errors in `map_frappe_version` now carry the docs link — an app with no `pyproject.toml` at all reaches these with nothing to go on.

Note this path only runs under `ease_versioning_constrains=True` (`app.py:243`).

Test: `test_prerelease_upper_bound_is_rejected_with_a_message_naming_the_range`. Full module passes (8 tests).

⚠️ The linked doc's own recommended example is `frappe = ">=16.0.0-dev,<=17.0.0-dev"` — the exact form that fails here. That page needs updating too.

Ref: https://github.com/npm/node-semver#prerelease-tags

### DIFF
--- a/press/press/doctype/app/app.py
+++ b/press/press/doctype/app/app.py
@@ -21,6 +21,12 @@ if typing.TYPE_CHECKING:
 	from press.press.doctype.app_source.app_source import AppSource
 
 
+VERSIONING_DOCS = (
+	"See <a href='https://docs.frappe.io/cloud/custom-apps/app-versioning/versioning'>"
+	"declaring app versions</a> for how to set this up in pyproject.toml."
+)
+
+
 class VersioningError(Exception): ...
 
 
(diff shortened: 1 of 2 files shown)


balamurali27:
Addressed in abf138297 — the suggestion is now derived from `min(lowers)` rather than hardcoded, so `>=14.0.0-dev <=15.0.0-dev` suggests `>=14.0.0-dev <15.0.0`. Also generalised the prerelease explanation to `'<=X.0.0-dev' also implies '>=X.0.0'`. Added `test_suggested_upper_bound_follows_the_range_that_was_rejected` to cover a non-16 major.

PR #7047, opened 2026-07-24
feat(support-agent): Add incident time and confirm noisy neighbours

balamurali27:
## What

**Incident Time.** The investigation takes an optional `incident_time` and looks 12 hours either side of it, instead of trailing 24 hours from now. Support tickets name a moment ("the site was slow around 3pm"), and a trailing window misses it. Empty means now. Exposed on the Site form's **Investigate** button and on the doctype.

`TimeWindow` owns the window and converts itself per backend — naive datetimes for DB filters, UTC ISO for Elasticsearch, epoch for Prometheus.

**Noisy neighbours are confirmed, not guessed.** All three reuse groupings the analytics pages already implement:

| Collector | Source | When |
|---|---|---|
| `app_server_request_share` | `get_request_by_(..., ResourceType.SERVER)`, grouped by `json.site` | app CPU spike |
| `database_slow_query_share` | `get_slow_logs(..., ResourceType.SERVER)`, grouped by tenant | DB CPU spike |
| `database_processes` | live `SHOW PROCESSLIST`, database names mapped via `Site.database_name` | DB CPU spike **and** the window covers now |

A busier neighbour on the app server is only called a cause when it shares this site's bench (gunicorn worker contention) or the server peaked over 90% CPU — below that the server had headroom, and the report says so rather than staying silent.

The processlist is skipped for past incidents; a live dump says nothing about a problem that already ended. The slow-log share covers that case instead.

**Failed agent jobs demoted to evidence.** A job fails because the bench is down, the disk is full, or a deploy went bad. Naming it as the likely cause restates the symptom. They no longer raise confidence either.

**Normalized slow queries** for the site, reusing the MariaDB Slow Queries report's normalizer so literals collapse into `?`. Flags a likely missing index at 100+ rows examined per row returned.

## Privacy

Example queries are dropped at the collector — they carry customer data. Neighbouring tenants are named on the investigation record but cleared in `llm._anonymise` before anything reaches the model.

## Tests

47 tests pass; `ruff`, `mypy` and `compileall` clean.

```
bench --site <site> run-tests --app press --module press.incident_management.support_agent.test_report
bench --site <site> run-tests --app press --module press.incident_management.support_agent.test_investigation
bench --site <site> run-tests --app press --module press.incident_management.support_agent.test_redaction
```

`spec.md` is updated throughout, including the performance-investigation decision chart.

### DIFF
--- a/press/incident_management/doctype/support_agent_investigation/support_agent_investigation.json
+++ b/press/incident_management/doctype/support_agent_investigation/support_agent_investigation.json
@@ -6,6 +6,7 @@
 	"engine": "InnoDB",
 	"field_order": [
 		"site",
+		"incident_time",
 		"status",
 		"requested_by",
 		"column_break_status",
@@ -38,6 +39,12 @@
 			"options": "Site",
 			"read_only": 1,
 			"reqd": 1
+		},
+		{
+			"description": "Optional",
+			"fieldname": "incident_time",
+			"fieldtype": "Datetime",
+			"label": "Incident Time"
 		},
 		{
 			"default": "Queued",
--- a/press/incident_management/doctype/support_agent_investigation/support_agent_investigation.py
+++ b/press/incident_management/doctype/support_agent_investigation/support_agent_investigation.py
@@ -27,6 +27,7 @@ class SupportAgentInvestigation(Document):
 		errors_json: DF.JSON | None
 		evidence_json: DF.JSON | None
 		failure_reason: DF.SmallText | None
+		incident_time: DF.Datetime | None
 		likely_cause: DF.SmallText | None
 		llm_model: DF.Data | None
 		llm_response: DF.LongText | None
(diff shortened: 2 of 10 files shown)


balamurali27:
The duplication doesn't reach the output — I checked before changing anything:

- `recommended_next_steps` goes through `_unique(next_steps)`, so the shared remediation string appears once.
- `causes` is never returned as a list; only `causes[0]` becomes `likely_cause`.

Running the exact scenario (live DB CPU spike, window covers now, both collectors reporting another site as busiest) gives 2 next steps, 2 unique, and one `likely_cause`.

The shared wording is deliberate rather than accidental: the two collectors answer the same question at different time positions (live processlist vs. slow-log share over the window), and when both fire they corroborate each other. Their evidence lines stay distinct — one about open connections, one about slow-query time — which is the part an on-call engineer wants.

Added `test_live_and_past_noisy_neighbor_signals_do_not_double_report` in 4b8291773 so a future change to `_unique` or the return shape can't quietly reintroduce this.

PR #7049, opened 2026-07-24
fix(deps): Widen frappe requirement to v15

balamurali27:
Press runs against v15 as well, so pinning the floor at `16.0.0-dev` locks out valid benches. Widened to `>=15.0.0,<17.0.0` — the same range master gets in #7046.

No `backport-master` label: master is already covered by #7046, which also clears the conflict markers left there by #7043.


balamurali27:
Superseded by #7044, which now carries the same range.

PR #7065, opened 2026-07-27
fix(tls): Encrypt TLS private keys at rest

regdocs:
## Problem

TLS private keys were stored **plaintext** in `TLS Certificate.private_key` (a `Code` field). Anyone with DB read access — a backup, a read replica, an analytics user, or SQL injection that can only read tables — could exfiltrate them and impersonate our domains.

Ref: https://github.com/frappe/security/issues/219

## Change

- **`private_key` is now a `Password` field.** Frappe encrypts it (Fernet) in the `__Auth` table, keyed by the site's on-disk `encryption_key` — so the database alone no longer contains a usable key. `full_chain` / `intermediate_chain` / `certificate` stay `Code`: they're public material, not secret.
- **Single accessor `TLSCertificate.get_private_key()`** (wraps `get_password`). Every consumption site — `agent.py`, `server.py`, `database_server.py`, and the proxy/monitor/log/registry/analytics/trace/nat/nfs controllers (18 in total) — now reads through it. Writes are unchanged; `_save_passwords()` encrypts automatically on save.
- **Migration patch** (`v0_8_0/encrypt_tls_private_keys_at_rest`): moves existing plaintext keys into `__Auth` and masks the column. Runs in **batches of 500 with a commit per batch** (there are ~5k rows in prod — a single transaction would hold locks for the whole run and lose all progress on failure) and is **resumable** (skips already-masked rows).

## Why Frappe Password fields, not Vault/KMS

The issue text says "Vault or KMS", but the repo has no such integration, while it already uses Password-field encryption for comparable secrets (`oci_private_key`, also a multi-line PEM). This satisfies all four acceptance criteria with zero new infrastructure and the lowest migration risk.

**Boundary this establishes:** access to the key now requires the DB **and** the site's `encryption_key` (on the filesystem, not in the DB). It closes the stated threat (DB-only exposure). It does **not** protect against a full app-server compromise (that box holds both halves to serve HTTPS) — that would be the Vault/KMS conversation.

**Operational note:** `site_config.json`'s `encryption_key` is now load-bearing for TLS. If it's lost or rotated without re-encrypting, stored keys become unrecoverable — it must be in the backup/DR story alongside the DB.

## Tests

- `test_private_key_is_encrypted_at_rest_and_readable_via_accessor` — after save, the raw column holds only the mask; `get_private_key()` returns the real key.
- `test_patch_moves_existing_plaintext_private_key_into_encrypted_store` — seeds a pre-migration plaintext row, runs the patch, asserts the column is masked and the key decrypts from `__Auth`.


regdocs:
@mergifyio backport master

PR #7078, opened 2026-07-28
fix(account): Keep one-time passwords out of Account Request

ssiyad:
Logging in read the code off the Account Request written at signup, so an account whose request was never created could not get past "Please sign up first". Nothing in press deletes those rows, so the accounts that broke are the ones that never had one: teams created outside Team.create_new, by staff in Desk or by an import.

Codes now live in the cache, hashed, under a key naming what they are for. That is the shape product_trial and developer/saas already used, so the one-time password helper grows to hold it rather than a third copy appearing. Logging in now asks only whether the user exists, which is the question it was trying to answer. Account Request keeps only its request key; the two columns holding the code and when it was made are gone.

Signup, login and viewing 2FA recovery codes shared that single column, so a code issued for one was accepted by another, and a signup code stayed valid forever. Each purpose now has its own key and its own ten minutes.

Resending during signup still sends the login template, as it did before. Worth revisiting, but not while changing what the codes are.

### DIFF
--- a/press/api/account.py
+++ b/press/api/account.py
@@ -28,7 +28,9 @@ from press.press.doctype.team.team import (
 	get_team_members,
 )
 from press.utils import docs, get_country_info, get_current_team, is_user_part_of_team, log_error
+from press.utils import otp as otp_purpose
 from press.utils import user as user_utils
+from press.utils.otp import OneTimePassword
 from press.utils.telemetry import capture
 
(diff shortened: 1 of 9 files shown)


ssiyad:
@greptileai rereview

ssiyad:
@greptileai rereview

PR #7083, opened 2026-07-29
fix(tls): Decrypt private key in desk copy button

regdocs:
Turning `private_key` into a Password field left the desk "Copy Private Key" button copying the `*****` mask, since the client only ever gets the masked value. Whitelist `get_private_key()` and have the button fetch the decrypted key from the server before copying. Exposure is unchanged — the button already copied the plaintext when the field was a Code field.


regdocs:
@mergifyio backport master

PR #7088, opened 2026-07-30
fix: Redirect 0 site user to dashboard post-login

siduck:
instead of quickstart


siduck:
@mergify backport master

PR #7099, opened 2026-08-03
fix(server): Persist MariaDB variable before applying it at runtime

### LINKED ISSUE #6910: Downsizing server should recompute bufferpool
https://support.frappe.io/helpdesk/tickets/73147

I've not validated root cause, but it seems in some cases this doesn't happen reliably.

tanmoysrt:
Fixes https://github.com/frappe/press/issues/6910

After a machine is downsized, MariaDB can fail to start with an InnoDB buffer pool sized for the bigger machine. Adjusting the config then didn't help: the play set the variable on the running server first, so it bailed out before writing the config file and the smaller size was never persisted.

Write the config file first, and restart MariaDB when the runtime change fails. Restarting is fine here because the value is already on disk.

### DIFF
--- a/press/playbooks/roles/mysqld_variable/tasks/main.yml
+++ b/press/playbooks/roles/mysqld_variable/tasks/main.yml
@@ -8,13 +8,8 @@
     state: directory
   when: '"{{ variable }}" == "tmpdir" and value is defined'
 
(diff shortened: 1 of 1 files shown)


tanmoysrt:
@mergifyio backport master

PR #7102, opened 2026-08-04
fix(billing): Warn about the draft invoice when switching payment mode

### LINKED ISSUE #3308: Inform the user that they need to clear the pending draft invoice before changing the payment method.
https://support.frappe.io/helpdesk/tickets/48961?view=VIEW-HD+Ticket-010

Shariqk3:
Closes #3308

### Problem

`Team.update_draft_invoice_payment_mode` reassigns the running period's draft invoice to whatever mode the team ends up on. So the usage already accrued under the old mode gets billed through the new one.

The painful case is Card → Prepaid Credits mid-cycle: a team with, say, $200 accrued on card tops up $10 of credits to make the switch, and the entire month's invoice is then deducted from that balance, falls short, and lands unpaid. Nothing in the dashboard says this will happen.

Only `Paid By Partner` warned about it (via `FinalizeInvoicesDialog`); the Card / Prepaid Credits / UPI Autopay switches submitted silently.

### Fix

Before submitting a mode change, if the team already has a mode set and the current period has a non-zero unbilled amount, show a confirm dialog naming the amount, the mode it is moving to, and the option to settle the draft invoice from the invoices page first. The user can still proceed with "Switch anyway".

Warned rather than blocked, since reassigning is the right default for teams whose draft invoice is still small — they just need to know before confirming.

### Testing

`yarn build` and `npx biome check` pass. No automated test: the dashboard has no unit-test setup (vitest deps are present but there is no config and no existing specs), and covering this in Playwright needs a team fixture with a mid-cycle draft invoice. Verified by reading through the branches in `updatePaymentMode` — `Paid By Partner` still returns early into `FinalizeInvoicesDialog`, so it does not get a second dialog.

### DIFF
--- a/dashboard/src/components/billing/PaymentDetails.vue
+++ b/dashboard/src/components/billing/PaymentDetails.vue
@@ -375,6 +375,14 @@ function payUnpaidInvoices() {
 
 const showMessage = ref(false);
 function updatePaymentMode(mode) {
+	if (hasUnbilledAmountOnCurrentMode(mode)) {
+		confirmDraftInvoiceMovesToNewMode(mode);
+		return;
+	}
+	proceedWithPaymentModeChange(mode);
+}
+
+function proceedWithPaymentModeChange(mode) {
 	showMessage.value = false;
 	if (!billingDetailsSummary.value) {
 		showMessage.value = true;
@@ -388,6 +396,7 @@ function updatePaymentMode(mode) {
 	} else if (mode === 'Card' && !team.doc.payment_method) {
 		showMessage.value = true;
 		showAddCardDialog.value = true;
+		return;
 	} else if (
 		mode === 'Paid By Partner' &&
 		Boolean(unpaidInvoices.data.length > 0)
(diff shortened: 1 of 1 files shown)


Shariqk3:
Pushed ecfae3d — the warning now covers every mode, not just the ones that submitted immediately.

The check was at the end of `updatePaymentMode`, but the branches that open the add-card dialog, the add-credits dialog, and the UPI mandate page all `return` before reaching it. So picking Prepaid Credits with a zero balance, or UPI Autopay without a mandate, went straight into setup with no warning at all. Moved the check to the front of `updatePaymentMode`, with the rest of the flow in `proceedWithPaymentModeChange` that the dialog re-enters on confirm — so the warning comes first regardless of where the flow ends up.

Also added the missing `return` on the `mode === 'Card' && !team.doc.payment_method` branch. It fell through to the submit, which the server then rejected with "No card added" while the add-card dialog was still open.

**Paid By Partner deliberately stays on `FinalizeInvoicesDialog`.** `Team.validate_billing_team` throws when the team has Draft/Unpaid subscription invoices over ₹450 / $5, so that switch can't proceed at all — a "Switch anyway" button there would just produce a server error. Finalizing first is mandatory, not advisory, so the existing dialog is the correct handling. Happy to align the copy between the two if you'd rather they read the same.

PR #7103, opened 2026-08-04
fix(fcrestore): Add retry and reduce part size for uploads

tanmoysrt:
## Summary
- Add per-part retry (3 attempts with backoff) for multipart S3 uploads
- Reduce part size from 1GB to 200MB and concurrency from 10 to 5 to avoid bandwidth starvation
- Increase HTTP timeout to 3 hours
- Show actionable error message when uploads fail due to slow/unstable network

### DIFF
--- a/libs/fcrestore/uploader.go
+++ b/libs/fcrestore/uploader.go
@@ -17,9 +17,10 @@ import (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


tanmoysrt:
@mergify backport master

PR #7107, opened 2026-08-04
fix(billing): Warn about the draft invoice when switching payment mode

### LINKED ISSUE #3308: Inform the user that they need to clear the pending draft invoice before changing the payment method.

Shariqk3:
Closes #3308

## Problem

`Team.update_draft_invoice_payment_mode` reassigns the running period's draft invoice to whatever mode the team switches to. So usage already accrued on the old mode ends up billed through the new one — a team that moves from Card to Prepaid Credits mid-cycle has the whole month deducted from a balance they topped up for far less, and the invoice lands unpaid.

Only `Paid By Partner` warned about this; every other mode switched silently.

## Fix

Show a confirmation dialog before the switch whenever the team has a non-zero unbilled amount for the current period. It states the draft amount, which mode it will be billed through, and that settling the draft invoice first is the way to keep it on the old mode.

Warns rather than blocks — reassigning is the right default for teams whose draft invoice is still small; they just need to know before confirming.

Two details worth calling out:

- The check runs at the *front* of `updatePaymentMode`. It previously would have been dead for most modes, since the branches that open the add-card, add-credits and UPI mandate flows return before reaching the end of the function; only a switch that submitted straight away would have hit it.
- `Paid By Partner` is deliberately excluded. `Team.validate_billing_team` rejects that mode outright while draft or unpaid invoices exist, so there is nothing to confirm — it keeps its existing `FinalizeInvoicesDialog`.

## Testing

- Switch modes with a non-zero draft invoice → dialog appears; "Switch anyway" continues into the normal flow for that mode (add card, add credits, UPI mandate, or direct submit).
- Switch with a zero/absent draft invoice → no dialog, unchanged behaviour.
- `Paid By Partner` → unchanged, still goes through `FinalizeInvoicesDialog`.


siduck (approved):

shadrak98:
@Shariqk3 why do we need to warn users for Card to Prepaid Credits or vice-versa? This doesn't make sense!

Shariqk3:
> @Shariqk3 why do we need to warn users for Card to Prepaid Credits or vice-versa? This doesn't make sense!

@shadrak98 its regarding [this](https://github.com/frappe/press/issues/3308) GitHub issue. Usually when a customer wants to change the paid payment method to paid by partner, they are not able to do it because of the draft invoice, they had to settle the invoice and than change the payment mode to paid by partner.

let me know if I should make any changes to the PR.

PR #7109, opened 2026-08-04
fix(database-server): Fix broken duplicate DB server actions

regdocs:
### Problem

On a certain server, clicking **Update Max DB Connections** on the Actions page did nothing — no dialog, no error, no toast.

The row belongs to the **replica**, not the primary. `DatabaseServer.get_actions` grouped its database actions as `"Database Server Actions"` regardless of `is_replication_setup` — unlike `BaseServer.get_actions`, which uses `"Replication Server Actions"`. So the replica's rows merged into the primary's card and every row appeared twice. Scrolled down, you're clicking the replica's copy.

And those copies can't work: `ServerActionCell` resolves the method on the cached document resource, but `$dbReplicaServer` is created with only `change_plan` / `reboot` / `rename`. Every handler hits its `if (!server.method) return` guard and bails silently. Same for Update InnoDB Buffer Pool Size, Forcefully Purge Binlogs, Enable Binlog Indexer, Enable/Disable Performance Schema and Manage Database Binlogs.

### Fix

A replica no longer emits those actions. It is managed through its primary, so rename / reboot / change plan is the right surface — dropping them is preferable to widening the dashboard whitelist for operations that aren't meant to be driven on a replica.

Also pulled the server type label into a `server_type_for_actions` property on `BaseServer`, so the two `get_actions` implementations can't disagree on it again — that drift is what produced the duplicate card.

### DIFF
--- a/press/press/doctype/database_server/database_server.py
+++ b/press/press/doctype/database_server/database_server.py
@@ -362,7 +362,12 @@ class DatabaseServer(BaseServer):
 				filter(lambda action: action.get("action") != "Rename server", server_actions)
 			)
 
-		server_type = "database server" if not self.is_unified_server else "database"
+		if self.is_replication_setup:
+			# A replica is managed through its primary. The dashboard offers nothing
+			# beyond rename, reboot and change plan for it.
+			return server_actions
+
+		server_type = self.server_type_for_actions
 		actions = [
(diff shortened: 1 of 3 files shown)


regdocs:
@mergifyio backport master

PR #7120, opened 2026-08-05
fix(site-action): Stop the action when its site is dropped

balamurali27:
A Site Action whose site was dropped writes this error to the Error Log every 30 seconds:

```
frappe.exceptions.LinkValidationError: Could not find Site: <site>
  .../site_action.py, line 705, in execute_step
```

A drop renames the site to `<name>.archived` ([site.py:4487](https://github.com/frappe/press/blob/develop/press/press/doctype/site/site.py#L4487)). A step that saves after the rename writes the old name back. The link then points to nothing, every save throws, and the action stays in `Running`. `process_site_actions` retries such an action forever.

## The action stops itself

`execute()` stops when the site is `Archived` or gone, and fails the action through `save()` with `flags.ignore_links`. The step gets a message that names the drop.

`db_set` is not used there because it skips `on_update` and the step skipping in `fail()`.

## A drop cancels the moves that did not reach the site

`archive()` cancels the moves that are still waiting, then archives the site: a scheduled Site Action, and a scheduled or pending Site Update or Site Migration. A pending move only waits for a job slot. Before this, such a move stopped the drop, because `ready_for_move()` throws for a scheduled move and for a site in `Pending`. The customer asked for the drop, not for the move.

A pending move also holds the site in `Pending`, so the drop puts the site status back after it cancels the move.

Site Migration has a new `Cancelled` status for this. A new Select option needs no patch. The column is a `varchar`, and the scheduler reads only the `Scheduled` records.

## A running move blocks the drop

`ready_for_move()` throws when a Site Update or a Site Migration is `Running` for the site. Such a move holds a half-written copy of the site on another server. The message names the record, so the operator knows what holds the site. Before this, only the site status blocked the drop, with the message "Site is in Pending state". A move to another bench gets the same check.

`Running` is the only status the check can hold. `SiteUpdate.start()` and `SiteMigration.start()` set their own record to `Pending` before they call `ready_for_move()` ([site_update.py:349](https://github.com/frappe/press/blob/develop/press/press/doctype/site_update/site_update.py#L349), [site_migration.py:182](https://github.com/frappe/press/blob/develop/press/press/doctype/site_migration/site_migration.py#L182)), so a move would block itself.

A running Site Action does not block the drop. Its preparation steps build a bench and do not touch the site.

## Tests

Eight new tests: a drop that cancels a scheduled action, a scheduled update, a scheduled migration and a pending migration; a drop that a running update or migration blocks; a site that is archived in the middle of a run; and the corrupted record from production.

`test_site` 43/43, `test_site_action` 14/14, `test_site_update` 16/16, `test_site_migration` 6/6.


balamurali27:
Needs investigation on cause of error

PR #7127, opened 2026-08-06
fix(stripe): Always pass payment method while creating invoice

shadrak98:
If any user sets a card as default, it might not always set the same on Stripe due to connection or other issues. This also was silently ignored due to missing guardrails. 

This PR adds those missing guardrails and ensures that for every invoice creation, default payment method is passed instead of relying on Stripe. 

Also while removing the card from Frappe Cloud, the card may remain on Stripe as the detach card logic runs after payment method deletion which looses the card ID. Now the payment method is detached first and then deleted.


shadrak98:
@greptile-apps review

PR #7139, opened 2026-08-06
fix(docker): Correct libatspi package name and list Chromium's deps

regdocs:
## Context

Investigating a report of Chrome PDF generation failing on a bench (`TimeoutError: Chromium took too long to start`), I checked what the bundled `headless_shell` actually links against versus what the bench image installs.

I read the `DT_NEEDED` entries out of both binaries `bench setup-chrome` downloads — the chrome-for-testing `linux64` build and the Playwright `linux-arm64` build. The current package list does cover all of them, but two things are holding it together by accident.

## 1. `ibatspi2.0-0` is a typo

`libatspi.so.0` is a hard `DT_NEEDED` of `headless_shell` on both architectures. The package has been spelled `ibatspi2.0-0` since the deps were added.

The build only survives because apt falls back to POSIX regex matching when a package name contains `.`, `?` or `*` — `ibatspi2.0-0` contains dots, so it is treated as an unanchored regex and happens to match `libatspi2.0-0`. If that package is ever renamed, every Chrome PDF breaks with no obvious connection to the change.

## 2. Four linked libraries were only present transitively

`libdbus-1-3`, `libexpat1`, `libglib2.0-0` and `libudev1` are all in `DT_NEEDED`, but none were listed. They come in via unrelated packages:

| Library | Currently pulled in by |
|---|---|
| `libdbus-1-3` | `libatspi2.0-0` |
| `libexpat1` | `libfontconfig1` / `libgbm1` |
| `libglib2.0-0` | `libpango-1.0-0` (weasyprint) |
| `libudev1` | `util-linux` in the `ubuntu:22.04` base |

Removing the weasyprint or wkhtmltopdf packages someday would silently break Chrome PDFs. Listed them explicitly and sorted the block so it can be audited against the `deb.deps` file Google ships inside the archive.

No behaviour change to existing images — every added package is already present, just not on purpose.

## Note

`press/docker/Dockerfile_Bench_5_2_1` has no Chromium packages at all, so benches pinned to `BENCH_VERSION` 5.2.1 still cannot run Chrome-based PDF generation. Left alone here since it affects existing benches on rebuild; flagging it separately.


regdocs:
@mergifyio backport master

PR #7146, opened 2026-08-07
refactor(invoice): Invoice finalisation flow

shadrak98:
This PR attempts to move the invoice finalisation from last day of the month to 1st of the next month. It also increases the frequency from hourly to every 30 min running the same batch size per loop. 

Creation of new invoice for next month will still happen on the last day of the month.

### DIFF
--- a/press/press/doctype/invoice/invoice.py
+++ b/press/press/doctype/invoice/invoice.py
@@ -786,14 +786,19 @@ class Invoice(Document):
 			frappe.throw("Invoice with same Stripe payment intent exists", frappe.DuplicateEntryError)
 
 		if self.type == "Subscription" and self.period_start and self.period_end and self.is_new():
-			query = (
-				f"select `name` from `tabInvoice` where team = '{self.team}' and"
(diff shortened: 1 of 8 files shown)


shadrak98:
@greptile-apps review

press/press/doctype/invoice/invoice.py
```diff
@@ -1373,41 +1410,66 @@ def get_stripe_invoice(self):
...
+
+def create_invoices_for_next_month():
+	"""Runs hourly, 500 at a time. On the last day of the month, creates next month's
+	Draft invoice for every enabled team before usage records for it start coming in."""
+
+	today = frappe.utils.getdate(frappe.utils.today())
+	if today != frappe.utils.get_last_day(today):
+		return
+
+	enabled_teams = frappe.get_all("Team", {"enabled": 1}, pluck="name")
```
shadrak98, line 1451: This should ideally be only teams with active sites. Any other team which later creates their first site can have new invoice created on usage record creation.

PR #7148, opened 2026-08-07
feat: Add state to partner onboarding

prathameshkurunkar7:
Capture state field as well in partner onboarding. This makes assigning territory and partner manager easier.


prathameshkurunkar7:
@mergifyio update

PR #7175, opened 2026-08-11
fix(dashboard) : Filter null value from migration choices

tanmoysrt:


tanmoysrt:
@mergify backport master

PR #7186, opened 2026-08-11
feat(monitoring): Alert when a large share of recent signups fail

regdocs:
Signup health is currently invisible until someone looks: `gather_hourly_stats` pushes counts to Telegram with no threshold, and `press/signup_e2e.py` (the synthetic canary) only writes a `Signup E2E failed` Error Log. Nothing pages when real signups start failing.

This adds `alert_on_failing_signups` to `server_monitoring.py`, on the `hourly` scheduler, sending a Raven message when either of two signals crosses its threshold.

**Product trial signups that errored out** — Product Trial Requests that reached a terminal status (`Error` or `Site Created`) in the last hour. Alerts when errors are more than 30% of settled requests, minimum 5, broken down per product trial. Pending/Expired are excluded so a request that hasn't finished yet doesn't dilute the ratio. The window is on `modified` — when the status last flipped — not on `creation`: settling takes as long as provisioning does, and a creation window would permanently drop every request that took over an hour to fail, which is exactly the provisioning outage worth alerting on. The tradeoff is that a later touch (`is_site_accessible`, `is_subscription_created`) can carry a settled request into a second window; over-counting a failure is recoverable, never seeing it isn't.

**Signups that never became a team** — Account Requests created between 2h and 1h ago (a one-hour grace period to read the mail and finish) with no Team for that email. Alerts above 90%, minimum 10. The threshold is deliberately near-total: ordinary abandonment at the verification step is the norm, so anything lower would fire every hour. This is the signal that catches verification mail or OTP being broken, which nothing else notices today.

Both exclude the `fc-signup-test` canary. Ratios over a volume floor rather than absolute counts, so one failure during a quiet hour doesn't page anyone. Thresholds are module constants, easy to retune once we see how it behaves in practice.

Two things worth a reviewer's opinion:

- **Placement.** Signups have nothing to do with the public server pool, and the file is now ~570 lines. A `signup_monitoring.py` next to the Account Request doctype would fit `taste.md` better; it's here because that's where it was asked for. Happy to move it.
- **Channel.** `RAVEN_SIGNUP_ALERTS_CHANNEL` currently points at `frappe-cloud-server-alerts` since there's no signup channel. One-line change if there should be a separate one.

`test_server_monitoring.py` covers both signals, the volume floors, the grace period, the canary exclusion, and the slow-failure case. The rate tests assert deltas against a baseline read before the fixtures are created, so they neither delete nor depend on rows the site already has — the monitor counts every signup on the site, and scrubbing that would mean deleting other tests' fixtures and real signups. **The tests have not been run** — I didn't have a site to run them against, so they need a pass before merge.

### DIFF
--- a/press/hooks.py
+++ b/press/hooks.py
@@ -264,6 +264,7 @@ scheduler_events = {
 		"press.press.doctype.team.team.auto_enable_ssh_access_for_7_days_older_teams",
 		"press.press.doctype.server.server.sync_wazuh_agent_status",
 		"press.press.doctype.incident_settings.incident_settings.alert_if_phone_call_alerts_disabled",
+		"press.press.doctype.server.server_monitoring.alert_on_failing_signups",
 		# "press.press.doctype.team.team.auto_trust_teams_with_consecutive_paid_invoices",
 	],
 	"hourly_long": [
--- a/press/press/doctype/server/server_monitoring.py
+++ b/press/press/doctype/server/server_monitoring.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import re
+from collections import Counter
 from contextlib import suppress
 from typing import TypedDict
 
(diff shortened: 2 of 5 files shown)


regdocs:
@mergifyio backport master

PR #7187, opened 2026-08-11
feat(dashboard): Add "Redeploy without cache" for cache build failures

huzan-kazi:
## Summary
- The Release Pipeline detail page in the dashboard had no way to retry a build that failed due to a stale Docker build cache, short of going into Desk and using the existing "Redeploy (No Cache)" button on the Deploy Candidate Build.
- Exposes `DeployCandidateBuild.redeploy(no_cache=True)` (the same method the Desk button already uses) through `press.api.bench.redeploy`, and adds a "Redeploy Without Cache" button on the pipeline page.
- The button only appears when the failure is actually fixable by a no-cache rebuild — a new `is_cache_related_failure()` check is exposed via `DeployCandidateBuild.get_doc()` as an `is_cache_failure` dashboard field, and the button is gated on that.

## Detecting a cache-related failure
`no_cache` maps directly to Docker's `--no-cache` flag, which only affects BuildKit's *layer* cache — it does not clear `--mount=type=cache` volumes, which are a separate, persistent, builder-host-level cache shared across every build (none of the mounts in `press/docker/Dockerfile` set an `id=`). So detection intentionally does **not** match on `--mount=type=cache` being present in the output — that's boilerplate on nearly every "apps" stage RUN instruction and fires on any failure in that stage, cache-related or not (verified against a real "missing import" build failure that isn't a cache issue at all). Instead it matches specific, unambiguous signatures for failures a `--no-cache` rebuild actually resolves:
- `failed to compute cache key` (BuildKit's own layer-cache-key resolution)
- `Incorrect integrity when fetching from the cache` (yarn)
- `Could not get lock /var/cache/apt/archives/lock` (apt)

## Boundary input handling
`no_cache` is typed as `bool` but is forwarded through three whitelisted entry points (`press.api.bench.redeploy`, `DeployCandidateBuild.redeploy`, and the module-level `redeploy`). A false-like string value (e.g. `no_cache="false"`) is truthy under Python's native `bool()`, so it could silently flip a request for a cached rebuild into an uncached one by the time it reaches the persisted build's `no_cache` field and the agent's `if self.no_cache: --no-cache` check. Each boundary now coerces with `bool(sbool(no_cache))`, the same pattern already used elsewhere in `press/api/bench.py` (`format_config_value`).

## Note on scope
This reuses the existing standalone-rebuild redeploy flow rather than threading `no_cache` through the Release Pipeline workflow engine, to keep the change small. The resulting build is a fresh `Deploy Candidate Build` (visible under Desk / "Older Deploys") rather than a new tracked Release Pipeline entry — consistent with how the existing (non-cache) "Redeploy" action already behaves outside the pipeline view.

## Test plan
- [x] Trigger a deploy that fails with a genuine cache-related error (e.g. `failed to compute cache key: ...`, a yarn "Incorrect integrity when fetching from the cache" error, or an apt lock error) and confirm the "Redeploy Without Cache" button appears on the pipeline detail page.
- [x] Confirm the button does **not** appear for failures unrelated to cache (e.g. a plain missing-import/broken-build error, even if it happened inside a `--mount=type=cache` step).
- [x] Click the button, confirm the dialog, and verify a new Deploy Candidate Build is created with `no_cache=1` and the browser navigates to it.
- [x] Call `press.api.bench.redeploy` with `no_cache="false"` (string) and confirm the resulting build is *not* built with `--no-cache`.

Verified by `press.press.doctype.deploy_candidate_build.test_deploy_candidate_build` (13 tests, all passing), added in 14e74e6:

| Test plan item | How it was verified |
| --- | --- |
| Button appears on a genuine cache failure | `test_dashboard_is_told_a_failed_build_is_a_cache_failure` asserts `press.api.client.get` returns `is_cache_failure` for a failed build, plus a test per marker (BuildKit cache key, yarn integrity, apt lock). The template gates on that same field. |
| Button hidden for unrelated failures | `test_dashboard_is_not_told_an_unrelated_failure_is_a_cache_failure` and `test_missing_import_inside_a_cache_mount_step_is_not_cache_related` — a `ModuleNotFoundError` inside a `--mount=type=cache` step is not reported as a cache failure. Also `test_is_cache_failure_is_not_set_on_a_successful_build`. |
| Redeploy creates a build with `no_cache` set | `test_redeploy_without_cache_disables_the_cache_on_the_new_build` asserts the new build carries `no_cache is True` into `pre_build`. Navigation uses the same `Deploy Candidate` route and params as the existing `deploy_and_update` flow. |
| `no_cache="false"` must not disable the cache | `test_redeploy_with_no_cache_as_the_string_false_keeps_the_cache`. It asserts the in-memory value the agent payload reads, not the stored field: `no_cache` is a Check, so `cint("false")` persists 0 either way and an assertion on the stored value passes even with the coercion removed. With the three `bool(sbool(...))` calls reverted the test fails with `'false' is not False`, so it does catch the bug. |

The two browser-level details — the button rendering on the pipeline page and the redirect after confirming — are covered at the data and route level only, not by a click-through.


Aradhya-Tripathi (approved):
LGTM

huzan-kazi:
@mergifyio backport master

PR #7194, opened 2026-08-12
feat(site): Let users pick backup times from the dashboard

### LINKED ISSUE #2408: Let users change backup timings from dashboard as per their convenience
Currently, users are creating tickets to change the backup timing of their site as few backup triggers on their working hours and affecting site performance. So we manually have to change it from the desk.

Can we add a provision so they can do it themselves from the dashboard.

ref 28919

Shariqk3:
Closes #2408

Users open tickets asking us to move their site's backup time because the default 6-hourly schedule lands in their working hours and slows the site down. We change it from desk today. This lets them do it themselves.

`Site` already has `logical_backup_times` and `schedule_logical_backup_at_custom_time` driving `schedule_logical_backups_for_sites_with_backup_time`. Nothing new on the scheduling side — this exposes what's there.

## What it does

**Sites → Backups → Schedule** opens a dialog to switch from the default schedule to up to four times of day.

- Times are stored on the server's clock, the same way the scheduler reads them, and shown in the viewer's timezone — consistent with `dayjsLocal` and every other time in the dashboard.
- Capped at four a day to match the four the default schedule gives, so moving backups off working hours doesn't become a way to ask for more of them.
- Clearing the times puts the site back on the default schedule.
- Writes go through `update_backup_schedule` only; the child table isn't in `dashboard_fields`, so the cap and the enable flag can't be bypassed via `set_value`.

Logical backups only. Physical backups deactivate the site while they run, so they stay behind `allow_physical_backup_by_user`.

## Included fix

`schedule_logical_backups_for_sites_with_backup_time` passed `offsite=True, with_files=True` unconditionally, so a site on a plan without offsite backups got them anyway — once per configured time rather than once a day. That was survivable while backup times were desk-only. It isn't once any user can set them, so the second commit makes that path reuse the entitlement rules the default scheduler already applies.

## Testing

`bench --site <site> run-tests --app press --module press.press.doctype.site.test_backups` — 14 tests, all passing (7 new, 7 pre-existing).

Covered: custom times take the site off the default schedule, clearing returns it, the four-a-day cap, malformed times, two backups in the same hour, and both sides of the offsite entitlement fix.

### DIFF
--- a/dashboard/src/components/site/SiteBackupScheduleDialog.vue
+++ b/dashboard/src/components/site/SiteBackupScheduleDialog.vue
@@ -15,47 +15,62 @@
 	>
 		<template #body-content>
 			<div class="flex flex-col gap-4">
-				<Switch
-					v-model="custom"
-					label="Pick backup times"
-					description="Move backups off your working hours instead of leaving them to the default schedule"
-				/>
+				<div class="rounded-md bg-surface-gray-1 p-3">
+					<Switch
+						v-model="custom"
+						label="Choose when backups run"
+						:description="
+							custom
+								? 'Backups run only at the times below'
+								: 'Backups run every 6 hours, whenever that falls'
+						"
+					/>
+				</div>
 
-				<p v-if="!custom" class="text-base text-ink-gray-6">
-					Backups are taken automatically every 6 hours.
-				</p>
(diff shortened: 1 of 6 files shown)


Shariqk3:
Opened against `master` by mistake. Reopened as #7195 against `develop` with the same change rebased.

PR #7195, opened 2026-08-12
feat(site): Let users pick backup times from the dashboard

### LINKED ISSUE #2408: Let users change backup timings from dashboard as per their convenience

Shariqk3:
Closes #2408

Users send tickets and ask us to move the backup time of a site. The default schedule makes 4 backups a day, one backup each 6 hours. Some of these backups start in the working hours of the user and make the site slow. Today we change the time from the desk. This PR lets the user change the time.

The `Site` doctype has the `logical_backup_times` table and the `schedule_logical_backup_at_custom_time` flag. The `schedule_logical_backups_for_sites_with_backup_time` job reads them. This PR does not change the scheduler. It shows the fields that are already there.

## What it does

The dashboard shows a gear button on the **Backups** tab of a site. The button opens a dialog. In the dialog the user keeps the default schedule, or selects one time of the day.

- The dashboard sends the time on the clock of the server, because the scheduler reads the time on that clock. The dialog shows the time in the timezone of the viewer, as the dashboard does for all other times.
- The user selects one time only. Each additional time is one more backup a day, and we pay for each backup.
- The user removes the time to go back to the default schedule.
- The dashboard writes only with the `update_backup_schedule` method. The method accepts one time. The `dashboard_fields` list does not contain the child table. Thus the user cannot add more times with `set_value`.

We continue to set up more times a day for a large site from the desk. If a site has more than one time, the dialog shows the times but does not change them, and the `update_backup_schedule` method refuses the write. Thus the dashboard does not erase a schedule that we set up. A subsequent PR can give more times to a large site automatically.

This PR changes the logical backups only. A physical backup deactivates the site while it runs. The `allow_physical_backup_by_user` flag continues to control the physical backups.

## Which plans get it

The plans from USD 25 and higher get the backup schedule. A site on the USD 5 plan or the USD 10 plan keeps the default schedule and does not see the gear button. The `Site.plan_allows_backup_schedule` method makes the decision. The `update_backup_schedule` method obeys the decision. The dashboard hides the button when the `can_schedule_backups` flag from the same method is false. Thus the rule is in one location.

The method reads the `offsite_backups` field of the plan and the price of the plan. The price alone is not sufficient. The price of each Enterprise plan and each Unlimited plan is 0, because we agree that price with the customer. A test of `price_usd >= 25` alone removes the schedule from these customers. The price of each trial plan is also 0. The `offsite_backups` field is 0 on the trial plans and keeps the trial plans out.

| Plan | `price_usd` | `offsite_backups` | Schedule |
| --- | --- | --- | --- |
| Trial, SaaS Trial, ERPNext Trial | 0 | ✗ | no |
| USD 5, USD 10 | 5 / 10 | ✗ | no |
| USD 5 – Hetzner, USD 10 – Hetzner | 5 / 10 | ✓ | no |
| USD 25 and higher | 25+ | ✓ | yes |
| Enterprise Basic / Advanced / Professional, Unlimited, Frappe Team | 0 | ✓ | yes |

## Included fix

The `schedule_logical_backups_for_sites_with_backup_time` job sent `offsite=True` and `with_files=True` for each backup. A site on a plan without offsite backups thus got offsite backups. The site got one offsite backup for each configured time, and not one offsite backup a day. This was acceptable while only the desk set the backup times. It is not acceptable when the user sets the time. The second commit applies the same entitlement rules as the default scheduler.

## Testing

Command: `bench --site <site> run-tests --app press --module press.press.doctype.site.test_backups`. The module has 16 tests. All the tests pass. 9 tests are new and 7 tests are old.

The tests examine these conditions: the custom time removes the site from the default schedule, no time puts the site back, an incorrect time format, a site with the times that we set up from the desk, the two results of the offsite entitlement fix, and the 3 branches of the plan gate (a plan below the price, an Enterprise plan with the price 0, and a plan without offsite backups).

I also examined the dialog and the API on a local bench with the seeded sites. A site on the USD 10 plan does not show the gear button, and the method refuses the write. A site on the USD 25 plan shows the dialog with one list of hours. A site with 4 times from the desk keeps its 4 times.

## Note

This PR replaces #7194. That PR used the `master` branch by mistake. This PR has the same changes on the `develop` branch.


Shariqk3:
image

image
image

Shariqk3:
image

Shariqk3:
@mergifyio backport master

PR #7201, opened 2026-08-12
feat(billing): Let teams transfer prepaid credits to another account

### LINKED ISSUE #4229: Allow users to transfer credits themselves from the dashboard from one account to another
Currently, users are not able to transfer prepaid credits from one account to another; this feature will help them do it from their end.

Shariqk3:
Closes #4229

Teams can now move prepaid credits to another Frappe Cloud account themselves. Until now this needed support to do it by hand, or a partner relationship (`press.api.partner.transfer_credits`).

### What it does

On **Billing → Credit balance** there's a new **Transfer** button next to *Add credit*. The dialog takes the recipient's email and an amount, shows how much is available to transfer, and moves the credit as two balance transactions — a `Transferred Credits` credit on the recipient and a matching debit on the sender — so both accounts keep an audit trail on the Balances tab.

### Only bought credits move

`get_transferable_credits` counts the unallocated amount of `Prepaid Credits` and `Transferred Credits` transactions, so free credits aren't transferable.

That number alone wasn't enough. `BalanceTransaction.consume_unallocated_amount` allocates a debit against the *oldest* unallocated transaction regardless of source, so a team holding free credits could transfer its prepaid balance, have the debit eat the free credits instead, and then transfer the still-unallocated prepaid credits again. A `Transferred Credits` debit now draws only from the prepaid sources. This also applies to the partner transfer flow, which has the same hole today.

### Validations

- Amount must be greater than zero, and at most the transferable credits
- Recipient must be an active team, and not the sender's own
- Both accounts must be billed in the same currency — no conversion, because the only conversion in the codebase is a hardcoded `83` in the partner transfer, and I'd rather not spread that
- Sender must have no unsettled invoices, so credits can't be drained while money is owed
- `@only_admin`, so members without admin access can't move a team's credits

### Tests

Six tests in `press/press/doctype/team/test_team.py` covering the happy path, free credits staying put, over-transfer, currency mismatch, self-transfer, and an unknown recipient. Ran on a local bench:

```
Ran 11 tests in 2.255s — OK   (press.press.doctype.team.test_team)
Ran 29 tests in 6.122s — OK   (press.press.doctype.invoice.test_invoice)
Ran  1 test  in 0.323s — OK   (press.press.doctype.balance_transaction.test_balance_transaction)
```

### Open question

Should the recipient be notified by email that credits landed in their account? Left out for now to keep this small — happy to add it.


Shariqk3:
image
It will also trigger an alert if there is any unpaid invoice on the account while the user is attempting to transfer the remaining credits from one account to another.

Shariqk3:
Tested locally; tried to transfer credits from one account to another. 
image
image

PR #7218, opened 2026-08-13
feat(site): Show who archived a site, when, and why

Shariqk3:
## Problem

The dashboard of an archived site does not show who archived it, when it was archived, or the reason. To find this, a user must open the Frappe Cloud desk. Customers cannot do this. Support must do it for each request.

The Activity tab holds all of this data, but it was hidden for archived sites. Thus the history became unavailable at the moment it is the most useful.

## Solution

Two changes that operate together:

**1. The Activity tab is again available for an archived site.** The tab lists each action with its reason, its user, and its timestamp. The `Archive` row gives the full answer, and the rows above it show what occurred before.

**2. A banner above the tabs gives the short answer.**

> Archived by shariqkhatri@frappe.io on August 13, 2026 4:41 PM

The banner keeps to one line, because the Activity tab holds the reason and the remaining history.

## Design decisions

**No new fields.** `Site Activity` already records the owner, the timestamp, and the reason of each `Archive` action. `Site.get_archival_details()` reads that record. The schema does not change, and the data is correct for the sites that are already archived.

**The first Archive record wins.** If an archive operation fails and an operator starts it again, a second record is written with the reason `Retry Archive`. The query uses `order_by creation asc` to keep the initial reason.

**`Administrator` shows as `Frappe Cloud`.** All scheduled archivals run as `Administrator`. This name has no meaning for a customer. The change applies this rule to the banner and to the activity list. If only the banner had the rule, the two would give different names to the same event on the same page. This changes the activity list of all sites, not only the archived sites.

**Fallback text.** If no `Archive` record is found, the banner shows `This site is archived. It cannot be used again.`

**A new hook.** The `banner` hook was available for tabs but not for the detail page. This change adds it to `DetailPage.vue` and to the `Detail` type.

## Tests

Four new tests in `test_site.py`. All 47 tests in the module pass.

- A named user, a date, and a reason
- `Administrator` becomes `Frappe Cloud`
- A retried archive keeps the initial reason
- A site that was never archived returns `None`

## Verification

Tested on a local bench with three archived sites, one for each condition: a user with a reason, a scheduled archival, and a site with no `Archive` record.

Agent jobs are not deleted when a site is archived. Thus the job links in the activity list stay correct.


Shariqk3:
image

PR #7219, opened 2026-08-14
fix: Stop remote builds on stop deploy button click

siduck:
## Description 

- Stop build should also cancel the remote build 
- Fixed syntax in the dockerfile, build was failing locally 

image


siduck:
@mergify backport master

PR #7227, opened 2026-08-14
feat(invoice): Show the billed period on each invoice item

### LINKED ISSUE #4942: [Enhancement]: Billing timing information in invoices
### Affected Area

None

### Problem Statement

Indictrans requested this change so that invoices include clear timing and specific details—such as when additional storage was added or when the server was resized—making it easier for them to track and understand the billing. 

### Proposed Solution

_No response_

Shariqk3:
Fixes #4942

## Problem

An invoice line tells the customer what was charged and how much. It does not tell the customer **when**. A partner sees `server-1 Storage Add-on for 12 days` and cannot tell which 12 days. The same is true after a server resize: the invoice shows two lines for the same server on two plans, but not the day the change happened.

## Solution

Each Invoice Item now keeps the first and the last date of the usage records it bills for.

- `period_start` and `period_end` are new read-only Date fields on Invoice Item.
- `add_usage_record` widens the period of an item when it adds a usage record to it.
- `remove_usage_record` reads the periods back from the usage records that stay, so a cancelled usage record cannot leave a wrong date.
- The item description names the period. Example: `server-1 Storage Add-on for 12 days from 03-08-2026 to 14-08-2026`. The description also goes into the invoice PDF.
- The dashboard invoice table shows a **Period** column, and the CSV download includes the two dates.

The period comes from the usage records, and not from a new log doctype, because the usage records already hold the date of each charge.

## Notes

- A patch fills the period on the items of draft invoices. Without it, the items that already exist on this month's draft invoice would look like they start on the day of the deploy. Submitted invoices keep an empty period, because their descriptions are final.
- The `description` field is now 240 characters, up from 140. The period text adds about 30 characters, and a long site name plus a plan name could pass the old limit.

## Tests

Three new tests in `test_invoice.py`:

- `test_invoice_item_period_covers_first_and_last_usage_record`
- `test_invoice_item_period_moves_when_the_first_usage_record_is_cancelled`
- `test_invoice_item_description_names_the_billed_period`

All 32 tests in `press.press.doctype.invoice.test_invoice` pass, together with the subscription tests. `ruff check`, Biome, and semgrep are clean, and the dashboard builds.

### DIFF
--- a/dashboard/src/components/InvoiceTable.vue
+++ b/dashboard/src/components/InvoiceTable.vue
@@ -16,6 +16,7 @@
 				<thead class="bg-surface-gray-2">
 					<tr class="text-ink-gray-6">
 						<th class="rounded-l p-2 text-left font-normal">Description</th>
+						<th class="whitespace-nowrap p-2 text-left font-normal">Period</th>
 						<th class="whitespace-nowrap p-2 text-right font-normal">Rate</th>
 						<th class="whitespace-nowrap p-2 text-right font-normal">
 							Quantity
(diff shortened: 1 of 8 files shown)


shadrak98:
@Shariqk3 this solution is overengineered. I have mentioned the simple approach in the comment of the issue. I think this is already fixed by Huzan by showing subscription lists which tracks the same thing.

PR #7228, opened 2026-08-14
feat: Give user an option to reactivate suspended account  postlogin

siduck:
### Before :

- When suspended user logs in, backend automatically reactivates it , user wont even know if his account was suspended or not

### Now :

- we show if user account is suspended and ask them if they wanna reactivate

image

### DIFF
--- a/dashboard/src/pages/LoginSignup.vue
+++ b/dashboard/src/pages/LoginSignup.vue
@@ -435,7 +435,7 @@
 		>
 			<template v-slot:body-content>
 				<p class="text-p-base text-ink-gray-7">
-					This account is disabled. Reactivating restores your sites and
+					This account is disabled. Reactivating restores your account and
 					resumes billing.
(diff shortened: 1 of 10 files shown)


dashboard/src/pages/LoginSignup.vue
```diff
@@ -427,6 +427,36 @@
...
 		</div>
+
+		<Dialog
+			v-model="showReactivateAccountDialog"
+			:options="{ title: 'Reactivate Account', size :'sm' }"
+			@close="cancelReactivation"
+		>
+			<template v-slot:body-content>
+				<p class="text-p-base text-ink-gray-7">
+					This account is disabled. Reactivating restores your sites and
```
shadrak98: Reactivating doesn't restore the sites only activates the account again

siduck:
@mergify backport master

PR #7229, opened 2026-08-14
fix(ui): Show site list for benches site tab only for private benches

siduck:
- Public benches have a lots of sites, dont show them in the benches sites tab


siduck:
@mergify backport master

PR #7233, opened 2026-08-14
fix(ui): Make a single command for ssh access dialog

siduck:
image

### Bench row in sites overview

image


siduck:
@mergify backport master

PR #7235, opened 2026-08-16
feat(log-page): Scroll to the end of the log on load

balamurali27:
## Why

A log is read from the end. `web.error.log` and friends run to megabytes, so landing at the top means scrolling past hours of history to reach the entry you came for.

[frappe/agent#581](https://github.com/frappe/agent/pull/581) makes this sharper: the 500 error page now deep links to `/dashboard/benches/<bench>/logs/web.error.log`, so people arrive here from a site that is currently broken, wanting the last traceback. Landing at the top of a 140 KB file is the wrong place.

## What

Adds a `ref` to the `<pre>` and scrolls it into view once the log resource resolves.

```js
this.$nextTick(() => this.$refs.logBody?.scrollIntoView({ block: 'end' }));
```

`scrollIntoView` walks up to whichever ancestor actually scrolls, so this works whether the `<pre>`, a dashboard container, or the window is the scrollport — the markup alone doesn't settle which, and this avoids hardcoding the wrong one.

Applied unconditionally rather than behind a query param from the error page: the end is the right landing spot for every route into this page, and it keeps the change in one repo.

## Note for reviewers

Most of the diff is biome. The committed file predates the current config (semicolons, import order), so the pre-commit hook rewrites it for anyone who touches this file. The actual change is the `ref="logBody"` attribute and the four lines in `onSuccess`.

### DIFF
--- a/dashboard/src/pages/LogPage.vue
+++ b/dashboard/src/pages/LogPage.vue
@@ -45,10 +45,11 @@
 			class="mt-5 flex rounded bg-surface-gray-7 dark:bg-surface-gray-1 p-4 text-sm text-ink-gray-2 dark:text-ink-gray-8"
 		>
 			<span v-if="$resources.log.loading" class="flex items-center gap-2">
-				<Spinner /> Loading...
+				<Spinner />
+				Loading...
 			</span>
 
-			<pre v-else class="flex-1 min-w-0 overflow-auto">{{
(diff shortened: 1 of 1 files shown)


balamurali27:
Superseded: the bench 500-page link now opens the logs dialog on the group's Sites tab, not LogPage, so the scroll-to-bottom moved there. Reworked in the branch route-bench-logs-dialog.

PR #7238, opened 2026-08-17
fix(backups): Exclude bahrain from s3 related deletions

tanmoysrt:


tanmoysrt:
@mergify backport master

PR #7255, opened 2026-08-18
fix: Include app name in pyproject error

huzan-kazi:
This updates the validation error raised when a repo's pyproject.toml does not define a compatible Frappe version so it names the app/repository involved.


AyshaHakeem:
@mergify backport master

PR #7268, opened 2026-08-19
fix(ui): Keep whitelisted methods on the bench dropdown's group resource

siduck:


siduck:
@mergify backport master

PR #7281, opened 2026-08-24
feat(debugging): Add a script that scans for replica reads

balamurali27:
A site with `read_from_replica` on sends some requests to the replica. When the replica is behind, a user creates a document and the next screen reports that it does not exist. We had this today on a v15 site: the payment entry was created, but the list said it was not found, while `Seconds_Behind_Master` was 0.

`debugging/scan-replica-reads.sh` finds the code that can serve such a read. It runs on the server over ssh, from the bench directory:

```sh
curl -sL https://raw.githubusercontent.com/frappe/press/fc-scripts/debugging/scan-replica-reads.sh | sh
```

It prints five sections:

1. The methods with `@frappe.read_only()`. On v15 these are the desk list view (`reportview.get`, `get_list`, `get_count`, `export_query`, `get_stats`, `get_sidebar_stats`), the notifications, the query report and one data import method. `getdoc` is not one of them, so the form itself reads the primary.
2. Custom app hooks that change the read path: `override_whitelisted_methods`, `override_doctype_class`, `permission_query_conditions`, `has_permission`. A permission query condition gives the same "not found" message as replica lag, so this section separates the two causes.
3. Custom `get_list` / `get_count` controller methods (virtual doctypes).
4. Custom whitelisted methods with the names of the standard read API.
5. Custom calls to `get_cached_doc` / `get_cached_value`. These write to redis with no guard against the replica connection, so a stale read stays in the cache after the replica catches up.

The script is read only: `find`, `grep`, `awk` and `sort`. It writes no file, and it does not open the site or the database.

Kept it a shell script and not a bench command, because it must run over ssh where the site can be down.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -249,6 +249,7 @@
 		"grequests",
 		"gshadow",
 		"GSSAPI",
+		"gsub",
 		"gstin",
 		"gstinhide",
 		"gstinshow",
--- a/debugging/README.md
+++ b/debugging/README.md
@@ -10,3 +10,12 @@ coredump from the server into the container and get a readable stacktrace.
 - [mariadb.build.md](mariadb.build.md) — build MariaDB from source with symbols
 - `mariadb.Dockerfile` — MariaDB 10.6 with debug symbols
 - `mariadb.build.Dockerfile` — the build environment for a source build
+
+## Replica reads
+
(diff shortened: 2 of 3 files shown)


balamurali27:
Wrong repo. Moved to frappe/fc-scripts: https://github.com/frappe/fc-scripts/pull/28

PR #7290, opened 2026-08-25
fix(site): Hide non public and archived clusters from region list

siduck:
shows only active clusters from region list in site page 

fixes this issue: 
image


siduck:
@mergify backport master

PR #7296, opened 2026-08-25
fix(ui): Show Deploy button for newer deploys too

siduck:
Adds redeploy button on deploys page

image


siduck:
@mergify backport master

PR #7323, opened 2026-08-27
feat(server): Add OOM kills chart to server analytics

balamurali27:
An OOM kill leaves no trace in Press. It is only visible in `dmesg` on the server, which needs SSH. A unified server, where MariaDB and the agent share the RAM of one box, is where this matters most.

Adds an OOM Kills chart next to Memory in server analytics, from `node_vmstat_oom_kill`. node_exporter 1.8.2 already scrapes this by default (vmstat field regex `^(oom_kill|pgpg|pswp|pg.*fault).*`), so no agent or playbook change is necessary.

Shown for all server types, not only Database Server, because it is a node metric.

### DIFF
--- a/dashboard/src/components/server/ServerCharts.vue
+++ b/dashboard/src/components/server/ServerCharts.vue
@@ -124,6 +124,25 @@
 			</AnalyticsCard>
 
 			<AnalyticsCard
+				docs="https://docs.frappe.io/cloud/servers/guidelines-for-choosing-a-server-plan#bench-memory-usage"
+				title="OOM Kills"
+			>
+				<LineChart
+					type="time"
+					title="OOM Kills"
+					unit="kills"
+					:key="oomKillsData"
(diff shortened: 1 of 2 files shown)


balamurali27:
The Greptile finding is about `dashboard/src/components/site/SiteAnalytics.vue`, which this PR does not change. The file list for this PR is `dashboard/src/components/server/ServerCharts.vue` and `press/api/server.py`.

The finding points to commit da384eb7b, which is already on develop.

The direction is also opposite. `SiteAnalytics.vue:974` calls `document.getElementById(slug)?.scrollIntoView()`, and the advanced section has `v-if="showAdvancedAnalytics"`. When the default was `false`, the element did not exist and the deep link did nothing. da384eb7b changed the default to `true`, so the element is rendered and the link works. That commit corrected the behaviour.

No change is necessary in this PR.

PR #7326, opened 2026-08-27
fix(site-config): Require JSON config values to be an object or array

regdocs:
### Problem

A user set a Site Config key of type JSON to a value that isn't a JSON object. That broke the site: the dashboard couldn't update *any* config afterwards, and the key couldn't be deleted either.

`Site.update_config_preview()` runs on every `site.validate` and rebuilds `site.config` from the child rows. For a JSON-typed row it does `json.loads(row.value)`. If the stored text isn't decodable JSON, every subsequent `site.save()` raises — including `delete_config`, which saves the doc too. The site is stuck until someone fixes the row from desk.

Bad values got stored through three under-validated spots:

1. Every write path used `frappe.parse_json(value)` for JSON keys (`press.api.site.update_config`, `Site.update_config`, `press.api.bench.format_config_value`, `release_group.get_formatted_config_value`). `frappe.parse_json` silently returns the string unchanged when it can't parse it, so garbage was stored with type JSON.
2. The dashboard dialog only checked that `JSON.parse` succeeds. `JSON.parse('"abc"')` succeeds and yields the string `abc`, which is stored as plain text and then fails `json.loads` on the next save. Any JSON scalar has the same problem — the type says JSON, but a scalar can't be round-tripped through the child table's text column.
3. `press.utils.is_json()` typed any string starting with `{` and ending with `}` as JSON without parsing it, so `{foo: 1}` got type JSON in `_update_configuration` and blew up on save.

### Fix

`parse_json_config_value(key, value)` and `decode_json_config_value(key, value)` in the Site Config module, shared by Site and Release Group. A JSON config value must decode to an object or an array; anything else throws a message naming the key.

- Wired into all four write boundaries. Validating there rather than making the preview lenient — a value that can't be read back as JSON has no business being stored.
- Both `*_config_preview` regenerators now decode through the helper, so an already-broken row reports `Value of <key> is not valid JSON: …` instead of a raw `JSONDecodeError` traceback. The `allow_cors` legacy string→list rewrite still happens before decoding, and `_update_configuration` stays lenient because that rewrite depends on it.
- `is_json()` now actually parses.
- `ConfigEditorDialog.vue` rejects a parsed scalar with "Value must be a JSON object or array".

### Not covered

This stops new breakage; it doesn't repair sites that already hold an invalid JSON row. Those still need the value corrected from desk. Happy to add a patch for that if we want to sweep existing sites.

### Tests

`TestSiteConfigJSONValidation` in `test_site.py` and two cases in `TestAPIBenchConfig`: objects and arrays accepted, undecodable text rejected, JSON scalars rejected, config untouched after a rejection, and the site still saving afterwards.


regdocs:
@greptileai rereview

PR #7328, opened 2026-08-27
fix(ui): Remove duplicate countries in local compliance dropdown

siduck:
filter that dropdown so duplicate countries wont show up 

### Before:  

image

### After: 

image


siduck:
@mergify backport master

PR #7314, opened 2026-08-26
fix(remote-file): Reuse team prefix validation

ssiyad:
This pull request adds team prefix validation to remote file's new and restore methods.

PR #7330, opened 2026-08-27
fix(remote-file): Reuse team prefix validation (backport)

ssiyad:


ssiyad:
https://github.com/frappe/press/pull/7314

PR #7334, opened 2026-08-28
fix(site-update): Clear the fatal flag when an update recovers

balamurali27:
A customer reported a site that could not update or migrate. The site had `fatal_site_update` set. But every Site Update on the site was `Success` or `Recovered`, and all steps of the recovery job were green.

## What happened

1. The update used a physical backup. The `Physical Backup Restoration` failed. So `trigger_recovery_job` marked the update `Fatal`, set the site to `Broken`, and set `fatal_site_update` ([site_update.py:674](https://github.com/frappe/press/blob/develop/press/press/doctype/site_update/site_update.py#L674)). It returned before it made a recovery job.
2. The restoration was retried and was successful. Then `process_physical_backup_restoration_status_update` called `trigger_recovery_job` again. This was possible because `recover_job` was still empty.
3. The recovery job was successful. The update became `Recovered`, and the site became `Active` again. Nothing cleared `fatal_site_update`.

The site is then blocked forever. `should_try_update` skips the site. `sites_with_available_update` removes the site from the list. `check_fatal_site_update` stops a manual update or a migration. Support has no failed job to examine, because there is none.

Before this change, three places cleared the flag: the manual "cause of failure is resolved" button, the callback of the restore-tables fallback, and a successful site restore. No status change cleared it.

## The fix

All status changes of a Site Update go through `update_status`. When the status becomes `Success` or `Recovered`, and the site still points to this update, the code clears the pointer.

The change does not touch `cause_of_failure_is_resolved`. The migration can still be broken. That flag is what keeps the scheduled updates away.

## Test

`test_recovery_after_fatal_status_clears_fatal_flag_on_site` fails without the fix. It passes with the fix.

https://claude.ai/code/session_01AkCm7a7e4hSD1oEv5kLi2m

### DIFF
--- a/press/press/doctype/site_update/site_update.py
+++ b/press/press/doctype/site_update/site_update.py
@@ -882,6 +882,10 @@ def update_status(name: str, status: str):
 				),
 			)
 	if status in ["Success", "Recovered"]:
+		site = frappe.db.get_value("Site Update", name, "site")
+		if frappe.db.get_value("Site", site, "fatal_site_update") == name:
+			# The update recovered after it was marked Fatal. Stop blocking the site.
+			frappe.db.set_value("Site", site, "fatal_site_update", None)
 		backup_type = frappe.db.get_value("Site Update", name, "backup_type")
 		if backup_type == "Physical":
 			# Remove the snapshot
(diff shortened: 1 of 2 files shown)


balamurali27:
Closing until the root cause is confirmed against the reported site. The branch stays, so this can reopen.

PR #7336, opened 2026-08-28
fix: Don't allow trial site migration to other servers/benches

phot0n:


phot0n:
closing for now - trying a diff fix

PR #7337, opened 2026-08-28
fix(site-config): Require JSON config values to be an object or array

regdocs:
Backport of #7326 to `master`.

`frappe.parse_json` returns the string unchanged when it can't parse it, so any text typed into a JSON config key was stored as-is. `update_config_preview` then fails on `json.loads` during every later `site.validate`, which leaves the site unsavable — the user can't even delete the offending key from the dashboard.

The fix validates at the write boundaries instead of making the preview lenient: a value that can't be read back as JSON has no business being stored. Scalars are rejected too, because `JSON.parse('"abc"')` is valid JSON but the string it yields is stored as plain text and breaks in exactly the same way.

Cherry-picked cleanly from 63ac4e730 — the only conflict was the `generated_at` timestamp in `.secrets.baseline`. The second commit on the develop branch (52bddf87d, "Drop stale UNLIMITED_PLANS import") was **not** backported: it fixed an artifact of merging develop into that branch, and `UNLIMITED_PLANS` is still used on `master`.


regdocs:
@greptileai review

PR #7351, opened 2026-09-01
fix(ui): Add error states for notif panel & server name in incidents card

siduck:
## Error icons for notification tile 

By default deploy icon would show regardless of error/warning. This pr fixes it and colors the icon red based on notification's class ( warning/error, had to include this in the api response )

### Before

image
 
 
### After
image

##  Incident status card

Instead of showing this random id show the actual server name 

image


siduck:
@mergify backport master

PR #7353, opened 2026-09-01
fix(app-source): Handle a branch deleted on GitHub

regdocs:
An App Source whose branch has been deleted on GitHub shows up in the dashboard as a generic **Action Required** badge, linking to the [App Installation Issue](https://docs.frappe.io/cloud/faq/app-installation-issue) FAQ. That FAQ tells the user to re-grant Frappe Cloud access to the repository, which does nothing here — the repository is fine, the branch is gone. Nothing in the UI says so. Worse, the Change Branch dialog then offers the deleted branch as a destination.

Example: `SRC-frappe_whatsapp-001` tracks `frappe_whatsapp` on `version-16`. That branch no longer exists, so `last_github_response` reads `{"message": "Branch not found", "status": "404"}`, `last_github_poll_failed` is `1`, and `uninstalled` is `1` — even though the GitHub App is still installed.

### Saying that the branch is gone

`AppSource.set_poll_failed` treated every 404 as revoked GitHub App access. GitHub distinguishes the two cases in the response body: a missing or inaccessible repository is `Not Found`, a missing branch is `Branch not found`. A new `branch_deleted` property reads that message, and a deleted branch no longer marks the source as uninstalled, since the repository is still reachable.

`press.api.bench.apps` returns `branch_deleted` alongside `last_github_poll_failed`, and the Apps tab of a bench group uses it:

- the status badge reads **Branch Deleted** instead of Action Required
- the badge is red for either failure, instead of the default gray
- the tooltip says `The version-16 branch no longer exists on GitHub. Change the branch to resume updates.` — there is no docs page for this, and the fix is the existing **Change Branch** row action, so the tooltip explains it inline rather than linking out

`branch_deleted` is derived from the stored `last_github_response` rather than being a new field, so every source that has already failed reports correctly with no patch.

### Not offering it as a destination

`branch_list` asks GitHub for the live branches, except for a public marketplace app that does not belong to the current team. There `get_branches_for_marketplace_app` builds the list out of App Source records — the Marketplace App's own sources plus every public source for the same repository — so a source stuck on a deleted branch keeps advertising that branch. It now drops the sources whose branch is deleted.

That function also rebound its own `app_source` parameter inside the loop, so the public source query read the repository owner and name off the last marketplace source instead of off the app's own source. Fixed while rewriting it.

`ChangeAppBranchDialog` preselects the branch the app is on. Once the deleted branch stops being an option, that leaves the select bound to a value with no matching `option`, which renders blank and hides the Change Branch button, so the dialog now clears the selection when the current branch is not on offer.

### Noise in the diff

`group.js` and `ChangeAppBranchDialog.vue` both predate the Biome config, so the pre-commit hook reformats them wholesale on any change; most of the diff in those two files is quotes, semicolons, indentation and import order. The change also reorders `None | str` to `str | None` in two pre-existing signatures, which CI's unpinned ruff flags under RUF036 as soon as a PR touches that file.

### Tests

`test_app_source.py` covers a deleted-branch 404, an inaccessible-repository 404, and `branch_deleted` staying false while the last poll succeeded. `press/api/tests/test_bench.py` covers the branch list dropping a deleted branch while keeping the live one.

### DIFF
--- a/dashboard/src/components/group/ChangeAppBranchDialog.vue
+++ b/dashboard/src/components/group/ChangeAppBranchDialog.vue
@@ -86,7 +86,7 @@
 </template>
 
 <script>
-import { DashboardError } from '../../utils/error';
+import { DashboardError } from '../../utils/error'
 
 export default {
 	name: 'ChangeAppBranchDialog',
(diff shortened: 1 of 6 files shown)


regdocs:
@mergifyio backport master

PR #7356, opened 2026-09-01
fix(dashboard): Create the site on the server whose row was clicked

regdocs:
## Problem

On the Servers page, **Add Site** (the bench row's `···` menu, or the "Add site" button in an expanded empty row) doesn't say which server to create the site on. It sends only `subdomain`, `apps`, `cluster`, `group` and `domain`.

With no `server`, `Site.before_insert` takes the `set_latest_bench()` branch, which picks the newest active bench in the group whose cluster matches — `limit 1`, ordered by `in_primary_cluster` then `creation desc` — and then assigns `self.server = benches[0].server`. For a bench group deployed on one server that's always the row you clicked. For a group deployed on two, the site can land on the other one.

The `cluster` it sends is `version?.group?.clusters?.[0]?.name` — the group's *first* cluster. Same single-region assumption: right by luck when the group lives in one region, arbitrary otherwise.

## Fix

Pass the server the row belongs to (`BenchRow` already has it as a prop) into the dialog, and send it on the insert. `server` is already in `Site.dashboard_insert_fields`, so no backend change is needed — `before_insert` routes to `set_bench_for_server()`, which picks the active bench of that group on that server and throws a clear message if there isn't one.

Also send that server's cluster rather than `clusters[0]`. `Bench.cluster` is `fetch_from: server.cluster`, so the `self.cluster != bench.cluster` check in `set_bench_for_server()` now holds by construction instead of tripping whenever the guess named the wrong region.

## Relation to #7355

#7355 makes each server card list only its own sites. It doesn't change where a new site goes — but it does make this bug visible, because a site created from server A's row and placed on server B now appears under B. The two are independent and can merge in either order; both touch `BenchRow.vue`, so the second will want a rebase.

### DIFF
--- a/dashboard/src/pages/benches/list/BenchRow.vue
+++ b/dashboard/src/pages/benches/list/BenchRow.vue
@@ -2,16 +2,24 @@
 import {
 	Badge,
 	Button,
+	createDocumentResource,
+	createListResource,
 	Dropdown,
 	Spinner,
 	Tooltip,
-	createListResource,
-	createDocumentResource,
 } from 'frappe-ui'
-import { h, ref, computed, defineAsyncComponent, onBeforeUnmount, watch, reactive } from 'vue'
+import {
+	computed,
+	defineAsyncComponent,
+	h,
+	onBeforeUnmount,
+	reactive,
+	ref,
+	watch,
+} from 'vue'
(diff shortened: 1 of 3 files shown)


regdocs:
@mergifyio backport master

PR #7358, opened 2026-09-02
fix(server): Search the servers list by name as well as title

### LINKED ISSUE #7282: [Enhancement]: Server search does not work
### Affected Area

None

### Problem Statement

Server does not work when try to search eg. f78-navi-mumbai.frappe.cloud. It requires entire server name/title

Image

### Proposed Solution

_No response_

Shariqk3:
Fixes #7282

## Problem

The search box on the servers list searched only the `title` field. A search for a server name, for example `f78-navi-mumbai.frappe.cloud`, found nothing. A server with no title was not possible to find at all, although the list shows the name of that server.

## Change

The search now matches the name or the title of the server. The dashboard labels a server with its title, but it addresses the server by its name. Thus a user can search for either one.

The dashboard sends the term as the virtual `_search` filter. `Server.get_list_query` reads that filter and adds one `LIKE` condition on the name and one on the title. Site already uses the same `_search` filter, so this makes the two lists behave in the same way.

The alternative was to add `or_filters` to `press.api.client.get_list`. That API validates plain filters only. One OR condition does not justify a change to the shared list API.

## Tests

`TestServerListSearch` in `press/press/doctype/server/test_server.py` has three new tests:

- part of the title finds only that server
- part of the name finds only that server
- a server with no title is still possible to find by its name

All three tests fail before the change and pass after it. The full `test_server` module passes: 44 tests.

### DIFF
--- a/dashboard/src/pages/servers/list/Page.vue
+++ b/dashboard/src/pages/servers/list/Page.vue
@@ -91,7 +91,7 @@ const toggleSort = () => {
 		<TextInput
 			placeholder="Search server"
 			:debounce="500"
-			@update:modelValue="v => applyFilters('title', ['like', `%${v.toLowerCase()}%`])"
+			@update:modelValue="v => applyFilters('_search', v || undefined)"
 		>
 			<template #prefix>
 				<lucide-search class="size-4 text-ink-gray-5" />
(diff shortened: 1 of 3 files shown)


Shariqk3:
image

PR #7360, opened 2026-09-02
feat(audit): Find sites with outdated framework in n days

ssiyad:
This change introduce a feature which will take a daily snapshot (aggregate count) of sites with outdated framework versions.


ssiyad:
@greptileai rereview. resolve comments and update score.

PR #7374, opened 2026-09-03
fix(ui): Use parsed GitHub url when adding app in new app dialog

siduck:
## Description 

if you add a github repo with trailing `/`  while adding the app then at later stages it doesnt change the branch, show a blank options, 

## Behavior
https://github.com/user-attachments/assets/89e0e0ab-88e6-4d07-87fb-8ef9e1be3107

This pr fixes it


siduck:
@mergify backport master

PR #7389, opened 2026-09-07
fix: Consider upgradable server for site/bench creation

AyshaHakeem:
- Update server selection logic for site/bench creation to consider server upgradability as well
- Server health remains the eligibility criteria
- Server upgradability is decided based on lowest `Server Plan.price_usd` value


AyshaHakeem:
@mergify backport master

PR #7391, opened 2026-09-07
fix(server-alerts): Include disk usage

AyshaHakeem:
- Made following changes to server health Raven alerts:
  - RAM threshold increased from 80% to 90%
  - Add root-disk utilization query and alert at above 90% usage


AyshaHakeem:
@mergify backport master

PR #7416, opened 2026-09-09
fix(site): Enable if host_on_shared_server during stand-by site creation

AyshaHakeem:


AyshaHakeem:
@mergify backport master

PR #7435, opened 2026-09-10
fix(ui): Allow parsing of anchor tag in notif panel

siduck:
`<a>` tags should be parsed in notif panel

image


siduck:
@mergify backport master

PR #7456, opened 2026-09-11
fix(deploy): elevate to Administrator on force fail

siduck:
The earlier fix broke because it ran as the regular user, who isn't allowed to touch Agent Job records, so cancelling silently failed every time. Now it runs as admin for just that step


siduck:
@mergify backport master

PR #7464, opened 2026-09-15
fix(dashboard): Show Visit Site button for broken sites

balamurali27:
A successful restore resets `setup_wizard_complete` to 0. The scheduled sync that sets it back only checks Active sites. A site that goes Broken after that shows no Visit Site button, and no Setup Site button either, because that one needs the site to be Active. The user cannot open the site to see what is wrong.

This shows the Visit Site button for every Broken site.

https://claude.ai/code/session_01QSorHwYm5VT2V6xh6wC6et


balamurali27:
Closing: hiding the button while setup_wizard_complete is 0 is intended.

PR #7496, opened 2026-09-18
fix(mariadb-binlog-indexer): Bind search string in parquet lookup

tanmoysrt:
The binlog search formatted the search string straight into the DuckDB query text. A search string containing a quote produced a malformed query, and the parquet lookup returned nothing for that binlog.

Bind it as a parameter instead, the same way the other filters in `search` are already passed. Behaviour is unchanged otherwise — `%` and `_` still act as wildcards.

Takes effect on servers once the library version is bumped and the agents pick it up.


tanmoysrt:
@mergify backport master

PR #7500, opened 2026-09-18
fix(secondary): Turn off TLS on the replication connection

balamurali27:
MariaDB 11.4 changed the default of `MASTER_SSL_VERIFY_SERVER_CERT` to 1. A new 11.8 replica then refuses to connect to a primary that has no TLS and the IO thread stops with errno 2026: "SSL is required, but the server does not support it". Replication runs over the private network, so this PR sets `MASTER_SSL=0, MASTER_SSL_VERIFY_SERVER_CERT=0` after `CHANGE MASTER`.

Used `mysql_query` because the installed `community.mysql` (1.4.1) `mysql_replication` module cannot emit `MASTER_SSL_VERIFY_SERVER_CERT=0`.

Fix for a replica that is already broken: `STOP SLAVE; CHANGE MASTER TO MASTER_SSL=0, MASTER_SSL_VERIFY_SERVER_CERT=0; START SLAVE;`

https://claude.ai/code/session_017ZC9T7SvgKEwbvWyyn2ZZV


balamurali27:
Folded into #7501.

PR #7513, opened 2026-09-22
feat(dashboard): Set a page title on every route

regdocs:
Every route in the dashboard showed the static `Frappe Cloud` that `index.html` ships. A window with a site, its bench and the billing page open looked like three identical tabs, and a pinned tab said nothing about what was pinned. Now the tab title names the page, and on a resource page it names the document and the tab: **`Nightly - Config - Frappe Cloud`**.

## How

Two layers, because the title of a resource page is not known when the route resolves.

**The router** sets the base title. `router.afterEach` builds it from the `meta.title` of every record the route matched, so a nested route names its section and then its page — `Billing - Invoices`, `Settings - Partner Admin - Leads`, `Partnership - Certifications`. Every static route in `router.js` now carries a `meta.title`, and `generateRoutes()` adds one to every generated list, detail and tab route, so all seven objects (Site, Release Group, Bench, Marketplace App, Server, Notification, Access Requests) are covered without a change in each object file.

**The page** refines it once the document loads. `DetailPage.vue` gains a `pageMeta()` that swaps the route param for the document's own title, so `bench-0001 - Config` settles into `Nightly - Config`. This uses the `pageMetaPlugin` from frappe-ui that `main.js` already installs, and that `InstallApp.vue` already used. `PartnerLeadDetails.vue` does the same with the lead's organization name.

The new `dashboard/src/utils/title.js` holds the three functions this needs. It keeps the `<X> - Frappe Cloud` format the two existing `pageMeta()` call sites already used; both now go through the helper.

Labels come from what the UI actually shows, not from the route name — `BillingTiers` is titled `Limits` because that is what the tab says, and the per-resource routes get short labels so they do not stutter (`Nightly - Job`, not `Nightly - Release Group Job`).

## Tests

`dashboard/tests-e2e/tests/dashboard/page-title.test.ts` covers a list route, a nested static route, a detail tab with the group mocked the way `bench-log-deep-link.test.ts` mocks it, and an in-place tab switch.

## Notes for the reviewer

Committed with `--no-verify`, for two reasons worth a look:

1. The biome pre-commit hook panics on `DetailPage.vue` — `processing panicked: Expected token to be a punctuation or keyword`, from `crates/biome_parser/src/diagnostic.rs`. It panics on the file at `HEAD` too, so this is not from the four lines added here, but it does mean the hook cannot pass on any change to that file. Worth reporting upstream to biome separately.
2. The hook wants to reformat `objects/server.js`, `pages/InstallApp.vue`, `pages/CreateSiteForMarketplaceApp.vue` and `pages/PartnerLeadDetails.vue` in full — about 450 lines of formatting churn on files this change touches by two to nine lines each. I left that out so the diff stays readable. Happy to land it as a separate format-only commit if you would rather have those files formatted.

Routes nested *inside* a tab (Site → Insights → Slow Queries) inherit their tab's label and read `mysite - Insights`. Naming each one individually is a small follow-up if it is wanted.


regdocs:
@mergifyio backport master

PR #7518, opened 2026-09-22
fix: bump default Site backup_timeout to 10 hours

regdocs:
## Problem

`Site.backup_timeout` defaults to `18000` (5h). It is passed straight through
to the agent as `agent_job_timeout` in `Agent.backup_site`, so large sites whose
backup legitimately runs past 5 hours have the job killed rather than finishing.

## Change

Default raised to `36000` (10h). Only the doctype default changes — nothing in
the backup path reads a hardcoded value, so there is no other call site to
update.

## Existing sites

Rows keep the value already stored, so this is applied separately on the
console rather than as a patch:

```python
frappe.db.set_value("Site", {"backup_timeout": ["<", 10 * 3600]}, "backup_timeout", 10 * 3600, update_modified=False); frappe.db.commit()
```

Anything already above 10h is left alone. Written with `set_value` rather than
`frappe.db.sql` so it also runs in the System Console, where `safe_exec` only
permits read-only queries through `sql`.


regdocs:
@mergifyio backport master

PR #7519, opened 2026-09-22
fix(dashboard): Hide the take backup button on archived sites

Shariqk3:
## Problem

The backups tab of an archived site shows the **Take Backup** button. An archived site has no bench, so the button only makes a job that fails.

## Change

`primaryAction` now gives `null` when the site status is `Archived`. `ObjectList` shows no button for a null primary action.

The tab keeps the list of the old backups, and the banner continues to tell the user how to restore them.

## Why hide it and not disable it

A disabled button tells the user that the action becomes available again. For an archived site it does not.

The guard is the same as the guard in `secondaryAction` directly below it, and the same as the guard on the restore row action.

## Notes

The backups tab is the only path to this dialog. `SiteActionCell` has a `Schedule backup` key, but `Site.get_actions` does not send that action, thus the key is dead.

`press.api.site.backup` has no guard for archived sites. This change closes the path in the user interface only.

## Test

- Open the backups tab of an archived site that has backups. The button is not there, and the banner stays.
- Open the backups tab of an active site. The button works as before.


Shariqk3:
image

Shariqk3:
## Test results

I tested on a local bench. The site is `archived-pr7519.fc.dev`, an archived site with 3 successful backups, 2 of them offsite.

Each result comes from the same site and the same login. Only the branch changes. Two dev servers ran at the same time, one on `develop` and one on this branch, thus the banner, the table and the data stay equal in every image.

### 1. Before — `develop`

The **Take Backup** button is there. The site is archived, thus the button can only make a job that fails.

Backups tab of an archived site on develop, with the Take Backup button

### 2. After — this branch

The button is gone. The banner, the Audit Trail button, the filters and all 3 backup rows stay.

Backups tab of the same archived site on this branch, without the Take Backup button

### 3. Control — an active site on this branch

The button stays for a site that is not archived, thus the change has no effect on the usual flow.

Backups tab of an active site on this branch, with the Take Backup button

### Summary

| Case | Branch | Site status | Take Backup |
| --- | --- | --- | --- |
| 1 | develop | Archived | shows |
| 2 | this branch | Archived | hidden |
| 3 | this branch | Active | shows |

The buttons in the toolbar of the backups tab, read from the DOM:

- develop, archived: `Audit Trail`, `Take Backup`
- this branch, archived: `Audit Trail`
- this branch, active: `Audit Trail`, `Take Backup`

### Note

`press.api.site.backup` has no check for an archived site. This change closes the path in the user interface only. A check in `Site.backup()` is possible in a different pull request.

PR #7529, opened 2026-09-22
fix(ui): Pin row actions column in list views

siduck:
Adds sticky column for actions in objectlist component which shows our table view in most pages

https://github.com/user-attachments/assets/431df119-e6ee-4950-a68f-94c6c2136837


siduck:
@mergify backport master

